PR #21, opened 2023-07-19
Different handling label analysis if result already calculated

giovanni-guidini:
This change is intended to work with the PATCHing of labels by the CLI after collecting them.
Currently doing that doesn't trigger another task in the worker. SO if the labels arraive after we finished calculating we are left with incomplete results on the database. That is fine as the CLI will calculate again.

I think this change might have been a bit premature, but the idea is to have a different way of handling the calculation if we already have results. Then we simply calculate the final result using the saved labels and add in the requested labels that we might not have had the first time around. This would be much much faster than calculating everything again.

### Legal Boilerplate

### DIFF
--- a/tasks/label_analysis.py
+++ b/tasks/label_analysis.py
@@ -62,9 +62,13 @@ class LabelAnalysisRequestProcessingTask(BaseCodecovTask):
             "Starting label analysis request",
             extra=dict(
                 request_id=request_id,
+                external_id=label_analysis_request.external_id,
                 commit=label_analysis_request.head_commit.commitid,
             ),
         )
+        if label_analysis_request.state_id == LabelAnalysisRequestState.FINISHED.db_id:
+            return self._handle_larq_already_calculated(label_analysis_request)
+
         try:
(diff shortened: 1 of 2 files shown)


giovanni-guidini:
> Not sure I entirely understand the workflow here but the code changes look OK to me.
> 
> So the CLI will PATCH the label analysis w/ additional labels and that triggers off another task that presumably hits this `LabelAnalysisRequestState.FINISHED` conditional. In that case we just combine the previous results with the calculated results for the new labels. Is that roughly correct?

Yes, that's the idea.
The thing that happens today is:

CLI sends a POST request to api that creates the original LabelAnalysisRequest instance and triggers this task. Initially this has no `requested_labels`.

Later on - after collecting the labels present - the CLI will PATCH the labels in the api, but (right now) the task will not be re-triggered. This is not a problem for using ATS because the CLI can sort out what labels to include / remove locally.

The thing is that we want to have metrics around how many tests ATS is saving. For that reason we need to re-trigger the LabelAnalysis task with the larq that has already been calculated and update the result.

scott-codecov (approved):

PR #29, opened 2023-07-24
poc for local + hot reloaded worker + shared

matt-codecov:
alrighty, i think this is ready for review

there are more ways to build `worker` images now:
```
make build # totally unchanged
make dev # new setup, hot-reloads worker
make dev-shared # new setup, hot-reloads worker and shared
```
the latter two tag their resulting docker images with `codecov/worker-dev` instead of `codecov/worker` so if you switch you will want to update your `docker-compose.yml` or whatever.

the downside of hot-reloading shared is, when you initially start the container, it takes an extra couple minutes to fully build shared. it'll look like it froze during installation, but if you come back with coffee it should be running. hot-reloading will be very speedy, it's just that initial container startup.

to toggle hot-reloading shared, just run `make dev` or `make dev-shared` and restart the `codecov/worker-dev` container.

to work on or update shared:
```
# this will reset your shared to the committed submodule hash
git submodules update --init --recursive

cd shared
git checkout <new hash>
cd ..
git add shared
git commit -m 'update shared'
```

this same approach should work for API as well. if it's approved here, joseph or i will do it there as well.

### the submodule part
instead of installing `shared` from GitHub at a specified hash, make it a `git submodule` with that same hash checked out and install it with `file:./shared#egg=shared`

with just this, it's easier to work with shared. change `shared`, rebuild your worker image. no more pushing WIPs to GitHub to copy the hash. however, it's not enough to hot-reload

### the hot-reload part
we install `watchmedo` which is a util that does hot-reloading. if you change a python file it'll re-run `worker.sh` and your changes will take effect

but shared is special. shared is imported from the `site-packages` directory rather than the submodule. we need another step to hot-reload shared

### hot-reloading shared

we can perform an "editable" install of `shared` with `pip install -e ./shared`. this binds the installed module to the original, in-tree sources. however, if we do that while building the docker images, it'll bind the module to the copies of the sources inside the docker container.

instead, we want to run it after we've started the container. at that point, we've bind-mounted our own copy of `worker`/`shared` into the container, and when we run `pip install -e ./shared`, it'll bind to our copy.

for reasons unknown, this only wants to work after a non-editable install has also been performed.

### Legal Boilerplate


requirements.dev
```diff
@@ -0,0 +1,313 @@
+#
```
matt-codecov, line 1: there's only one `requirements.in` file and it looks like this uses the same `requirements.in` as `requirements.txt`. what's the difference with this one?

maybe rename the normal `requirements.txt` to `requirements.base.txt` and then create `requirements.txt` with:
```
-r requirements.base.txt
file:/shared#egg=shared
```
so that it includes the normal deps and then adds shared. and for dev builds we just install the base reqs. we'd update `requirements.base.txt` with `pip-compile --output-file requirements.base.txt requirements.in`

Makefile
```diff
@@ -14,10 +14,20 @@ build.local:
 build.base:
 	DOCKER_BUILDKIT=1 docker build -f dockerscripts/Dockerfile.requirements . -t codecov/baseworker:latest --ssh default
 
+build.base_dev:
+	DOCKER_BUILDKIT=1 docker build --build-arg="REQS=requirements.dev" -f dockerscripts/Dockerfile.requirements . -t codecov/baseworker:latest --ssh default
+
+build.dev:
+	DOCKER_BUILDKIT=1 docker build -f dockerscripts/Dockerfile.local . -t codecov/worker:latest --build-arg RELEASE_VERSION="${release_version}" --ssh default
+
```
matt-codecov, line 22: i don't have an opinion on this myself, but i wonder if we should change these `-t` arguments to create different tags for these different builds? in case anybody accidentally pushes the wrong kind of image to dockerhub under the normal name

it means that to use the local shared image of worker you have to edit `docker-compose.yml` but that isn't a huge deal

matt-codecov (commented):
could you add `--ignore=shared` to the `addopts` list in worker's `pytest.ini`?

requirements.dev
joseph-sentry: The goal was to have all of the dependencies of shared be taken into account when doing the pip-compile so that there aren't any version conflicts, but that would require having shared in the requirements.in when we compile it, which complicates things slightly. I agree that we should have a requirements.base.txt but for the compilation step I'm not sure what to do.

dockerscripts/Dockerfile.local
```diff
@@ -0,0 +1,28 @@
...
+                cargo
+
+COPY            . /worker
+WORKDIR         /worker
+
+RUN             pip install watchdog[watchmedo]
+
+WORKDIR         /worker/shared
+RUN             pip install setuptools-rust pip-tools
+RUN             pip install -r requirements.txt
```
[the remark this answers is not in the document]
joseph-sentry: This is installing `shared`'s `requirements.txt`

giovanni-guidini:
One thing that is not very clear to me (possibly because I haven't really worked with git submodules), how do you specify what is the shared SHA when deploying?

Or these changes don't modify the deploy to staging/prod at all, just the local development setup?

giovanni-guidini (commented):
I'll say looks good, and I'm fine with the changes. But please update README.md with the instructions that are in the PR comment 🙏

PR #43, opened 2023-07-31
Update shared version

scott-codecov:
Updates the `shared` version and accommodates changes from:

* https://github.com/codecov/shared/pull/13 - handle null `fork` value in repo data
* https://github.com/codecov/shared/pull/16 - different format returned from `list_repos_using_installation`


tasks/sync_repos.py
```diff
@@ -89,7 +89,7 @@ async def sync_repos_using_integration(self, db_session, git, ownerid, username)
         with metrics.timer(f"{metrics_scope}.sync_repos_using_integration.list_repos"):
             repos = await git.list_repos_using_installation(username)
         if repos:
-            service_ids = {repo["id"] for repo in repos}
+            service_ids = {str(repo["repo"]["service_id"]) for repo in repos}
```
matt-codecov, line 92: why don't we just make this a string in the return format for both this and `list_repos()`? i don't think we use them anywhere else

matt-codecov (approved):
no blocking concerns, just consider the two Qs before merging. thank you!

tasks/sync_repos.py
scott-codecov: Made this change in shared and updated things here accordingly

matt-codecov (approved):
also great catch re: `.get()` vs `[]`

PR #47, opened 2023-08-02
feat: add "flow logger" + instrument upload flow with it

matt-codecov:
### tl;dr:
- new logging util, sends metrics to sentry like "time from the start of Task A to the end of Task B"
- two things to review:
  - `CheckpointLogger` and its tests
  - `UploadFlow` and the instrumentation of `UploadTask`, `UploadFinisherTask`, and `NotifyTask` (and tests)
- i'm experimenting here, feedback or better ideas are appreciated if you have them

tested end-to-end, see this dashboard in sentry with widgets for each of the subflows created:
Screenshot 2023-08-11 at 1 44 03 PM

will see about creating a dashboard in the real Codecov org when this is deployed to staging

related issue: https://github.com/codecov/platform-team/issues/84

### `CheckpointLogger` + `UploadFlow`
there are currently two things i want to measure that don't really fit in sentry transaction traces or statsd:
- the time the entire set of `UploadProcessorTask`s takes to run, median + p75 + p95
- the time between when we first receive an upload and when we send a notification about it, median + p75 + p95

you can figure these out for specific samples in sentry, but it didn't seem like you can aggregate/dashboard them effectively. and statsd timers are very code-local and can't capture these high-level event intervals. so i'm playing with a new thing

```python
class MyFlow(Enum)
    A = auto()
    B = auto()

checkpoints = CheckpointLogger(MyFlow)
checkpoints.log(MyFlow.A)
checkpoints.log(MyFlow.B)
checkpoints.submit_subflow("a_to_b", MyFlow.A, MyFlow.B)
```

the above snippet will record timestamps for A and B and then submit the interval to sentry as a [custom performance metric](https://docs.sentry.io/product/performance/metrics/#custom-performance-metrics) on the current transaction with the name "a_to_b". in worker, the current transaction is (i think) the task you're in when you submit the subflow. create a `CheckpointLogger` at the start of your flow and pass it as an argument to functions/tasks until you reach the end of your flow.

i've instrumented the upload flow and am hoping to make a dashboard in sentry with these metrics:
- `UploadTask`: `time_before_processing`, `initial_processing_duration`
- `UploadFinisherTask`: `batch_processing_duration`, `total_processing_duration`
- `NotifyTask`: `notification_latency`

`batch_processing_duration` is the "all the `UploadProcessorTask`s" metric and `notification_latency` is the end-to-end metric.

in the future, i am interested in a version of this that provides reliability metrics for these flows too. basically i want to add funnel analysis: of the N users who enter this flow (upload task), what % end in a success state (notifications sent), and what's the breakdown of where they fall out of the funnel (failed to process uploads, too many retries, no valid credentials)? let me know if you have any ideas :)

**OPEN Q**: is there anything i should know about clock synchronization between hosts in our deployments? the higher percentiles of these metrics are up there in the minutes so clock drift of a few seconds is probably fine, but i'm kind of hand-waving here

### Legal Boilerplate


giovanni-guidini (dismissed):
I like this, so I'll leave my ✅ , but interested on other people's opinions too.

I don't know of any ready made tools to do this, but I think it's clean implementation.

tasks/notify.py
```diff
@@ -50,6 +50,7 @@ async def run_async(
         commitid: str,
         current_yaml=None,
         empty_upload=None,
+        checkpoints=None,
```
scott-codecov, line 53: Would this be an instance of `CheckpointLogger`?  I believe all these task arguments need to be serializable in some way (not sure if JSON is used or something else) - so just wanted to make sure this was tested in some sort of E2E fashion.

scott-codecov (approved):
This looks really great @matt-codecov - excited to see it in action!

PR #57, opened 2023-08-10
feat: move `Pull.flare` to storage

giovanni-guidini:
This commit is the initial work to move `Pull.flare` to storage.
This column has crashed the DB in only 1 occasion in the past.

Notice that Pull doesn't have `commit` (because multiple commits refer to the same pull).
To deal with that I changed the writing of data to have 2 paths: one with commit, and one
without the commit. So pulls data will be written in the repo level (in storage).

codecov/platform-team#70

### Legal Boilerplate


database/models/core.py
```diff
@@ -303,6 +303,36 @@ def get_head_commit_notifications(self):
...
+    @property
+    def id(self):
+        return self.id_
+
+    def should_write_to_storage(self) -> bool:
+        if self.repository is None or self.repository.owner is None:
+            return False
+        is_codecov_repo = self.repository.owner.username == "codecov"
+        return should_write_data_to_storage_config_check(
+            master_switch_key="pull_flare",
```
scott-codecov, line 325: I see this is referencing a `pull_flare` config value but is a generic method on the model.  I feel like maybe we chatted about this scenario in the past but what if there are multiple `ArchiveField`s on a single model?

Not a big deal right now but this might require a little reworking if that case does arise in the future.

scott-codecov (approved):

PR #73, opened 2023-08-21
fix: handle reports with line 0 executed

giovanni-guidini:
Recently when testing the docs for integrating ATS I did [this commit](https://app.codecov.io/gh/giovanni-guidini/ats-data/commit/c24c1db8aa27fcda2d311fdcd293c20697030389)
for which processing failed. Checking the logs it failed because of

```
  [... removed stack trace...]
  File "/worker/services/report/languages/pycoverage.py", line 56, in process
    report_file.append(
  File "/usr/local/lib/python3.10/site-packages/shared/reports/resources.py", line 341, in append
    raise ValueError("Line number must be greater then 0. Got %s" % ln)
ValueError: Line number must be greater then 0. Got 0
```

Indeed when I donwloaded and checked the uploaded files there were 3 of them with `executed_lines: [0]`.
They were all `__init__.py` files. Idk why that happened, to be honest. It looks strange because the 3 files are empty.

In any case it's an easy fix that will not hurt us.
I think line 0 should not exist. It might indicate that when importing the code or something like that the file was
executed somehow, but for `__init__.py` files with some content in them the list of lines executed is as expected.

### Legal Boilerplate

### DIFF
--- a/services/report/languages/pycoverage.py
+++ b/services/report/languages/pycoverage.py
@@ -53,14 +53,15 @@ class PyCoverageProcessor(BaseLanguageProcessor):
                             str(ln), []
                         )
                     ]
-                    report_file.append(
-                        ln,
-                        report_builder_session.create_coverage_line(
-                            fixed_filename,
-                            cov,
-                            coverage_type=CoverageType.line,
-                            labels_list_of_lists=label_list_of_lists,
-                        ),
-                    )
+                    if ln > 0:
+                        report_file.append(
+                            ln,
+                            report_builder_session.create_coverage_line(
+                                fixed_filename,
+                                cov,
+                                coverage_type=CoverageType.line,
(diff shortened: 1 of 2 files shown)


matt-codecov (approved):
looks like `executed_lines` comes from `coverage.py`'s json format, and far as i can tell, a line number of 0 is a bug in `coverage.py`

https://github.com/nedbat/coveragepy/blob/c2b238e86cade2e6163b24656a7d0549f4508e8f/coverage/sqldata.py#L937 this function appears to filter out 0s in one path but not the other. no idea what the root cause is of the 0s is, but i take that as evidence that 0s are not supposed to be returned

if you still have the bugged `coverage.json`, maybe file an issue on https://github.com/nedbat/coveragepy to let the author know

PR #74, opened 2023-08-28
feat: Only send status checks when payload has changed

scott-codecov:
Resolves https://github.com/codecov/engineering-team/issues/350

Only send status notifications when the payload has changed from the last time we sent the same notification.

### DIFF
--- a/services/notification/notifiers/status/base.py
+++ b/services/notification/notifiers/status/base.py
@@ -6,6 +6,7 @@ from shared.analytics_tracking import track_event
 from shared.torngit.exceptions import TorngitClientError, TorngitError
 from shared.utils.sessions import SessionType
 
+from helpers.cache import DEFAULT_TTL, NO_VALUE, cache, make_hash_sha256
 from helpers.environment import is_enterprise
 from helpers.match import match
 from helpers.metrics import metrics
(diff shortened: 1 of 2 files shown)


services/notification/notifiers/status/base.py
```diff
@@ -209,7 +210,32 @@ async def notify(self, comparison: Comparison):
...
+                    repoid=head_commit.repoid,
+                    base_commitid=base_commit.commitid if base_commit else None,
+                    head_commitid=head_commit.commitid if head_commit else None,
+                    notifier_name=self.name,
+                    notifier_title=self.title,
+                )
+            )
+
+            last_payload = cache.get_backend().get(cache_key)
+            if last_payload is NO_VALUE or last_payload != payload:
```
giovanni-guidini, line 229: if the payload is JSON serializable (not sure if it is), can we use the payload as the key?
It would save us the `last_payload != payload` check... if the payload is there (as a key) we know we sent that before (in the last TTL seconds)

giovanni-guidini (approved):
looks good, left a comment about the cache key tho

services/notification/notifiers/status/base.py
scott-codecov: I'm not sure that would work as expected.  Imagine the following sequence of events...

1. notify with payload X -> hashes to ABC (not in cache so we send notification)
2. notify with payload X -> hashes to ABC (already cached, skip notification)
3. notify with payload Y -> hashes to DEF (not in cache so we send notification)
4. notify with payload X -> hashes to ABC (already in cache, skip notification)

We need to make sure that the last item there actually DOES send a notification.

giovanni-guidini: ah I see... Then indeed my understanding was incorrect.
I thought that case 4 there would indicate a task that got delayed for whatever reason (had to retry for example) and was "stale" data (because something more recent, in this case payload Y) was already delivered.

But indeed I see that case 4 would skip notification and it's not what we want.
Thanks for taking the time to explain.

PR #77, opened 2023-08-29
patch only cov feedback link

dana-yaish:
This edits the footer to have a feedback link dedicated to the patch cov only format 

### Legal Boilerplate


joseph-sentry (approved):
lgtm, just want to double check that the feedback link is meant to be different if project coverage is turned off vs on

dana-yaish:
yup they should be different

PR #78, opened 2023-08-29
chore: backfill commit data to storage

giovanni-guidini:
Now that we are writing data to GCS with acceptable degree of success we want to backfil data from existing commits.
For such purpose we are introducing a new task.
The present changes implement said new task.

closes codecov/engineering-team#189

### Legal Boilerplate


tasks/backfill_commit_data_to_storage.py
```diff
@@ -0,0 +1,153 @@
...
+            # write to storage and clears out db field
+            commit.report_json = commit._report_json
+            return {"success": True, "errors": []}
+        log.warning(
+            "Neither _report_json nor _report_json_storage_path are set. Nothing to do.",
+            extra=dict(commitid=commit.id, commit=commit.commitid),
+        )
+        return {"success": False, "errors": [BackfillError.missing_data.value]}
+
+    def handle_all_report_rows(
```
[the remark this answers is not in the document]
giovanni-guidini: The subtle difference that I've found that made me almost copy the code from there was that the `initialize_and_save_report` calls [save_full_report](https://github.com/codecov/worker/blob/8012c86931021e5c39df7f5ae9a8ec9db1f33cee/services/report/__init__.py#L965)

And that actually creates an `Upload` instance. We don't have one that isn't in the database, so I thought that was a no-no.

So the version here just does `save_report`.

tasks/backfill_commit_data_to_storage.py
```diff
@@ -0,0 +1,153 @@
...
+        if report_details is None:
+            report_details = ReportDetails(
+                report_id=report_row.id_,
+                _files_array=[],
+                report=report_row,
+            )
+            db_session.add(report_details)
+            db_session.flush()
+
+            repo_yaml = get_repo_yaml(commit.repository)
```
scott-codecov, line 111: Should this be `get_current_yaml` instead so that it takes into account the commit's YAML?

scott-codecov (approved):
One little question but otherwise look good

tasks/backfill_commit_data_to_storage.py
giovanni-guidini: I don't think it matters for the operations we are doing. And using the repo_yaml saves us a request to the git provider.

However from a correctness pov - given we will be working mostly with old commits - I guess the most accurate would be to use *only* the commit yaml... (given that more recent changes are probably merged to the repo yaml already and might have been done in the owner yaml)

What do you think is best?

scott-codecov: I think you're right that it doesn't matter.  Looks like there are 2 methods being called here:

* `report_service.get_existing_report_for_commit_from_legacy_data`
* `report_service.save_report`

I just looked through both and AFAICT neither rely on `self.current_yaml` in any way.  You could probably even pass `ReportService({})` here.

PR #89, opened 2023-09-06
feat: Add different comment message for 1st time pulls

scott-codecov:
https://github.com/codecov/engineering-team/issues/389

Companion PR to https://github.com/codecov/codecov-api/pull/108

If a pull is the 1st for a repo then we want to show different copy in the PR comment.

### DIFF
--- a/database/models/core.py
+++ b/database/models/core.py
@@ -2,6 +2,7 @@ import random
 import string
 import uuid
 from datetime import datetime
+from functools import cached_property
 
 from sqlalchemy import Column, ForeignKey, Index, UniqueConstraint, types
 from sqlalchemy.dialects import postgresql
@@ -297,7 +298,9 @@ class Pull(CodecovBaseModel):
     behind_by_commit = Column(types.Text)
 
     author = relationship(Owner)
-    repository = relationship(Repository, backref=backref("pulls", cascade="delete"))
+    repository = relationship(
+        Repository, backref=backref("pulls", cascade="delete", lazy="dynamic")
+    )
 
     __table_args__ = (Index("pulls_repoid_pullid", "repoid", "pullid", unique=True),)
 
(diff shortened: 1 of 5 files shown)


database/models/core.py
```diff
@@ -297,7 +297,9 @@ class Pull(CodecovBaseModel):
     behind_by_commit = Column(types.Text)
 
     author = relationship(Owner)
-    repository = relationship(Repository, backref=backref("pulls", cascade="delete"))
+    repository = relationship(
+        Repository, backref=backref("pulls", cascade="delete", lazy="dynamic")
```
giovanni-guidini, line 301: Very nice this `lazy='dynamic'` thing.

services/notification/notifiers/comment/__init__.py
```diff
@@ -351,18 +352,36 @@ def is_enabled(self) -> bool:
             self.notifier_yaml_settings, dict
         )
 
+    async def is_first_pull(self, comparison: Comparison) -> bool:
```
giovanni-guidini, line 355: Possibly a stupid question, but why is this `async`?

services/notification/notifiers/comment/__init__.py
```diff
@@ -351,18 +352,36 @@ def is_enabled(self) -> bool:
...
+        pull = comparison.pull
+        first_pull = (
```
giovanni-guidini, line 357: Should we think of caching this value, or you think that we would only ever access `is_first_pull()` roughly once / few times per `Pull` instance?

giovanni-guidini (approved):
looks good in general. I like the welcome message.
Just left a couple of potentially silly questions, but it's honestly fine as is.

services/notification/notifiers/comment/__init__.py
scott-codecov: It does IO to the database and is called in an async context so I figured this should be async as well.  I honestly don't understand how we're doing database queries in other async contexts elsewhere in the code - like why don't we have to `await` the calls into SQLAlchemy?

services/notification/notifiers/comment/__init__.py
scott-codecov: Would be nice to cache somehow.  Can't use `@cached_property` since it needs to cache per comparison - should I maybe just move this method to be defined on `Comparison` instead?

giovanni-guidini: I think it makes sense to move to `Comparison`.
I would argue it would make even more sense to move to `Pull` directly, in fact...

scott-codecov: Added an `is_first_pull` cached property to `Pull` here instead

PR #101, opened 2023-09-18
feat: Activate user who fired trial on trial expiration  

RulaKhaled:
On free plans, we offer 1 seat and we'd want to keep user who fired the trial activated post the trial. 

https://github.com/codecov/engineering-team/issues/406

### Legal Boilerplate


database/models/core.py
```diff
@@ -73,6 +73,7 @@ class Owner(CodecovBaseModel):
     trial_start_date = Column(types.DateTime, server_default=FetchedValue())
     trial_end_date = Column(types.DateTime, server_default=FetchedValue())
     trial_status = Column(types.Text, server_default=FetchedValue())
+    trial_fired_by = Column(types.Integer, server_default=FetchedValue())
```
dana-yaish, line 76: is this column already in the database?

dana-yaish: I guess this value is set from the API side when the trial is activated?

RulaKhaled: yes yes

PR #107, opened 2023-09-21
Add templates using jinja2

joseph-sentry:
This commit creates the TemplateService that uses
jinja to get and render templates.

This TemplateService provides the get_template method to get a rendered template. It takes the file name of the template as an arg and the template variables as the kwargs.

Templates should be added into the templates directory at the top level of the repository. This commit creates two example templates.

### DIFF
--- a/services/template.py
+++ b/services/template.py
@@ -1,10 +1,6 @@
 from jinja2 import Environment, PackageLoader, StrictUndefined, select_autoescape
 
 
-def get_template_service():
-    return TemplateService()
-
-
 class TemplateService:
     def __init__(self):
         # this loads the templates from the templates directory in this repository since it's looking for a dir named `templates` next to app.py
@@ -14,6 +10,6 @@ class TemplateService:
             undefined=StrictUndefined,
         )
 
-    def get_template(self, name, **kwargs):
+    def get_template(self, name):
         template = self.env.get_template(f"{name}")
-        return template.render(**kwargs)
+        return template
--- a/services/template.py
+++ b/services/template.py
@@ -9,12 +9,13 @@ class TemplateService:
         return cls.env is not None
 
     def __init__(self):
-        # this loads the templates from the templates directory in this repository since it's looking for a dir named `templates` next to app.py
-        TemplateService.env = Environment(
-            loader=PackageLoader("app"),
-            autoescape=select_autoescape(),
-            undefined=StrictUndefined,
-        )
+        if TemplateService.env is None:
+            # this loads the templates from the templates directory in this repository since it's looking for a dir named `templates` next to app.py
+            TemplateService.env = Environment(
+                loader=PackageLoader("app"),
+                autoescape=select_autoescape(),
+                undefined=StrictUndefined,
+            )
 
     def get_template(self, name):
         template = TemplateService.env.get_template(f"{name}")
--- a/services/template.py
+++ b/services/template.py
@@ -1,0 +1,28 @@
+from jinja2 import Environment, PackageLoader, StrictUndefined, select_autoescape
+
+_template_service = None
+
+
+def get_template_service():
+    return _get_cached_template_service()
+
+
+def _get_cached_template_service():
+    global _template_service
+    if _template_service is None:
+        _template_service = TemplateService()
+    return _template_service
+
+
+class TemplateService:
+    def __init__(self):
+        # this loads the templates from the templates directory in this repository since it's looking for a dir named `templates` next to app.py
+        self.env = Environment(
+            loader=PackageLoader("app"),
+            autoescape=select_autoescape(),
+            undefined=StrictUndefined,
+        )
+
+    def get_template(self, name, **kwargs):
+        template = self.env.get_template(f"{name}")
+        return template.render(**kwargs)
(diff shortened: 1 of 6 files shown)


services/template.py
```diff
@@ -0,0 +1,28 @@
+from jinja2 import Environment, PackageLoader, StrictUndefined, select_autoescape
+
+_template_service = None
+
+
+def get_template_service():
+    return _get_cached_template_service()
+
+
+def _get_cached_template_service():
```
giovanni-guidini, line 10: Do we need this to be a singleton?...
I ask because it doesn't seem to be a pattern we use much with `services` in the worker, and I don't know if it provides benefits to implement it like that.

My real concern is that we had issues before with tests (read ATS) around the cached ENV.

I'm not opposed to this idea, I'm just curious what are the benefits of doing it as a singleton. Does the loading process takes long? Is that recommended?

joseph-sentry: I don't think it has to be, I guess it does add complexity for no _real_ reason, i just thought it's probably an object that we don't need to instantiate a bunch of times, it's sort of just like load it and have it stay the same until we redeploy, but you're correct that we don't _need_ it to be a singleton

services/template.py
```diff
@@ -0,0 +1,19 @@
+from jinja2 import Environment, PackageLoader, StrictUndefined, select_autoescape
+
+
+def get_template_service():
```
giovanni-guidini, line 4: I think this is unnecessary. We can very well just instantiate `TemplateService()` directly

services/template.py
```diff
@@ -0,0 +1,19 @@
...
+class TemplateService:
+    def __init__(self):
+        # this loads the templates from the templates directory in this repository since it's looking for a dir named `templates` next to app.py
+        self.env = Environment(
+            loader=PackageLoader("app"),
+            autoescape=select_autoescape(),
+            undefined=StrictUndefined,
+        )
+
+    def get_template(self, name, **kwargs):
```
giovanni-guidini, line 17: [nit] change "name" to "template_name".

[less nit]
Personally I don't love that we return a rendered template from this function. I think that this function should simply get and return the template we specified, AND handle errors when we fail to get the template (like you ask for one that doesn't exist)

Then we modify the template object as we see fit and render it ourselves later on.

So I do think the responsibilities of this function should be limited to fetching a template and returning it (as the name suggests) and handling errors.

services/tests/test_template.py
```diff
@@ -0,0 +1,44 @@
+import pytest
+from jinja2.exceptions import UndefinedError
+
+from services.template import get_template_service
+
+
+class TestTemplate(object):
```
giovanni-guidini, line 7: I'd like to see a test where you ask for a template that doesn't exist

services/template.py
```diff
@@ -0,0 +1,15 @@
+from jinja2 import Environment, PackageLoader, StrictUndefined, select_autoescape
+
+
+class TemplateService:
+    def __init__(self):
+        # this loads the templates from the templates directory in this repository since it's looking for a dir named `templates` next to app.py
+        self.env = Environment(
```
matt-codecov, line 7: so, regarding using a singleton here:

https://jinja.palletsprojects.com/en/3.0.x/api/
> Most applications will create one [Environment](https://jinja.palletsprojects.com/en/3.0.x/api/#jinja2.Environment) object on application initialization and use that to load templates. In some cases however, it’s useful to have multiple environments side by side, if different configurations are in use.

https://stackoverflow.com/questions/618827/optimizing-jinja2-environment-creation
> My application is running on Google App Engine and most of requests constantly gets yellow flag due to high CPU usage. Using profiler I tracked the issue down to the routine of creating jinja2.Environment instance.

it looks like creating an `Environment` is expensive so we should try to only do it once. either that or we precompile templates with something like https://github.com/MiCHiLU/jinja2-precompiler but that approach doesn't seem well-documented.

i don't have a strong opinion on how to accomplish this. you could make `env` a global or class-level variable instead of an instance variable and be done with it, or you could make `TemplateService` a real singleton again. if you go the singleton route, there's a way to make it a singleton that you still invoke like `TemplateService().get_template()`: https://code.activestate.com/recipes/66531/ (maybe not technically a singleton but splitting hairs)

@giovanni-guidini what were the "cached ENV" issues you alluded to? were those, like, env vars or specifically a `jinja2.Environment`?

matt-codecov (changes requested):
all good to me, except unfortunately i have to give you whiplash on the singleton issue :v

i googled "jinja environment expensive" and learned you were on the right track making it a singleton before. i think gio's feedback about following the `FooService().bar()` convention of other services in our code is also valid, but we can bridge the gap under the hood

services/template.py
```diff
@@ -0,0 +1,21 @@
...
+class TemplateService:
+    env = None
+
+    @classmethod
+    def loaded(cls):
+        return cls.env is not None
+
+    def __init__(self):
+        # this loads the templates from the templates directory in this repository since it's looking for a dir named `templates` next to app.py
+        TemplateService.env = Environment(
```
matt-codecov, line 13: this will reinitialize `TemplateService.env` each time we create an instance, we should only initialize it if it's still `None`

matt-codecov (approved):
looks great, tiny tweak!

PR #108, opened 2023-09-21
Remove add_to_sendgrid_list task and related information

joseph-sentry:


giovanni-guidini (approved):
I know that we will update this with some other task, so looks good.

Please remember to check that we are, in fact, not using this prior to merging; and there's some info related to this task in `shared` as well.

PR #110, opened 2023-09-21
Add SMTPService and Email helper

joseph-sentry:
* Email helper class that wraps EmailMessage

* SMTPService class that creates connection to SMTP server using config options in services.smtp

* send method on SMTPService that takes Email as an arg to send an email using the SMTP connection

### DIFF
--- a/services/smtp.py
+++ b/services/smtp.py
@@ -34,6 +34,11 @@ class SMTPService:
         self.ssl_context = ssl.create_default_context()
         return True
 
+    def tls_and_auth(self):
+        self.try_starttls()
+        if self.username and self.password:
+            self.try_login()
+
     def try_starttls(self):
         # only necessary if SMTP server supports TLS and authentication,
         # for example mailhog does not need these two steps
@@ -75,9 +80,7 @@ class SMTPService:
             SMTPService.connection.connect(self.host, self.port)
         except smtplib.SMTPConnectError as exc:
             raise SMTPServiceError("Error starting connection for SMTPService") from exc
-        self.try_starttls()
-        if self.username and self.password:
-            self.try_login()
+        self.tls_and_auth()
 
     def __init__(self):
         if not self._load_config():
@@ -95,10 +98,7 @@ class SMTPService:
                     "Error starting connection for SMTPService"
                 ) from exc
 
-            self.try_starttls()
-
-            if self.username and self.password:
-                self.try_login()
+            self.tls_and_auth()
 
     def send(self, email: Email):
         if not SMTPService.connection:
--- a/services/smtp.py
+++ b/services/smtp.py
@@ -22,7 +22,7 @@ class SMTPService:
 
     @property
     def extra_dict(self):
-        return {self.host, self.port, self.username}
+        return {"host": self.host, "port": self.port, "username": self.username}
 
     def _load_config(self):
         if get_config("services", "smtp", default={}) == {}:
@@ -115,7 +115,8 @@ class SMTPService:
                 err_msg = " ".join(
                     list(map(lambda err_tuple: f"{err_tuple[0]} {err_tuple[1]}", errs))
                 )
-                raise SMTPServiceError(err_msg)
+                log.warning(f"Error sending email message: {err_msg}")
+                raise SMTPServiceError(f"Error sending email message: {err_msg}")
         except smtplib.SMTPRecipientsRefused as exc:
             log.warning("All recipients were refused", extra=self.extra_dict)
             raise SMTPServiceError("All recipients were refused") from exc
--- a/services/smtp.py
+++ b/services/smtp.py
@@ -1,44 +1,87 @@
+import logging
 import smtplib
 import ssl
-from functools import cached_property
 
 from shared.config import get_config
 
 from helpers.email import Email
 
-_smtp_service = None
 
-
-def get_smtp_service():
-    if len(get_config("services", "smtp", default={})) == 0:
-        return None
-    return _get_cached_smtp_service()
-
-
-def _get_cached_smtp_service():
-    global _smtp_service
-    if _smtp_service is None:
-        _smtp_service = SMTPService()
-    return _smtp_service
+log = logging.getLogger(__name__)
 
 
 class SMTPService:
-    def __init__(self):
+    connection = None
+
+    @classmethod
+    def active(cls):
+        return cls.connection is not None
+
+    def _load_config(self):
+        if get_config("services", "smtp", default={}) == {}:
+            return False
         self.host = get_config("services", "smtp", "host", default="mailhog")
         self.port = get_config("services", "smtp", "port", default=1025)
         self.username = get_config("services", "smtp", "username", default=None)
         self.password = get_config("services", "smtp", "password", default=None)
         self.ssl_context = ssl.create_default_context()
+        return True
 
-        self._conn = smtplib.SMTP(
-            host=self.host,
-            port=self.port,
-        )
-        self._conn.starttls(context=self.ssl_context)
-        if self.username and self.password:
-            self._conn.login(self.username, self.password)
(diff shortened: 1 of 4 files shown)


services/tests/test_smtp.py
```diff
@@ -0,0 +1,26 @@
+from ssl import create_default_context
+from unittest.mock import MagicMock, call
+
+from services.smtp import get_smtp_service
+
+
+class TestStorage(object):
```
giovanni-guidini, line 7: The class name 👀

giovanni-guidini (commented):
I think it's a good start.
We might add error handling / retries / metrics later on

Just the Test class name that I'd like to double check. The name is not very descriptive of what it does.

services/smtp.py
```diff
@@ -0,0 +1,44 @@
...
+def _get_cached_smtp_service():
+    global _smtp_service
+    if _smtp_service is None:
+        _smtp_service = SMTPService()
+    return _smtp_service
+
+
+class SMTPService:
+    def __init__(self):
+        self.host = get_config("services", "smtp", "host", default="mailhog")
```
giovanni-guidini, line 27: Does the connection work if I don't specify username and password?

services/smtp.py
```diff
@@ -0,0 +1,44 @@
...
+
+        self._conn = smtplib.SMTP(
+            host=self.host,
+            port=self.port,
+        )
+        self._conn.starttls(context=self.ssl_context)
+        if self.username and self.password:
+            self._conn.login(self.username, self.password)
+
+    def send(self, email: Email):
```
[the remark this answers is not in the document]
joseph-sentry: it was originally in the send_email task but I moved it over to this

services/smtp.py
joseph-sentry: Yes

services/smtp.py
```diff
@@ -0,0 +1,87 @@
...
+                    list(map(lambda err_tuple: f"{err_tuple[0]} {err_tuple[1]}", errs))
+                )
+        except smtplib.SMTPRecipientsRefused:
+            err_msg = "All recipients were refused"
+        except smtplib.SMTPSenderRefused:
+            err_msg = "Sender was refused"
+        except smtplib.SMTPDataError:
+            err_msg = "The SMTP server did not accept the data"
+
+        return err_msg
```
matt-codecov, line 87: should we just log this error message now instead of making the caller do it for us?

i think i'd prefer you create a new `SMTPServiceSendFailure` exception and raise it on error. it's more clear/aggressive than returning `None` on success and a value on failure, which is the opposite of how functions usually behave. but i think python's type system might be too weak to make me totally happy

_(**nothing actionable is below this point**, just explaining why i don't really like any options here)_

when i write error-handling code i keep two questions in mind:
- should the caller care about the outcome of this function?
- if so, how can i force the caller to deal with both success and failure?

on the first point, `SMTPService` doesn't have its own retry logic, so retrying is something the caller is responsible for. the caller should care if this function passes or fails. i think this makes sense; the caller is probably a celery task which has a ready-made `retry()` function but inside this function we'd need custom logic

but for the second point, python leaves me sad. compare to other languages:
- java makes you declare every type of exception your function can throw. so you either *must* place your call to `foo()` in a `try`/`catch` block that handles every exception `foo()` can throw, or you *must* declare your function to throw all the exceptions that `foo()` can throw (and punt the problem up a stack frame). either way, the compiler forces the caller to make a decision about how to handle the error
- C++ and Rust provide ways to require that you handle the return value after you call `foo()`. you can return `Result<Val, Err>` types annotated in this way and the compiler will make the caller explicitly decide how to handle them:
```
fn foo() -> Result<T, E>;

fn bar1() {
    foo(); // fails to build because return value is not assigned to anything
}

fn bar2() {
    let result = foo(); // fails because the result variable is never actually used
}

fn bar3() {
    let result = foo();
    if result.is_err() { // passes because we're actually using the result variable
        schedule_retry();
    }
}

fn bar4() {
    let _ = foo(); // passes because _ is a magic "i am intentionally ignoring this" identifier
}
```

python can make you handle neither return values nor exceptions. an uncaught exception will make the caller handle the error after they ship a bug into production, but python can't save them from the bug in the first place like other languages can

matt-codecov (commented):
despite how long my comments are i think you're very close to merging this. i just talk a lot, the PR is solid lol

services/smtp.py
```diff
@@ -0,0 +1,129 @@
...
+            log.warning("All recipients were refused", extra=self.extra_dict)
+            raise SMTPServiceError("All recipients were refused") from exc
+        except smtplib.SMTPSenderRefused as exc:
+            log.warning("Sender was refused", extra=self.extra_dict)
+            raise SMTPServiceError("Sender was refused") from exc
+        except smtplib.SMTPDataError as exc:
+            log.warning(
+                "The SMTP server did not accept the data", extra=self.extra_dict
+            )
+            raise SMTPServiceError("The SMTP server did not accept the data") from exc
```
matt-codecov, line 129: i think you could save some lines with the `err_msg` approach you used earlier

```
if not SMTPService.connection:
    // this one is returning right now which is inconsistent with raising the exception in the other error cases
    err_msg = "Connection was not initialized"
try:
    ....
except smtplib.SMTPReceipientsRefused:
    err_msg = "..."
except ...:
    err_msg = "..."

if err_msg:
    log.warning(err_msg)
    raise SMTPServiceError(err_msg)
```

services/smtp.py
```diff
@@ -0,0 +1,87 @@
...
+
+from shared.config import get_config
+
+from helpers.email import Email
+
+log = logging.getLogger(__name__)
+
+
+class SMTPService:
+    connection = None
```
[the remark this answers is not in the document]
matt-codecov: addressed in a different way, resolving

matt-codecov (approved):
see if you can add a test case for the "reconnect" logic:
```
mocker.patch.object(SMTPService.connection, 'noop', side_effect=SMTPServerDisconnected("foo"))
mock_fn = mocker.patch.object(smtp, 'make_connection')

smtp.send(...)
assert mock_fn.has_calls(...) # or whatever
```

approving with nits, great work seeing this through!

services/smtp.py
```diff
@@ -0,0 +1,131 @@
...
+
+            except smtplib.SMTPConnectError as exc:
+                raise SMTPServiceError(
+                    "Error starting connection for SMTPService"
+                ) from exc
+
+            self.try_starttls()
+
+            if self.username and self.password:
+                self.try_login()
```
matt-codecov, line 101: would be nice to get rid of this duplicated code. it's already in `make_connection()`, could we just use that here too? maybe add the `if SMTPService.connection is None` stuff inside `make_connection()` too

not a blocking comment, and sorry if it was like this before and i missed commenting on it!

services/tests/test_smtp.py
```diff
@@ -0,0 +1,257 @@
...
+class TestSMTP(object):
+    def test_correct_init(self, mocker, mock_configuration, set_username_and_password):
+        mocker.patch("smtplib.SMTP")
+
+        m = mocker.patch("ssl.create_default_context", return_value=MagicMock())
+        service = SMTPService()
+        service.connection.starttls.assert_called_with(context=m.return_value)
+        service.connection.login.assert_called_with("test_username", "test_password")
+
+    def test_idempotentconnectionection(self, mocker, mock_configuration):
```
matt-codecov, line 51: `test_idempotent_connection`

services/smtp.py
joseph-sentry: it was not there, I added these methods in one of my recent commits, the reason I'm not using make_connection here is that here we are creating the `SMTP` object, but in make_connection we're assuming that this step has already been done and we're just running `connect` again on that existing object

PR #113, opened 2023-09-22
fix: always modify line to account for special labels

giovanni-guidini:
fixes https://l.codecov.dev/fYqS4O

There was a situation where you only get `SpecialLabelsEnum` and so we were not calling
`_possibly_modify_line_to_account_for_special_labels` (which is the thing that transforms
the special label ENUM into a STR - so we can save it in the DB)

Now this is giving an error, so we need to fix it. These changes do fix it.
But why was that check there in the first place?

I can only theorize reasons... it might be to save processing time (since changing every line of
every file in the report is a lot of lines), and this condition simply slipped. OR it might indicate
some config error in the ATS process - e.g. we should never just receive the global label.

I suppose that we might have introduced this condition afterwards because when the code was written
it was impossible the case of uploading without running tests. I think now it's at least possible.

In any case I believe this fix should hold and we're not gonna have more problems because of it later on.

But wait... the actual error is `TypeError: Object of type SpecialLabelsEnum is not JSON serializable`,
why didn't you added that to `shared.utils.ReportEncoder.ReportEncoder`?

You are correct about that, little grasshopper. My reasoning was essencially that `SpecialLabelsEnum`
is defined in the worker level. We obviously have the logic to modify the labels in place there so
they never trickle down to `shared`.

I thought it more logical to handle `SpecialLabelsEnum` in the worker for all cases.
I was opposed to moving it to `shared`, and I think just adding `Enum` in general to the
`ReportEncoder` is too general.

### Legal Boilerplate

### DIFF
--- a/services/report/report_builder.py
+++ b/services/report/report_builder.py
@@ -81,9 +81,7 @@ class ReportBuilderSession(object):
         Returns:
             Report: The legacy report desired
         """
-        if self._present_labels and self._present_labels != {
-            SpecialLabelsEnum.CODECOV_ALL_LABELS_PLACEHOLDER
-        }:
+        if self._present_labels:
             for file in self._report:
                 for line_number, line in file.lines:
                     self._possibly_modify_line_to_account_for_special_labels(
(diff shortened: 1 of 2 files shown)


scott-codecov (approved):
I think this makes sense to me.  This would be a situation where only a global change was made and nothing else?  Does this somehow indicate that every test needs to be rerun?

giovanni-guidini:
This would be a situation where only a global change was made and nothing else?

This would indicate that we received and processed a report in which only global code was annotated to have been executed. I _think_ it means that no test was run, but some code was imported none the less... not exactly sure. 


Does this somehow indicate that every test needs to be rerun?

Yeah, pretty much. If there are changes to the global lines of code all tests need to re-run. It depends on the diff in the HEAD tho (assuming this will be used as BASE)

PR #122, opened 2023-09-28
perf/feat: use new generator variants of list_repos in sync_repos

matt-codecov:
updates shared to pull in https://github.com/codecov/shared/pull/46 (generator versions of list_repos) and https://github.com/codecov/shared/pull/41 (feature flag rollout util)

introduces a usage of the new `Feature` util to control rollout for a speed and UX improvement in `sync_repos`. currently it is rolled out to 0% of users but codecov repos are manually opted into it.

the problem: `list_repos()` frontloads fetching and doing light processing on every page of repos from the git provider before returning. when syncing hundreds of repos, the UI just has to spin and spin for potentially minutes and a new user can't see any progress that repos are being ingested into the system until they all appear at the very end.

the change (in `shared`): @adrian-codecov suggested making `list_repos()` a generator so that it can yield a page at a time back to `sync_repos` and `sync_repos` can incrementally feed repos into the database. if the UI refreshes its viewport or whatever, it will begin to see updated repos which probably makes for a better experience.

additionally, github syncs with this feature should go faster overall. github lets us fetch the total number of repos which we can use to precompute the range of pages, and therefore the list of page URLs, that we need to fetch. knowing the list of urls in advance lets us initiate all the fetches concurrently.

### Legal Boilerplate


giovanni-guidini (approved):
image


Nice how you re-used all tests and tested both paths 👍 
Strange to have unit tests that actually make a request to an external service

Issue #121, opened 2023-09-28
Mismatched counts of activated users


drazisil-codecov:
https://github.com/codecov/worker/blob/68f9fa1d4ae50a38d6036ead988e3ab51e3d4036/services/license.py#L54-L64

This query returns duplicate values, since there is no `DISTINCT` clause. As a result, we can count the license as being over the seat count and disable notifications, while still showing the customer that they are before the user count on the UI header, which pulls from here

https://github.com/codecov/codecov-api/blob/11e2a1ea42bd71ef9446bbe8b85d7513733e0171/services/self_hosted.py#L52-L68

These two queries should return the same count.

PR #132, opened 2023-10-04
Fix plan_activated_users count for an installation

joseph-sentry:
this PR changes the behaviour of `get_installation_plan_activated_users` to no longer count one user in multiple orgs as multiple users to match the behaviour in the API

Fixes: https://github.com/codecov/worker/issues/121

### DIFF
--- a/services/tests/test_activation.py
+++ b/services/tests/test_activation.py
@@ -92,8 +92,9 @@ class TestActivationServiceTestCase(object):
     def test_activate_user_failure_for_enterprise_pr_billing_no_seats(
         self, request, dbsession, mock_configuration, mocker, with_sql_functions
     ):
+        mocker.patch("services.license.is_enterprise", return_value=True)
+        mocker.patch("services.license._get_now", return_value=datetime(2020, 4, 2))
 
-        mocker.patch("helpers.environment.is_enterprise", return_value=True)
         # Create two orgs to ensure our seat availability checking works across
         # multiple organizations.
         org = OwnerFactory.create(
--- a/services/tests/test_activation.py
+++ b/services/tests/test_activation.py
@@ -62,7 +62,6 @@ class TestActivationServiceTestCase(object):
     def test_activate_user_success_for_enterprise_pr_billing(
         self, request, dbsession, mocker, mock_configuration, with_sql_functions
     ):
-
         mocker.patch("services.license.is_enterprise", return_value=True)
         mocker.patch("services.license._get_now", return_value=datetime(2020, 4, 2))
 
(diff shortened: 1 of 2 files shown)


services/tests/test_activation.py
```diff
@@ -89,11 +88,57 @@ def test_activate_user_success_for_enterprise_pr_billing(
...
+                    20,
+                    24,
+                )
+            ),
+            plan_auto_activate=True,
+        )
+        dbsession.add(org_second)
+        dbsession.flush()
+
+        encrypted_license = "wxWEJyYgIcFpi6nBSyKQZQeaQ9Eqpo3SXyUomAqQOzOFjdYB3A8fFM1rm+kOt2ehy9w95AzrQqrqfxi9HJIb2zLOMOB9tSy52OykVCzFtKPBNsXU/y5pQKOfV7iI3w9CHFh3tDwSwgjg8UsMXwQPOhrpvl2GdHpwEhFdaM2O3vY7iElFgZfk5D9E7qEnp+WysQwHKxDeKLI7jWCnBCBJLDjBJRSz0H7AfU55RQDqtTrnR+rsLDHOzJ80/VxwVYhb"
```
giovanni-guidini, line 122: How many seats does this license allows for?

services/tests/test_activation.py
```diff
@@ -89,11 +88,57 @@ def test_activate_user_success_for_enterprise_pr_billing(
         dbsession.commit()
         assert user.ownerid in org.plan_activated_users
 
+    def test_activate_user_success_user_org_overlap(
```
giovanni-guidini, line 91: I'm not sure I understand this test case.

I looks like you are creating 2 orgs with 5 seats taken each. Idk how many total seats this installation (self-hosted) can have. But it seems that all owners of the 2 orgs are distinct (e.g. with different IDs)

So the total count of seats used is 10, and you create a new owner to prove that we can extend that.

Is that the idea?


------
(assuming I'm correct - which I might very well not be)
What I'd like to see:
1. Call `get_installation_plan_activated_users` explicitly to see what users are activated (before and after creating the new user)
2. Have some overlap on the plan_activated_users across the orgs (cause that was the original problem, right?)

giovanni-guidini (commented):
Fix looks fine from the ticket description. I'm having a hard time understanding how the test relates. Asked some clarifying questions (thanks for indulging me 😅)

giovanni-guidini (approved):

services/tests/test_activation.py
joseph-sentry: added a comment to the code

PR #135, opened 2023-10-04
Add logs for some path fixes for each raw upload processing

joseph-sentry:
Fixes: https://github.com/codecov/engineering-team/issues/138


matt-codecov (approved):
should the fixes be in `extra` instead of the message itself? and what is the point of sampling? we just want to make sure that some are happening?

PR #139, opened 2023-10-05
adding log

dana-yaish:
this is the first step in debugging why the upload finisher gets executed multiple times by different worker pods

### Legal Boilerplate


tasks/upload_finisher.py
```diff
@@ -79,22 +78,40 @@ async def run_async(
...
+                    extra=dict(
+                        repoid=repoid,
+                        commit=commitid,
+                        parent_task=self.request.parent_id,
+                    ),
+                )
+                return result
+        except LockError:
+            log.warning(
+                "Unable to acquire lock for key %s. Retrying",
```
giovanni-guidini, line 108: Does it actually retry?...

giovanni-guidini (approved):
LGTM

tasks/upload_finisher.py
dana-yaish: ah, you're right, my bad

PR #140, opened 2023-10-05
feat: Add processing upload comment

scott-codecov:
https://github.com/codecov/engineering-team/issues/553

We'd like to be able to make an early PR comment (as soon as we receive an upload but before we've processed it) indicating that we've received the upload but it is still processing.  I'm reusing some of the "empty upload" functionality for this since it's similar in nature.  The API directly enqueues a notify task for empty uploads and I'm thinking we do the same for this - the API will enqueue both a notify task with `empty_upload="processing"` as well as a normal upload task (which later triggers another notify w/ report results).

Here's the corresponding PR for `codecov-api` that enqueues this early notify task: https://github.com/codecov/codecov-api/pull/179


services/notification/notifiers/comment/__init__.py
```diff
@@ -423,3 +425,8 @@ def _create_upgrade_message(self, comparison):
                 f"Coverage data is still being uploaded to Codecov Self-Hosted for the purposes of overall coverage calculations.",
                 f"Please contact your Codecov On-Premises installation administrator with any questions.",
             ]
+
+    def _create_processing_upload_message(self):
+        return [
+            "We're currently processing your upload.  This comment will be updated when the results are available.",
```
scott-codecov, line 431: Copy TBD here

dana-yaish (approved):
LGTM. One minor question, are we sure that this comment will get updated when the processing finishes? so we won't get 2 separate comments. It would be worth to check it with different behaviors of the comment, "default" and "new" for example

dana-yaish:
https://github.com/codecov/worker/pull/140#issuecomment-1749160655 maybe we can chain the notify task after the upload task just like we chain all 3 upload tasks together?

PR #143, opened 2023-10-12
Add link to requirements image caching guide

joseph-sentry:


docker/README.md
```diff
@@ -0,0 +1,2 @@
+# Use requirements image caching
```
giovanni-guidini, line 1: If you are adding private info (in the sense that you need to be a part of sentry to access it) in a repo that is publicly accessible - which I'm not sure is good practice - maybe mention it in the comment?

Also this doesn't seem to be necessary to build the image. So mention that too.

```
# Optional for Sentry devs - Use requirements image caching
```

-----
Idk if this is the best place to put this info. I'd totally overlook it.

joseph-sentry: Hmm, not sure where else to include this information to make it visible, you're right this is not necessary but does cut down the time needed to build the image by a significant amount, I'll close the PR for now however

PR #147, opened 2023-10-17
adding try except to catch any unexpected errors

dana-yaish:
Upload finisher is being retried mutiple times because of a raised error in timeseries. Adding a try-except block to catch any errors 
https://github.com/codecov/internal-issues/issues/97 

### Legal Boilerplate


services/timeseries.py
```diff
@@ -17,6 +17,21 @@
 
 def save_commit_measurements(
```
scott-codecov, line 18: This function is also called by the timeseries backfill task and I think we'd want to let it fail loudly in that case.  Could we move the try/except into the finisher task itself (where it calls this function)?

Alternatively, I wonder if enqueuing this as a separate task would make sense b/c otherwise we may be losing data if this fails spuriously.  Not sure though - that just might cause a backlog of tasks if Timescale is having issues or something.

dana-yaish: I'll move the try-except into the upload finisher for now. For creating a new task solution, what do you feel about having a dedicated queue for timescale tasks? will it be beneficial to have them away from the regular tasks that get triggered?

scott-codecov: Sounds good.  Yeah there's already a `timeseries` queue that the backfill stuff happens on.  We could offload saving a single commit's measurements on that same queue and then if Timescale is having issues that queue can just fill up for a bit w/o impacting everything else.

scott-codecov (approved):

services/timeseries.py
dana-yaish: do you want me to create a ticket for this to be planned in the next sprint maybe?

scott-codecov: That would be great - thank you!

PR #154, opened 2023-10-20
Update httpx and httpcore

scott-codecov:
I think we were running into this bug https://github.com/encode/httpcore/issues/564 (as seen here: https://codecov.sentry.io/issues/4463071748)

The fix was implemented in `httpcore 0.16.0`: https://github.com/encode/httpcore/pull/580.  I updated `httpx` and `httpcore` to incorporate that fix.

Sounds like the alternative is to use `asyncio.TaskGroup`s (which are available in Python 3.11) instead of `asyncio.gather` (which cancel the other tasks in the group when an exception arises in one of them)


drazisil-codecov (approved):
Assuming tests pass, looks good to me.

PR #156, opened 2023-10-23
fix: checkpointlogger: use enum declaration order instead of value order

matt-codecov:
i made a sloppy mistake. i treated `BaseFlow` subclass values as if they were sorted by declaration order, but they're just strings and they're sorted lexicographically. example:

```
class Foo(BaseFlow):
    PROCESSING_BEGIN = auto()
    INITIAL_PROCESSING_COMPLETE = auto()

assert Foo.PROCESSING_BEGIN < Foo.INITIAL_PROCESSING_COMPLETE
```

`PROCESSING_BEGIN` is declared before `INITIAL_PROCESSING_COMPLETE`, so i want `PROCESSING_BEGIN < INITIAL_PROCESSING_COMPLETE` to be true. but with lexicographic order, "I" is before "P" so it's flipped.

`Foo._member_names_` remembers the declaration order and we can fix the bug by overriding `__eq__()` and friends to make them compare based on declaration order.

### Legal Boilerplate

### DIFF
--- a/helpers/checkpoint_logger/__init__.py
+++ b/helpers/checkpoint_logger/__init__.py
@@ -43,7 +43,7 @@ class BaseFlow(str, Enum):
     is_failure: ClassVar[Callable[[T], bool]]
     log_counters: ClassVar[Callable[[T], None]]
 
-    def __new__(cls, value):
+    def __new__(cls: type[T], value: str) -> T:
         """
         Hook into the creation of each enum member and inject the class name
         into the enum's value (e.g. "MEMBER_NAME" -> "MyEnum.MEMBER_NAME")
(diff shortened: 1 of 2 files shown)


helpers/checkpoint_logger/__init__.py
```diff
@@ -43,22 +43,44 @@ class BaseFlow(str, Enum):
...
+    def __gt__(self, other: object) -> bool:
+        if isinstance(other, self.__class__):
+            return self.__class__._member_names_.index(
+                self.name
+            ) > self.__class__._member_names_.index(other.name)
+        return NotImplemented
+
+    def __lt__(self, other: object) -> bool:
+        if isinstance(other, self.__class__):
+            return (not self == other) and (not self > other)
```
scott-codecov, line 78: 👍

scott-codecov (approved):
Looks good.  I wonder if just using integer enum values would be overall a bit simpler than `auto()` and relying on definition order.

PR #158, opened 2023-10-24
fix: make sure to update commit info

giovanni-guidini:
The pre-process commit task doesn't check if the commit data has
been already pulled from the provider. This can cause an error where
the parent commit is not found to carryforward the reports.

These changes make sure that we try to update the commit info if necessary.
It also moves the function - that was repeated in a few places - to
RepositoryService.

### Legal Boilerplate

### DIFF
--- a/services/repository.py
+++ b/services/repository.py
@@ -126,6 +126,30 @@ async def fetch_appropriate_parent_for_commit(
     return closest_parent_without_message
 
 
+async def possibly_update_commit_from_provider_info(commit, repository_service):
+    repoid = commit.repoid
+    commitid = commit.commitid
+    try:
+        if not commit.message:
+            log.info(
+                "Commit does not have all needed info. Reaching provider to fetch info",
+                extra=dict(repoid=repoid, commit=commitid),
+            )
+            await update_commit_from_provider_info(repository_service, commit)
+            return True
+    except TorngitObjectNotFoundError:
+        log.warning(
+            "Could not update commit with info because it was not found at the provider",
+            extra=dict(repoid=repoid, commit=commitid),
+        )
+        return False
+    log.debug(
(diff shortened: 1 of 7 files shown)


scott-codecov (approved):
Change looks good.  I think we have a more general issue w/ our architecture around this though - just sharing some thoughts here.

The preprocess commit task is enqueued during the `create-report` stage of an upload.  The task gets put on a queue and a subsequent `do-upload` is run (which then enqueues an upload task).  Say the preprocess task gets popped by a worker that is slow to process the task ahead of it (tasks are prefetched in bulk).  Then the upload task might execute before preprocess and we have the same issue.  Granted this is probably not going to happen very often but I think we are not strictly enforcing the order of execution when we don't use Celery's `chain` functionality.

Am I overthinking this?

giovanni-guidini:
> Change looks good. I think we have a more general issue w/ our architecture around this though - just sharing some thoughts here.
> 
> The preprocess commit task is enqueued during the `create-report` stage of an upload. The task gets put on a queue and a subsequent `do-upload` is run (which then enqueues an upload task). Say the preprocess task gets popped by a worker that is slow to process the task ahead of it (tasks are prefetched in bulk). Then the upload task might execute before preprocess and we have the same issue. Granted this is probably not going to happen very often but I think we are not strictly enforcing the order of execution when we don't use Celery's `chain` functionality.
> 
> Am I overthinking this?

You are not overthinking. But I'd say strictly speaking we cannot assume a task will run before another. That is part of the celery architecture. But the upload task does have these same checks in place, so we shouldn't have problems.

giovanni-guidini:
Oh definitely you are right about `chain`. 
But these tasks come from different CLI commands. I don't think we can chain them together.

PR #166, opened 2023-11-01
feat: add team plan to billing and upload limits

JerrySentry:
- Add Team plan concept to upload limits, this plan has a 2500 limit.
- Add Team plan to be recognized as a billing plan

### Legal Boilerplate

### DIFF
--- a/services/tests/test_decoration.py
+++ b/services/tests/test_decoration.py
@@ -176,7 +176,9 @@ def gitlab_enriched_pull_root(dbsession, gitlab_root_group):
 
 
 class TestDecorationServiceTestCase(object):
-    def test_decoration_type_limited_upload(self, enriched_pull, dbsession, mocker):
+    def test_decoration_type_basic_plan_upload_limit(
+        self, enriched_pull, dbsession, mocker
+    ):
         mocker.patch("services.license.is_enterprise", return_value=False)
         pr_author = OwnerFactory.create(
             service="github",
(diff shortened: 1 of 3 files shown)


services/tests/test_decoration.py
```diff
@@ -205,6 +205,44 @@ def test_decoration_type_limited_upload(self, enriched_pull, dbsession, mocker):
...
+        enriched_pull.database_pull.repository.private = True
+
+        commit = CommitFactory.create(
+            repository=enriched_pull.database_pull.repository,
+            author__service="github",
+            timestamp=datetime.now(),
+        )
+
+        report = ReportFactory.create(commit=commit)
+        for i in range(2499):
```
adrian-codecov, line 230: How long does this test take? I wonder if there's an easier way to mock this 😅

services/tests/test_decoration.py
```diff
@@ -205,6 +205,44 @@ def test_decoration_type_limited_upload(self, enriched_pull, dbsession, mocker):
...
+            upload = UploadFactory.create(report=report, storage_path="url")
+            dbsession.add(upload)
+        dbsession.flush()
+
+        decoration_details = determine_decoration_details(enriched_pull)
+        assert decoration_details.decoration_type != Decoration.upload_limit
+        assert decoration_details.reason != "Org has exceeded the upload limit"
```
adrian-codecov, line 237: Nice add to this test. I'll leave this optional, but would be nice to add this to the `test_decoration_type_limited_upload` test (and also rename that one to `test_decoration_type_basic_plan_limit`

adrian-codecov (approved):
Overall lgtm! Optional suggestion otherwise gtg

services/tests/test_decoration.py
JerrySentry: 👍

services/tests/test_decoration.py
JerrySentry: Eh it added a quarter of a second on a pytest run with and without it in my docker container 🤷

PR #167, opened 2023-11-01
fix: remove some buggy validation of BaseFlow values

matt-codecov:
### tl;dr
fix `notification_latency` and other `UploadFlow` metrics by ceasing to inject the flow name into the value of each checkpoint in the flow
```
>>> json.dumps({ MyEnum.A: 1337 })

# Before
"{ 'MyEnum.A': 1337 }"

# After:
"{ 'A': 1337 }"
```

### explanation
when serializing `MyEnum` checkpoints between tasks, i tried to inject the enum name into each enum value so `MyEnum.A` would have a value of `"MyEnum.A"` instead of just `"A"`. the idea being, when trying to continue `MyEnum` in the next task, we could completely validate that the checkpoints data we're deserializing really came from `MyEnum`

my hacky method for accomplishing this didn't work and adding further complexity to this validation is not worth it, so this PR rips it out. there is still some protection against mismatching flows: checkpoints data is put into `kwargs` with a key that includes the flow name

### Legal Boilerplate

### DIFF
--- a/helpers/checkpoint_logger/__init__.py
+++ b/helpers/checkpoint_logger/__init__.py
@@ -65,6 +65,18 @@ CHECKPOINTS_SUBFLOW_DURATION = Histogram(
 )
 
 
+def _error(msg, flow, strict=False):
+    # When a new version of worker rolls out, it will pick up tasks that
+    # may have been enqueued by the old worker and be missing checkpoints
+    # data. At least for that reason, we want to allow failing softly.
+    metrics.incr("worker.checkpoint_logger.error")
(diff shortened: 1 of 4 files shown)


helpers/checkpoint_logger/__init__.py
```diff
@@ -65,14 +65,27 @@
 )
 
 
+def _error(msg, flow, strict=False):
+    # When a new version of worker rolls out, it will pick up tasks that
```
giovanni-guidini, line 69: Nice

giovanni-guidini (approved):

PR #168, opened 2023-11-02
add timeseries models and django scaffold

matt-codecov:
https://github.com/codecov/engineering-team/issues/547
https://github.com/codecov/shared/pull/72

this PR creates a django scaffold in worker which:
- imports database settings from `shared` (which were copied from `codecov-api`)
- installs `pg_telemetry` and `ts_telemetry` apps to gain access to those models
- calls `django.setup()` during worker startup
- uses no routing, asgi/wsgi, or other parts of django at all

it also introduces a helper `MetricContext` which wraps `pg_telemetry` and `ts_telemetry` so we can log to both until we pick one. and a little `TimeseriesTimer` timer utility as well. it uses `TimeseriesTimer` and `MetricContext` to log runtimes and success/failure/retry metrics in `CodecovBaseTask`. this gives us, for instance, per-repo success rates for `upload_task`, or per-owner runtimes for `sync_repos_task`

### Legal Boilerplate


joseph-sentry (approved):
I agree it doesn't make sense to test `settings.py` and `manage.py`, the rest LGTM

main.py
```diff
@@ -55,6 +56,9 @@ def mark_process_dead(pid, exitcode, **kwargs):
 def setup_worker():
     print(initialization_text.format(version=get_current_version()))
 
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "django_scaffold.settings")
+    django.setup()
```
giovanni-guidini, line 60: Does this attempts to open connections to DBs?

helpers/telemetry.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def wrapper(*args, **kwargs):
+        task = asyncio.create_task(fn(*args, **kwargs))
+        fire_and_forget.background_tasks.add(task)
+        task.add_done_callback(fire_and_forget.background_tasks.discard)
+
+    return wrapper
+
+
+class MetricContext:
```
giovanni-guidini, line 30: This is cool

helpers/telemetry.py
```diff
@@ -0,0 +1,169 @@
...
+        if self.populated:
+            return
+
+        repo = None
+        owner = None
+        commit = None
+        dbsession = get_db_session()
+
+        if self.repo_id:
+            repo = (
```
giovanni-guidini, line 76: Will this eventually be replaced with DJango ORM equivalents?
It seems that `PgSimpleMetric` and `TsSimpleMetric` are managed by Django (probably comes from `shared`) but this stuff isn't.

Not complaining or asking changes, just curious.

giovanni-guidini: Ah it's still undecided, I see
https://github.com/codecov/engineering-team/issues/547

giovanni-guidini (approved):

PR #171, opened 2023-11-03
handle new sections names 

dana-yaish:
handle new names for the new sections. newheader, newfooter and newfiles are now condensed_header, condensed_footer and condensed_files

### Legal Boilerplate


services/notification/notifiers/tests/unit/test_comment.py
```diff
@@ -4473,6 +4473,47 @@ async def test_build_message_no_project_coverage(
...
+        expected_result = [
+            f"## [Codecov]({pull_url}?src=pr&el=h1) Report",
+            f"Attention: `1 lines` in your changes are missing coverage. Please review.",
+            f"",
+            f"| [Files]({pull_url}?src=pr&el=tree) | Patch % | Lines |",
+            f"|---|---|---|",
+            f"| [file\\_1.go]({pull_url}?src=pr&el=tree#diff-ZmlsZV8xLmdv) | 66.67% | [1 Missing :warning: ]({pull_url}?src=pr&el=tree) |",
+            f"",
+            f"",
+            f":loudspeaker: Thoughts on this report? [Let us know!](https://about.codecov.io/pull-request-comment-report/).",
```
joseph-sentry, line 4510: It looks like the test is failing because of the `.` at the end of this line, otherwise LGTM

joseph-sentry (approved):

services/notification/notifiers/tests/unit/test_comment.py
dana-yaish: oh I didn't notice that, thank you I'll fix it

PR #180, opened 2023-11-13
WIP - encode labels into index

giovanni-guidini:
This is a lot of code, and I'll probably break this PR in several others. I'm looking for early feedback here. It's a lot of code and a lot of changes.

This depends on the changes in https://github.com/codecov/shared/pull/79. Check that PR for context.

# New things
* `services/report/labels_index.py` houses the `LabelIndexService`. This little bit of code controls read/write in the label index file (that is separate from `chunks.txt`). It does the loading and unloading of the `Report._labels_index` as needed.
* `MinioEndpoints.label_index` and functions to write-to and read-from the file. There can be 1 such file for each `CommitReport`.
* `services/report/raw_upload_processor.py::make_sure_orginal_report_is_using_label_ids` encodes the labels in a `Report` prior to merging it with another report (if needed)
* `services/report/raw_upload_processor.py::make_sure_label_indexes_match` "fixes" the index of the new report to be merged based on the original report's index so that we can be sure that both indexes point to the same labels.

# Important changes
* `pycoverage.py` language processor now spits out reports with a label lookup table and `CoverageDatapoint` already encoded
* The processing of raw reports with labels has extra steps (make_sure_orginal_report_is_using_label_ids, make_sure_label_indexes_match)
* The `label_analysis` task works with label_ids and only translates them in the end.

# Things missing
- [ ] Feature flag changes. I just changes everything to the "ideal" end state, but we'd have issues deploying this with different worker deploys co-existing
- [ ] `label_analysis` has to work if the report is not encoded with IDs as well.

### DIFF
--- a/helpers/labels.py
+++ b/helpers/labels.py
@@ -32,27 +32,23 @@ class SpecialLabelsEnum(Enum):
 
 
 @sentry_sdk.trace
-def get_labels_per_session(report: Report, sess_id: int):
+def get_label_indexes_per_session(report: Report, sess_id: int):
     all_labels = set()
     for rf in report:
         for _, line in rf.lines:
             if line.datapoints:
                 for datapoint in line.datapoints:
                     if datapoint.sessionid == sess_id:
-                        all_labels.update(datapoint.labels or [])
-    return all_labels - set(
-        [SpecialLabelsEnum.CODECOV_ALL_LABELS_PLACEHOLDER.corresponding_label]
-    )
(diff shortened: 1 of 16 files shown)


giovanni-guidini:
DON'T MERGE
This is a reference guide for the changes that will happen. It's the "end state" we're looking for. But I will try to separate the changes into slightly smaller PRs that are more digestible and easier to review. I will also add rollout capabilities for a safer rollout.

PR #185, opened 2023-11-20
chore(notify): add additional logging on deletion failures

drazisil-codecov:


giovanni-guidini (commented):
Good change. Tests appear to be broken...
Can `pull` be `None`? (because you are using `pull.repoid`)

giovanni-guidini (approved):

PR #187, opened 2023-11-20
feat: Add experimental AI PR review

scott-codecov:
Enable the feature in your `codecov.yml` and you'll start receiving GPT-4 powered PR reviews.  The YAML is checked in the `PullSyncTask` and, if that feature is enabled, then it spins off a new `AiPrReviewTask` which handles the interactions with GitHub and OpenAI.  This is GitHub only right now since the PR review API is a bit specific to GitHub (it's commenting on specific lines on code).

The GPT-4 prompt could probably use a lot of work as I have very little experience with this.  It's currently only passing in the Git patch and the comments received back from the OpenAI API have been somewhat lackluster.  Providing additional context about the codebase could potentially help here but is out-of-scope for this initial POC.

Depends on https://github.com/codecov/shared/pull/87 for the YAML schema update

~~TODO: update `shared` SHA once the PR linked above has been merged to `main`~~


services/ai_pr_review.py
```diff
@@ -0,0 +1,375 @@
...
+        log.debug(
+            "OpenAI response",
+            extra=dict(
+                res=res,
+            ),
+        )
+
+        try:
+            data = json.loads(res)
+        except json.decoder.JSONDecodeError:
```
giovanni-guidini, line 265: We should have metrics for these errors so we can track how successful this thing is (in terms of delivering any response, obviously it wouldn't measure how effective the response is)

giovanni-guidini (approved):
Looks solid. Interesting project, excited to see what comes out of it.

services/ai_pr_review.py
```diff
@@ -0,0 +1,375 @@
...
+        self.pullid = pullid
+        self._head_sha = None
+
+    @property
+    def token(self):
+        return self.torngit.get_token_by_type_if_none(None, TokenType.read)
+
+    async def fetch_diff(self) -> str:
+        async with self.torngit.get_client() as client:
+            diff = await self.torngit.api(
```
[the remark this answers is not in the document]
scott-codecov: The difference here is the `Accept` header below.  By passing that header, GitHub will return the raw Git patch (which you can also get while logged in by appending `.diff` to the end of a normal PR URL).

services/ai_pr_review.py
scott-codecov: Added in https://github.com/codecov/worker/pull/187/commits/3de8687d6ec6de57b7bd0321b82dd21e29a5e20e

PR #188, opened 2023-11-21
Prep the terrain for reports with label compression.

giovanni-guidini:
context: codecov/engineering-team #768

Creates a rollout for label compression. This is gonna help us to test and safely release the feature in the wild.
Notice that currently the label compression does nothing.
There are comments similar to `TODO: needs shared update`.
The update in question is https://github.com/codecov/shared/pull/79

So these changes mostly prep the terrain for future changes that will actually do something,
and add the guardrails to avoid issues when deploying.
In particular it adds some helper methods to the `ArchiveService`, creates a kinda-stubbed `LabelsIndexService`,
passes more data to the `_adjust_sessions` portion of `raw_upload_processor` where most changes will occur.

If you're curious as to what the end result will probably look like see https://github.com/codecov/worker/pull/180

### Legal Boilerplate

### DIFF
--- a/services/report/labels_index.py
+++ b/services/report/labels_index.py
@@ -1,0 +1,46 @@
+from typing import Dict
+
+from shared.reports.resources import Report
+
+from database.models.reports import CommitReport
+from services.archive import ArchiveService
+
+
+class LabelsIndexService(object):
+    _archive_client: ArchiveService
+
+    def __init__(self, commit_report: CommitReport) -> None:
+        self.commit_report = commit_report
+        self._archive_client = ArchiveService(
+            repository=commit_report.commit.repository
+        )
+        self.commit_sha = commit_report.commit.commitid
+
+    def set_label_idx(self, report: Report):
+        # TODO: Needs shared update.
+        # if report._labels_index is not None:
+        #     raise Exception(
+        #         "Trying to set labels_index of Report, but it's already set"
+        #     )
+        # Load label index from storage
+        # JSON uses strings are keys, but we are using ints.
+        map_with_str_keys = self._archive_client.read_label_index(
+            self.commit_sha, self.commit_report.code
+        )
+        loaded_index = {int(k): v for k, v in map_with_str_keys.items()}
+        return loaded_index
+        # TODO: Needs shared update.
+        # report.set_label_idx(loaded_index)
+
+    def unset_label_idx(self, report: Report, label_index: Dict[str, str]):
+        # Write the updated index back into storage
+        # TODO: Needs shared update
+        # self._archive_client.write_label_index(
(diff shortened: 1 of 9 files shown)


services/report/raw_upload_processor.py
```diff
@@ -164,6 +185,24 @@ def _adjust_sessions(original_report, to_merge_report, to_merge_session, current
...
+    if (
+        upload
+        and USE_LABEL_INDEX_IN_REPORT_PROCESSING_BY_REPO_SLUG.check_value(
+            repo_slug(upload.report.commit.repository), default=False
+        )
+        and to_partially_overwrite_flags
+    ):
+        label_index_service = LabelsIndexService(upload.report)
+        # TODO: Needs shared upload
+        # if original_report._labels_index is None:
```
scott-codecov, line 199: If `_labels_index` is meant to be accessed from the outside then maybe remove the leading `_` (it's convention for private methods and attrs)

services/report/raw_upload_processor.py
```diff
@@ -164,6 +185,24 @@ def _adjust_sessions(original_report, to_merge_report, to_merge_session, current
...
+        #     label_index_service.set_label_idx(original_report)
+        # # Make sure that the labels in the reports are in a good state to merge them
+        # make_sure_orginal_report_is_using_label_ids(original_report)
```
scott-codecov, line 202: Where will these functions come from?

giovanni-guidini: they will be implemented in that file... check https://github.com/codecov/worker/pull/189/files#diff-a6e093dd207a55f7406f52e6aaaa6fc0533ac9b5cf965a0d8ffdffc257216f49

services/report/raw_upload_processor.py
giovanni-guidini: That's fair... I'll change that in `shared` and adjust in the PR that actually makes use of it

scott-codecov (approved):

matt-codecov (approved):

services/report/labels_index.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def unset_label_idx(self, report: Report, label_index: Dict[str, str]):
+        # Write the updated index back into storage
+        # TODO: Needs shared update
+        # self._archive_client.write_label_index(
+        #     self.commit_sha, report._labels_index, self.commit_report.code
+        # )
+        self._archive_client.write_label_index(
+            self.commit_sha, label_index, self.commit_report.code
+        )
```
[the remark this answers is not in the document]
giovanni-guidini: Good question... In my mind you take the index from storage into memory, possibly modify it, and once you're done with it expect it to be re-written to storage to preserve changes.

I see that it's not obvious though...

PR #189, opened 2023-11-22
allow for encoded labels in `Report`

giovanni-guidini:
These changes extend the `Report` to use the `_labels_index`.
Everything is controlled via a rollout, so in practice there's a single repo that will use
the compressed labels, my sentry fork (giovanni-guidini/sentry). For testing purposes.

There are 2 important facts here:
1. Given worker / shared pairs we can only have (old, old) or (new, new).
    where "old" denote the code that doesn't do label encoding, and "new" the code that does.
2. The (new, new) pair can handle the cases the (old, old) does and the new one of label encoding,
    but the (old, old) pair will generate corrupted results if it tries to process an encoded report.

SO it's important to woll out the (new, new) pair into prod _before_ turning the feature on for everyone.
Which is what we are doing, so we're fine (probably).

At first I was very much in the "let's tear everything down and build again", but because the (new, new) pair
needs to handle the unencoded versions as well I decided to make smaller changes so the code can co-exist in harmony.

### DIFF
--- a/services/report/labels_index.py
+++ b/services/report/labels_index.py
@@ -1,5 +1,3 @@
-from typing import Dict
-
 from shared.reports.resources import Report
 
 from database.models.reports import CommitReport
@@ -16,31 +14,23 @@ class LabelsIndexService(object):
         )
         self.commit_sha = commit_report.commit.commitid
 
-    def set_label_idx(self, report: Report):
-        # TODO: Needs shared update.
-        # if report._labels_index is not None:
-        #     raise Exception(
-        #         "Trying to set labels_index of Report, but it's already set"
-        #     )
+    def set_label_idx(self, report: Report) -> None:
+        if report._labels_index is not None:
+            raise Exception(
+                "Trying to set labels_index of Report, but it's already set"
+            )
         # Load label index from storage
         # JSON uses strings are keys, but we are using ints.
         map_with_str_keys = self._archive_client.read_label_index(
             self.commit_sha, self.commit_report.code
         )
         loaded_index = {int(k): v for k, v in map_with_str_keys.items()}
-        return loaded_index
-        # TODO: Needs shared update.
-        # report.set_label_idx(loaded_index)
+        report.set_label_idx(loaded_index)
 
-    def unset_label_idx(self, report: Report, label_index: Dict[str, str]):
+    def unset_label_idx(self, report: Report) -> None:
         # Write the updated index back into storage
-        # TODO: Needs shared update
-        # self._archive_client.write_label_index(
-        #     self.commit_sha, report._labels_index, self.commit_report.code
-        # )
         self._archive_client.write_label_index(
-            self.commit_sha, label_index, self.commit_report.code
+            self.commit_sha, report._labels_index, self.commit_report.code
         )
         # Remove reference to it
-        # TODO: Needs shared update
-        # report.unset_label_idx()
+        report.unset_label_idx()
--- a/services/report/languages/pycoverage.py
+++ b/services/report/languages/pycoverage.py
@@ -1,4 +1,4 @@
-import typing
+from typing import Any, Dict, List, Optional, Union
 
 from shared.reports.resources import Report
 
@@ -22,23 +22,54 @@ class PyCoverageProcessor(BaseLanguageProcessor):
             and "files" in content
         )
 
-    def _convert_testname_to_label(self, testname, labels_table):
+    def _convert_testname_to_label(self, testname) -> str:
         if type(testname) == int or type(testname) == float:
             # This is from a compressed report.
             # Pull label from the labels_table
             # But the labels_table keys are strings, because of JSON format
-            testname = labels_table[str(testname)]
+            testname = self.labels_table[str(testname)]
         if testname == "":
             return SpecialLabelsEnum.CODECOV_ALL_LABELS_PLACEHOLDER
         return testname.split("|", 1)[0]
 
-    def process(
-        self, name: str, content: typing.Any, report_builder: ReportBuilder
-    ) -> Report:
+    def _get_list_of_label_ids(
+        self,
+        current_label_idx: Optional[Dict[int, str]],
+        line_contexts: List[Union[str, int]] = None,
+    ) -> List[int]:
+        if self.are_labels_already_encoded:
+            # The line contexts already include indexes in the table.
+            # We can re-use the table and don't have to do anything with contexts.
+            return sorted(map(int, line_contexts))
+
+        # In this case we do need to fix the labels
+        label_ids_for_line = set()
+        for testname in line_contexts:
+            clean_label = self._convert_testname_to_label(testname)
+            if clean_label in self.reverse_table:
+                label_ids_for_line.add(self.reverse_table[clean_label])
+            else:
+                label_id = max([*current_label_idx.keys(), 0]) + 1
+                current_label_idx[label_id] = clean_label
+                self.reverse_table[clean_label] = label_id
+                label_ids_for_line.add(label_id)
+
+        return sorted(label_ids_for_line)
+
+    def process(self, name: str, content: Any, report_builder: ReportBuilder) -> Report:
         report_builder_session = report_builder.create_report_builder_session(name)
-        labels_table = None
+        # Compressed pycoverage files will include a labels_table
+        # Mapping label_idx: int --> label: str
+        self.labels_table: Dict[int, str] = None
+        self.reverse_table = {}
+        self.are_labels_already_encoded = False
         if "labels_table" in content:
-            labels_table = content["labels_table"]
+            self.labels_table = content["labels_table"]
+            # We can pre-populate some of the indexes that will be used
+            for idx, testname in self.labels_table.items():
+                clean_label = self._convert_testname_to_label(testname)
+                report_builder_session.label_index[int(idx)] = clean_label
+            self.are_labels_already_encoded = True
         for filename, file_coverage in content["files"].items():
             fixed_filename = report_builder.path_fixer(filename)
             if fixed_filename:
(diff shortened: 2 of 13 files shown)


services/report/__init__.py
```diff
@@ -719,6 +722,30 @@ async def create_new_report_for_commit(self, commit: Commit) -> Report:
                 paths_to_carryforward,
                 session_extras=dict(carriedforward_from=parent_commit.commitid),
             )
+            # If the parent report has labels we also need to carryforward the label index
+            # Considerations:
+            #   1. It's necessary for labels flags to be carryforward, so it's ok to carryforward the entire index
+            #   2. As tests are renamed the index might start to be filled with stale labels. This is not good.
+            #      but I'm unsure if we should try to clean it up at this point. Cleaning it up requires going through
+            #      all lines of the report. It might be better suited for an offline job.
```
scott-codecov, line 730: I agree - could this happen in the upload finisher perhaps **after** triggering the notification task?

services/report/labels_index.py
```diff
@@ -2,45 +2,68 @@
...
+        self.commit_sha = commit_sha
+
+    def caryforward_label_idx(self, new_report: Report):
+        """Sets the labels_index for this commit in another report.
+        The other report can then save this index by creating
+        a LabelsIndexService instance for itself.
+        """
+        my_index = self._get_label_idx()
+
+        if my_index != dict():
```
scott-codecov, line 42: I think just `if my_index:` would accomplish the same here

services/report/labels_index.py
```diff
@@ -2,45 +2,68 @@
...
+from services.report.report_builder import SpecialLabelsEnum
 
 
 class LabelsIndexService(object):
     _archive_client: ArchiveService
+    commit_report_code: str
+    commit_sha: str
 
-    def __init__(self, commit_report: CommitReport) -> None:
-        self.commit_report = commit_report
```
scott-codecov, line 13: Curious about this change - what's wrong with passing in the commit report?

services/report/languages/pycoverage.py
```diff
@@ -22,23 +22,54 @@ def matches_content(self, content, first_line, name) -> bool:
             and "files" in content
         )
 
-    def _convert_testname_to_label(self, testname, labels_table):
+    def _convert_testname_to_label(self, testname) -> str:
```
scott-codecov, line 25: I find the `testname` terminology a little confusing since I wasn't sure what the difference between a label and a test name is.  Maybe `_normalize_label` would describe this behavior?

services/report/report_builder.py
```diff
@@ -173,7 +191,9 @@ def create_coverage_line(
         coverage,
         *,
         coverage_type: CoverageType,
-        labels_list_of_lists: typing.List[typing.Union[str, SpecialLabelsEnum]] = None,
+        labels_list_of_lists: Union[
+            List[Union[str, SpecialLabelsEnum]], List[int]
```
scott-codecov, line 195: 🤯

tasks/label_analysis.py
```diff
@@ -29,6 +31,8 @@
     SpecialLabelsEnum.CODECOV_ALL_LABELS_PLACEHOLDER.corresponding_label
 )
 
+GLOBAL_LEVEL_LABEL_IDX = 0
```
scott-codecov, line 34: Seems like this should go deeper in the report service or `shared` or something.  I saw `0` explicitly referenced elsewhere and should maybe use this constant instead

tasks/label_analysis.py
```diff
@@ -40,6 +44,24 @@ class LinesRelevantToChange(TypedDict):
     files: Dict[str, Optional[LinesRelevantToChangeInFile]]
 
 
+class ExistingLabelSetsEncoded(NamedTuple):
```
scott-codecov, line 47: What's the advantage of `NamedTuple` over using `@dataclass`?

scott-codecov (commented):
Dang, this gets complicated.  I almost wonder if it'd be easier to make sure the next label request (after this is deployed) just returns ALL labels and then we can just carry on from there using label IDs only.  Not entirely sure how to make that happen and maybe there's additional nuance that makes such an approach infeasible.  This is just a lot of logic to handle the 2 potential label encodings

services/report/languages/pycoverage.py
```diff
@@ -47,12 +78,20 @@ def process(
...
+                                file_coverage.get("contexts", {}).get(str(ln), []),
+                            )
+                        ]
+                    else:
+                        label_list_of_lists = [
+                            [self._convert_testname_to_label(testname)]
+                            for testname in file_coverage.get("contexts", {}).get(
+                                str(ln), []
+                            )
+                        ]
```
matt-codecov, line 94: top branch creates an outer list containing a single large inner list, bottom branch creates an outer list containing many single-entry inner lists. is it okay that they don't match?

services/report/languages/pycoverage.py
```diff
@@ -64,4 +103,7 @@ def process(
                             ),
                         )
                 report_builder_session.append(report_file)
+        # I don't know if we reuse this processor, but just to be sure
+        # Erase the reverse table
+        self.reverse_table = None
```
matt-codecov, line 108: unset `self.labels_table` as well?

services/report/raw_upload_processor.py
```diff
@@ -159,6 +186,100 @@ class SessionAdjustmentResult(object):
...
+            if label == label_or_id:
+                reverse_index_cache[label] = idx
+                return idx
+        # Label is not present. Add to index.
+        # Notice that this never picks index 0, that is reserved for the special label
+        new_index = max(original_report._labels_index.keys()) + 1
+        reverse_index_cache[label_or_id] = new_index
+        # It's OK to update this here because it's inside the
+        # UploadProcessing lock, so it's exclusive access
+        original_report._labels_index[new_index] = label_or_id
```
matt-codecov, line 220: `_labels_index` is just an instance var, right? what would the race be?

is there an alternative to this that would not make parallel upload processing harder?

matt-codecov (commented):
i don't really understand what's happening, i should maybe pause and ramp up and then try reviewing again

services/report/report_builder.py
```diff
@@ -84,21 +89,33 @@ def output_report(self) -> Report:
...
-            self._report._totals = None
+                for file in self._report:
+                    for line_number, line in file.lines:
+                        self._possibly_modify_line_to_account_for_special_labels(
+                            file, line_number, line
+                        )
+                self._report._totals = None
         return self._report
 
+    # TODO: This can be removed after label indexing is rolled out for all customers
```
matt-codecov, line 118: can you leave a github issue number in the comment so we can grep for it and actually make sure we clean everything up

services/report/labels_index.py
```diff
@@ -2,45 +2,68 @@
...
-        #     )
+
+    def __init__(
+        self, repository: Repository, commit_sha: str, commit_report_code: str
+    ) -> None:
+        self._archive_client = ArchiveService(repository=repository)
+        self.commit_report_code = commit_report_code
+        self.commit_sha = commit_sha
+
+    def caryforward_label_idx(self, new_report: Report):
```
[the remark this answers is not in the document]
giovanni-guidini: Thanks <3

services/report/languages/pycoverage.py
giovanni-guidini: Probably a good idea, yeah

services/report/raw_upload_processor.py
giovanni-guidini: The race would be different instances of the same report (that all pull the labels_index from the same file) writing different things to it and then trying to save the report to the same destination. But because all changes happen only in the _merge_ portion of processing it's fine. THe merge needs to be in a lock anyway.

However I should be more careful that the report might be processing and not processing at the same time... and only save it into storage if it was modified.

Anyway, it doesn't really matter for parallel processing, because the uploads will be processed into temporary reports with their indexes. You can even merge them into a single report to be merged with the original in parallel, just take care of the index (with the aux functions in this file). But ultimately you will have to put a lock in the original report when merging to it, right?

services/report/report_builder.py
giovanni-guidini: THat's a fine idea. I'll do that.

services/report/languages/pycoverage.py
giovanni-guidini: THat's a very good catch. There is a difference indeed.

This goes to `ReportBuilderSession.create_coverage_line` ([here](https://github.com/codecov/worker/blob/gio/label-compressed/update-shared/services/report/report_builder.py#L202)). On the top it creates a single datapoint with all labels. In the bottom it creates many datapoints, one label per datapoint.

This difference is crucial when updating the report with new uploads. In particular when we [delete datapoints](https://github.com/codecov/shared/blob/gio/label-lookup/shared/reports/editable.py#L52). It seems that the bottom one is more precise and keeps data available longer. Sadly it also uses more space. But I'll keep the same way it was before.

tasks/label_analysis.py
giovanni-guidini: No idea, really. But originally this was a tuple, so I thought I'd keep it similar to a tuple.

giovanni-guidini: Apparently there are considerations to be made: https://stackoverflow.com/questions/51671699/data-classes-vs-typing-namedtuple-primary-use-cases

But the usage is quite small for this case in particular.
What I want out of this is _typing_ info, to improve code readability.
I'm OK changing to dataclass if preferred.

services/report/labels_index.py
giovanni-guidini: There's nothing wrong with doing that per se, but it doesn't match all usages.

In the carry forward bit the `commit_report` doesn't exist yet - or I don't have access to it (for the parent commit) - and yet I create `LabelIndexServices` for the commits. So I needed to be able to create them without the `commit_report`.

giovanni-guidini: I have since taken Matt's suggestion around that bit of code tho, so I can revert back the code if preferred.

services/report/__init__.py
giovanni-guidini: I'm unsure... certainly after the notification task. But I'm not sure about the upload finisher because it is scheduled for every `upload` task. It does have a lock for it, but maybe I'll have to implement a multi-reader/single write lock for the labels file?

My biggest concern with putting it in the upload finisher is that there might still be uploads being actively processed at the same time. I think the cleanup should happen a few hours after the last upload came in, to be extra sure.

giovanni-guidini:
This PR will be replaced by another one.

PR #190, opened 2023-11-22
ref: Refactor upload task to accomodate multiple report types

scott-codecov:
This is a proposal for how we could potentially handle the ingest of multiple types of upload.  It assumes we're reusing the existing `CommitReport` model and adding a new `report_type` column to that table.  Existing behavior should be unchanged (for `report_type = "coverage"`, the default) and allows us to reuse much of the ancillary support behavior that the upload task currently provides (fetching commit info, all the locking around processing, etc.).

When `report_type = "coverage"` we create our regular chain of tasks (upload processor -> finisher -> notify) but we would likely trigger other (new) tasks to handle the processing and notification related to other types of ingests.

I noticed that we were passing around `repoid` + `commitid` + `report_type` + `report_code` a lot so I also refactored that into a new class called `UploadArgs` which encapsulates these (together with the arguments that are passed to the task via Redis).

The existing `ReportService` has some generic methods (like creating/retrieving uploads) but we'd want to create different adapters for each report type.  I left some comments in the code about specific interfaces we're currently relying on.


giovanni-guidini:
I like this idea in general. I think that if we can re-use the entry point of uploads processing it's gonna be a positive thing.
I think in the future there are other optimizations we might try to make, mostly in terms of reducing the span of the code that runs within a lock, but these should be done incrementally.

The refactor itself was very good. Special thanks to the helpful TODO comments.
I like how logic that was all over the place is isolated in a single class (`UploadArgs`), although (and this is nit) I don't like the name. There's much logic in `UploadArgs`, that sounds to me like a data container. I'd call it either `UploadContext` or `UploadContextManager`, and you may or may not choose to have an `UploadArgs` data container in there. Like I said, nit stuff.

PR #197, opened 2023-12-01
feat: add models and task for failed test ingestion

joseph-sentry:
Depends on: https://github.com/codecov/test-results-parser/pull/4


tasks/test_results_processor.py
```diff
@@ -0,0 +1,201 @@
...
+            "successful": True,
+            "testrun_list": [self.testrun_to_dict(t) for t in testrun_list],
+        }
+
+    def save_report(
+        self, db_session: Session, testrun_list: List[Testrun], upload: Upload
+    ):
+        for testrun in testrun_list:
+            test = (
+                db_session.query(Test)
```
trent-codecov, line 103: This is an n+1 query. Query for the list of tests outside the loop. Then check for existence in the loop.

tasks/test_results_processor.py
```diff
@@ -0,0 +1,201 @@
...
+        archive_service = ArchiveService(repository)
+
+        payload_bytes = archive_service.read_file(upload.storage_path)
+        data = json.loads(payload_bytes)
+
+        testrun_list = []
+
+        for file in data["test_results_files"]:
+            file = file["data"]
+            print(file)
```
trent-codecov, line 136: Debugging

tasks/test_results_processor.py
```diff
@@ -0,0 +1,201 @@
...
+            file_bytes = BytesIO(zlib.decompress(base64.b64decode(file)))
+            print(file_bytes)
```
trent-codecov, line 138: Debugging

tasks/test_results_processor.py
```diff
@@ -0,0 +1,201 @@
...
+            # TODO: improve report matching capabilities
+            # use file extensions?
+            # maybe do the matching in the testing result parser lib?
+            try:
+                if first_line.startswith(b"<?xml"):
+                    testrun_list = parse_junit_xml(file_content)
+                elif b"pytest" in first_line:
+                    testrun_list = parse_pytest_reportlog(file_content)
+                else:
+                    testrun_list = parse_vitest_json(file_content)
```
trent-codecov, line 154: Should we maybe match something for vitest and then "else" raise a not supported error? 

This might get weird otherwise.

tasks/test_results_processor.py
```diff
@@ -0,0 +1,200 @@
...
+        for args in arguments_list:
+            upload_obj: Upload = (
+                db_session.query(Upload).filter_by(id_=args.get("upload_pk")).first()
+            )
+
+            try:
+                parsed_testruns: List[Testrun] = self.process_individual_arg(
+                    upload_obj, upload_obj.report.commit.repository
+                )
+            except Exception:
```
giovanni-guidini, line 69: It's fine to leave a catch all exception for starters so we avoid crashing the task while we battle test it. But please be explicit about it, as having a catch-all exception is not good practice.

Also the warning message won't be helpful when debugging. Make sure to include the stack trace so we know what to fix (in case there's anything)

tasks/test_results_processor.py
```diff
@@ -0,0 +1,200 @@
...
+                log.warning(f"Error parsing testruns")
+                return {"successful": False}
+
+            # concat existing and new test information
+            testrun_list += parsed_testruns
```
giovanni-guidini, line 74: Looking at the `upload` task it seems that you decided to go with a `group`, that is parallel in nature (meaning we might be processing multiple uploads at the same time for the same commit).

This indicates that it's ok to be saving data concurrently for the same commit. I wonder then if it's necessary to accumulate the parsing results in `testrun_list` prior to saving them. I think this list might become possibly big (we know of users that have +100K tests, for example), and this seems like it could use a fairly large amount of data.

tasks/test_results_processor.py
```diff
@@ -0,0 +1,200 @@
...
+    def save_report(
+        self, db_session: Session, testrun_list: List[Testrun], upload: Upload
+    ):
+        repo_tests = (
+            db_session.query(Test).filter_by(repoid=upload.report.commit.repoid).all()
+        )
+        test_set = set()
+        for test in repo_tests:
+            test_set.add(f"{test.testsuite}::{test.name}")
+        for testrun in testrun_list:
```
giovanni-guidini, line 107: I wonder if we can leverage https://docs.sqlalchemy.org/en/13/orm/session_api.html#sqlalchemy.orm.session.Session.merge `session.merge` here to offload this logic to the database...

I have a feeling it'd be more efficient, if possible.

tasks/test_results_processor.py
```diff
@@ -0,0 +1,200 @@
...
+
+            first_line = file_bytes.readline()
+            file_bytes.seek(0)
+
+            file_content = file_bytes.read()
+
+            # TODO: improve report matching capabilities
+            # use file extensions?
+            # maybe do the matching in the testing result parser lib?
+            first_line = bytes("".join(first_line.decode("utf-8").split()), "utf-8")
```
giovanni-guidini, line 149: [nit] report matching should be in its own function

tasks/test_results_processor.py
```diff
@@ -0,0 +1,200 @@
...
+            try:
+                if first_line.startswith(b"<?xml"):
+                    testrun_list = parse_junit_xml(file_content)
+                elif first_line.startswith(b'{"pytest_version": "7.4.3'):
+                    testrun_list = parse_pytest_reportlog(file_content)
+                elif first_line.startswith(b'"{"numTotalTestSuites":'):
+                    testrun_list = parse_vitest_json(file_content)
+            except Exception as e:
+                log.warning(f"Error parsing: {file_content.decode()}")
```
giovanni-guidini, line 158: Please use a static message an add extra args to the logger. That will make it easier to search for logs and it's the norm for worker logs

tasks/tests/unit/test_test_results_processor_task.py
```diff
@@ -0,0 +1,518 @@
...
+            commit_yaml={"codecov": {"max_report_age": False}},
+            arguments_list=redis_queue,
+        )
+        expected_result = {
+            "successful": True,
+            "testrun_list": [
+                {
+                    "duration_seconds": 0.001,
+                    "name": "api.temp.calculator.test_calculator::test_add",
+                    "outcome": "Outcome.Pass",
```
giovanni-guidini, line 65: Where is this "Outcome.Pass" defined? In the model the field is an Integer, but I can't find the Enum for it....

joseph-sentry: it's defined in the `test-results-parser` library

tasks/test_results_processor.py
joseph-sentry: I could save the instances as soon as we parse them, but for tests I think it's better to avoid doing it on every iteration since we will be getting the list of all tests for that repository

tasks/tests/unit/test_test_results_processor_task.py
giovanni-guidini: Can we import the type definition into the python code? nice... works like an enum or does it think it's just a string?

tasks/test_results_processor.py
matt-codecov: i think we could upsert https://docs.sqlalchemy.org/en/14/dialects/postgresql.html#insert-on-conflict-upsert

matt-codecov (changes requested):
first off: awesome work on the feature, i think the code is designed pretty well and i think we'll be able to build on it well for flaky test detection, etc

just a callout: we don't have a pattern in place or all the missing pieces we need to avoid loading everything into memory here, so beware that this implementation is at risk of the same performance problems coverage is until we address that

- `archive_service.read_file()` reads into memory, we'd need to save to disk
- `json.loads()` doesn't stream, we'd need another json parser
- (above is enough to avoid loading more than one results file into memory at once)
- (if we are worried individual results files will be enormous we also need the following)
- `zlib` appears to have a streaming decompression util https://docs.python.org/3/library/zlib.html#zlib.decompressobj
- a streaming base64 decoder
- a rework of the rust code to take a bytes stream abstraction instead of a concrete `Vec<u8>` or `&[u8]`

matt-codecov (approved):

database/models/reports.py
```diff
@@ -240,3 +240,31 @@ class CompareComponent(MixinBaseClass, CodecovBaseModel):
...
+            "name",
+            "testsuite",
+            name="reports_test_repoid_name_testsuite",
+        ),
+    )
+
+
+class TestInstance(CodecovBaseModel, MixinBaseClass):
+    __tablename__ = "reports_testinstance"
+    test_id = Column(types.Integer, ForeignKey("reports_test.id"))
```
matt-codecov, line 264: hold up, if we change this to a hash of the test suite + test name we might be able to unlock performance improvements

the reason we can't do the db work in the processor task is because we need to synchronize to get/create records of the `Test` model because we need to put its id on the `TestInstance` model. if the ID could be computed from the test suite + test name, then we don't need to synchronize with other tasks to create our `TestInstance` records

as for creating the corresponding `Test` records, we could do `INSERT ... ON CONFLICT DO NOTHING` or `INSERT ... WHERE NOT EXISTS (...)` to attempt inserts in a fire-and-forget way and the database will handle synchronizing internally

matt-codecov (changes requested):
un-accepting until we talk about whether this perf change is feasible quickly, worth blocking for, or should be deferred

tasks/upload.py
```diff
@@ -678,6 +686,61 @@ def _schedule_bundle_analysis_processing_task(
...
+                        commit_yaml=commit_yaml,
+                        arguments_list=chunk,
+                        report_code=commit_report.code,
+                    ),
+                )
+                processor_task_group.append(sig)
+        if processor_task_group:
+            checkpoint_data = None
+            if checkpoints:
+                checkpoints.log(UploadFlow.INITIAL_PROCESSING_COMPLETE)
```
matt-codecov, line 715: i think we need to use different flows for coverage, bundles, and test results. we have totally different expectations/norms for how long each takes, and putting them together in one pool makes the metrics useless

database/models/reports.py
joseph-sentry: we would have to hash the env as well but I agree that this may get rid of the issue where we want to use a `returning` clause

tasks/test_results_processor.py
```diff
@@ -0,0 +1,238 @@
...
+            res = self.process_individual_upload(
+                db_session, repoid, commitid, upload_obj
+            )
+
+            # concat existing and new test information
+            testrun_dict_list.append(res)
+
+            upload_list.append(upload_obj)
+
+        # kick off notification task stuff
```
matt-codecov, line 78: comment inaccurate?

matt-codecov (approved):

database/models/reports.py
```diff
@@ -240,3 +240,31 @@ class CompareComponent(MixinBaseClass, CodecovBaseModel):
...
+
+class Test(CodecovBaseModel, MixinBaseClass):
+    __tablename__ = "reports_test"
+    repoid = Column(types.Integer, ForeignKey("repos.repoid"))
+    repository = relationship("Repository", backref=backref("tests"))
+    name = Column(types.String(256), nullable=False)
+    testsuite = Column(types.String(256), nullable=False)
+
+    __table_args__ = (
+        UniqueConstraint(
```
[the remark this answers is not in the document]
joseph-sentry: https://github.com/codecov/codecov-api/pull/357 because `id_` is just a sha256 hash of all of those

database/models/reports.py
```diff
@@ -240,3 +243,46 @@ class CompareComponent(MixinBaseClass, CodecovBaseModel):
...
+        UUID(as_uuid=True), default=uuid.uuid4, unique=True, nullable=False
+    )
+    created_at = Column(types.DateTime(timezone=True), default=get_utc_now)
+    updated_at = Column(
+        types.DateTime(timezone=True), onupdate=get_utc_now, default=get_utc_now
+    )
+
+    @property
+    def id(self):
+        return self.id_
```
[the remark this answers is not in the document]
joseph-sentry: this is just because we're replacing the `MixinBaseClass` properties, I'm not sure why but `MixinBaseClass` does this for the `id` field and i just wanted to keep it consistent

PR #199, opened 2023-12-02
fix: use original PR base to compute patch coverage

matt-codecov:
https://github.com/codecov/engineering-team/issues/814

this code could use some cleanup one day but this is a minimal change to remove unnecessary inaccuracy from patch coverage

currently for both patch coverage and project coverage deltas we compare the PR head's report against the base, or if the base isn't in our database, the next-oldest commit (the "adjusted base"). that's necessary for project coverage, it's unnecessary (and problematic) for patch coverage

this PR:
- saves `original_base_commitid` in the `Comparison` object
  - in places where we can't access the PR we just assume the original is the same as the adjusted base
- adds `use_original_base` switch to `ComparisonProxy.get_diff()`
- add `use_original_base=True` to calls to `get_diff()` in the notification service
  - `services/notification/notifiers/mixins/status.py`: context is explicitly patch-coverage-related
  - `services/notification/notifiers/checks/patch.py`: explicitly a patch-coverage check
  - `services/notification/notifiers/mixins/message/__init__.py`: the diff computed here is plumbed through all of the various comment section implementations. afaict the only usages of it are patch-coverage-related, so changing the top of the funnel shouldn't impact anything else

### extra notes
- afaict `tasks/save_report_results.py` is dead code
- `tasks/sync_pull.py` computes patch coverage against the correct base and saves it to `pull.diff` in the database! but then afaict nothing ever reads from it ([link](https://github.com/codecov/worker/blob/8508cc5b912383af2a224e39641f6cc00831f332/tasks/sync_pull.py#L256))
  - notification code should be able to get this value with `comparison.enriched_pull.database_pull.diff` and not have to recompute it every time, but the `sync_pull` task runs after the `notify` task and changing that is Too Much for right now
  - https://github.com/codecov/engineering-team/issues/884 documents this opportunity for improvement
- `tasks/upload_processor.py` computes patch coverage for just the current commit (not compared to the PR base) but doesn't do anything with it
  - internally, the `Report` model will use it to update `ReportFileSummary.diff_totals` for each file in the report, and that gets saved in `network_file_json()` or whatever it's called and then rehydrated. but after a quick scan i couldn't find anything that tries to read this field
- the reason we adjust the base is because we want the base to have a report, but we never actually check for it. maybe we can't/don't check for it because we're hoping by the time we get to sending notifications the report is there?
- i am leaving `tasks/compute_comparison.py` alone
- `pull.base` is the real base and `pull.compared_to` is the adjusted base, but everywhere else they use `base` to refer to the adjusted base so i had to call the real base `original_base`

### Legal Boilerplate


services/comparison/types.py
```diff
@@ -17,7 +17,19 @@ class FullCommit(object):
 @dataclass
 class Comparison(object):
     head: FullCommit
+
+    # To see how a patch changes project coverage, we compare the branch head's
+    # report against the base's report, or if the base isn't in our database,
+    # the next-oldest commit that is. Be aware that this base commit may not be
+    # the true base that, for example, a PR is based on.
     base: FullCommit
```
giovanni-guidini, line 25: I wonder if using `project_coverage_base` and `patch_coverage_base` would be better names from an understanding perspective. 

Although the comments do a good job of explaining the difference.

giovanni-guidini (approved):
LGTM

services/comparison/types.py
matt-codecov: oh you're right those are much better names. i tried renaming `base` to `adjusted_base` and it was a very messy rename so i backed out since merging this was a priority, but i assigned myself a GH issue to rename the fields with your suggested names when i get a spare minute https://github.com/codecov/worker/issues/203

PR #200, opened 2023-12-04
Use encoded labels in reports with them

giovanni-guidini:
Uses encoded labels with the labels_index to point to labels in the right direction.


helpers/labels.py
```diff
@@ -32,27 +33,45 @@ def __init__(self, val):
...
             if line.datapoints:
                 for datapoint in line.datapoints:
                     if datapoint.sessionid == sess_id:
-                        all_labels.update(datapoint.labels or [])
+                        all_labels.update(datapoint.label_ids or [])
     return all_labels - set(
-        [SpecialLabelsEnum.CODECOV_ALL_LABELS_PLACEHOLDER.corresponding_label]
+        [
+            SpecialLabelsEnum.CODECOV_ALL_LABELS_PLACEHOLDER.corresponding_label,
+            0,  # This is always the index for the SpecialLabelsEnum
```
scott-codecov, line 53: Us `GLOBAL_LEVEL_LABEL_IDX` here instead?

helpers/labels.py
```diff
@@ -32,27 +33,45 @@ def __init__(self, val):
...
         for _, line in rf.lines:
             if line.datapoints:
                 for datapoint in line.datapoints:
-                    all_labels.update(datapoint.labels or [])
+                    all_labels.update(datapoint.label_ids or [])
     return all_labels - set(
-        [SpecialLabelsEnum.CODECOV_ALL_LABELS_PLACEHOLDER.corresponding_label]
+        [
+            SpecialLabelsEnum.CODECOV_ALL_LABELS_PLACEHOLDER.corresponding_label,
+            0,  # This is always the index for the SpecialLabelsEnum
```
scott-codecov, line 75: `GLOBAL_LEVEL_LABEL_IDX`

services/report/__init__.py
```diff
@@ -546,7 +547,7 @@ async def _do_build_report_from_commit(self, commit) -> Report:
         report = self.get_existing_report_for_commit(commit)
         if report is not None:
             return report
-        return await self.create_new_report_for_commit(commit)
+        return await self.create_new_report_for_commit(commit, None)
```
scott-codecov, line 550: Just for readability I think passing the args as `(commit, report_code=None)` would be nice

services/report/__init__.py
```diff
@@ -662,7 +663,9 @@ async def _possibly_shift_carryforward_report(
                 )
             return carryforward_report
 
-    async def create_new_report_for_commit(self, commit: Commit) -> Report:
+    async def create_new_report_for_commit(
+        self, commit: Commit, report_code: str = None
```
scott-codecov, line 667: Is `report_code` needed any more?  I think that was so you knew where to read from storage.

giovanni-guidini: You are correct

scott-codecov (approved):

PR #209, opened 2023-12-08
Change default upper section from "header" to "condensed_header"

joseph-sentry:


services/notification/notifiers/mixins/message/__init__.py
```diff
@@ -209,7 +209,7 @@ def get_upper_section_names(self, settings):
         sections = list(map(lambda l: l.strip(), (settings["layout"] or "").split(",")))
         headers = ["newheader", "header", "condensed_header"]
         if all(not x in sections for x in headers):
-            sections.insert(0, "header")
+            sections.insert(0, "newheader")
```
giovanni-guidini, line 212: "condensed_header" is the new name for "newheader" given that "newheader" will not be "new" if it's the default. 

NIT comment being use "condensed_header" as the value, not "newheader"

giovanni-guidini (approved):
LGTM but left small comment

PR #210, opened 2023-12-08
feat: call CleanLabelsIndexTask

giovanni-guidini:
If the processed upload includes a flag that is carryforward
AND "labels" mode, we enqueue the CleanLabelsIndexTask.

Depends on https://github.com/codecov/worker/pull/205

I tried to be conservative, but it's possible that - despite
using the flag with labels - no labels are uploaded. It's
also possible that there's nothing to do. But these are small
details.

ps.: To see that indeed we have an "arguments" and "flag" in
"processing_results" I looked at logs for the message
"Scheduling notify task" :wink:

closes codecov/engineering-team#768

### Legal Boilerplate

### DIFF
--- a/tasks/tests/unit/test_upload_finisher_task.py
+++ b/tasks/tests/unit/test_upload_finisher_task.py
@@ -403,6 +403,7 @@ class TestUploadFinisherTask(object):
                 "app.tasks.notify.Notify": mocker.MagicMock(),
                 "app.tasks.pulls.Sync": mocker.MagicMock(),
                 "app.tasks.compute_comparison.ComputeComparison": mocker.MagicMock(),
+                "app.tasks.upload.UploadCleanLabelsIndex": mocker.MagicMock(),
             },
         )
         repository = RepositoryFactory.create(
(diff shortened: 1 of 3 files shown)


tasks/upload_finisher.py
```diff
@@ -207,13 +208,47 @@ async def finish_reports_processing(
...
+        for result in actual_processing_results:
+            if result["successful"]:
+                flags_str: str = result.get("arguments", {}).get("flags", "")
+                if flags_str:
+                    individual_flags = flags_str.split(",")
+                    for flag in individual_flags:
+                        config = commit_yaml.get_flag_configuration(flag)
+                        if config and config.get("carryforward_mode") == "labels":
+                            return True
+        return False
```
matt-codecov, line 250: i try to avoid needing this many levels of indentation, it can make logic hard to follow. can't always do it without perf consequences, but i think here perf doesn't matter

imo the code would be tidier if it broke the logic into three nested sub-checks that we could give names to:
- we need to clean the **label index** if any of our **processing results** need to be cleaned
- a **processing result** needs to be cleaned if a) it's successful, and b) any of **its flags** need to be cleaned
- a **flag** needs to be cleaned if a) it has a yaml config, and b) that config's mode is "labels"

```
def should_clean_labels_index(self, commit_yaml, processing_results):
    def should_clean_for_flag(flag):
        config = commit_yaml.get_flag_configuration(flag)
        return config and config.get("carryforward_mode", "") == "labels"

    def should_clean_for_processing_result(result):
        args = results.get("arguments", {})
        flags_str = args.get("flags", "")
        flags = flags_str.split(",") if flags_str else []
        return result["successful"] and any(map(flags, should_clean_for_flag))

    actual_processing_results = processing_results.get("processings_so_far", [])
    return any(map(actual_processing_results, should_clean_for_processing_result))
```

tasks/tests/unit/test_upload_finisher_task.py
```diff
@@ -507,3 +573,78 @@ async def test_upload_finisher_task_calls_save_commit_measurements(
...
+                            "name": "just-tests",
+                            "carryforward": True,
+                        },
+                    ]
+                }
+            }
+        )
+        instance = (
+            UploadFinisherTask()
+        )  # Replace with the actual class name where the method is defined
```
matt-codecov, line 648: ?

matt-codecov (approved):
pls consider the feedback but i don't need to see it again before you merge, your tests should catch any mistakes

PR #222, opened 2023-12-21
fix: commit to db after computing flag and component comparisons

joseph-sentry:
This is an attempt to fix an issue where flag comparisons are not being saved to the database.


adrian-codecov (approved):
Odd this isn't working as normal but lets do it!

PR #224, opened 2023-12-27
feat: Bundle analysis notify task

scott-codecov:
This is a separate task for now and results in a separate PR comment.  Ideally this would be part of the existing notify task but there's some significant refactoring work that needs to be done first to support that.

This new task runs at the end of the bundle analysis processor chain and posts a PR comment with information regarding the bundle analysis comparison across base/head commits.

Depends on the migration in https://github.com/codecov/codecov-api/pull/320

### DIFF
--- a/services/bundle_analysis.py
+++ b/services/bundle_analysis.py
@@ -2,19 +2,32 @@ import logging
 import os
 import tempfile
 from dataclasses import dataclass
+from functools import cached_property
 from typing import Any, Dict, Optional
 
 import sentry_sdk
-from shared.bundle_analysis import BundleAnalysisReport, BundleAnalysisReportLoader
+from shared.bundle_analysis import (
+    BundleAnalysisComparison,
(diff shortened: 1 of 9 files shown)


services/bundle_analysis.py
```diff
@@ -142,3 +155,227 @@ def process_upload(self, upload: Upload) -> ProcessingResult:
...
+            kilobytes = round(kilobytes, 2)
+            return f"{kilobytes}kb"
+
+        megabytes = kilobytes / 1000
+        if megabytes < 1000:
+            megabytes = round(megabytes, 2)
+            return f"{megabytes}mb"
+
+        gigabytes = round(megabytes / 1000, 2)
+        return f"{gigabytes}gb"
```
joseph-sentry, line 381: i think it may be better to upper case the characters for the units in this function

joseph-sentry (approved):

PR #226, opened 2024-01-02
feat: some per-customer ats metrics

giovanni-guidini:
context: codecov/engineering-team#526

Adds some basic per-customer ats metrics.
Actually because the `MetricsContext` usage is disabled this
just sets up the metrics once that's done.

Makes use of `MetricsContext` to annotate metrics per customer
(repo, commit, owner).

### Legal Boilerplate

### DIFF
--- a/tasks/label_analysis.py
+++ b/tasks/label_analysis.py
@@ -71,6 +71,15 @@ class LabelAnalysisRequestProcessingTask(
     dbsession: Session = None
     metrics_context: MetricContext = None
 
+    def reset_task_context(self):
+        """Resets the task's attributes to None to avoid spilling information
+        between task calls in the same process.
+        https://docs.celeryq.dev/en/latest/userguide/tasks.html#instantiation
+        """
+        self.errors = None
+        self.dbsession = None
+        self.metrics_context = None
+
     async def run_async(self, db_session, request_id, *args, **kwargs):
         self.errors = []
         self.dbsession = db_session
@@ -90,7 +99,7 @@ class LabelAnalysisRequestProcessingTask(
                 error_msg="LabelAnalysisRequest not found",
                 error_extra=dict(),
             )
-            return {
+            response = {
                 "success": False,
                 "present_report_labels": [],
                 "present_diff_labels": [],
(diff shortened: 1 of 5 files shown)


tasks/label_analysis.py
```diff
@@ -101,6 +106,10 @@ async def run_async(self, db_session, request_id, *args, **kwargs):
                 commit=label_analysis_request.head_commit.commitid,
             ),
         )
+        self.metrics_context = MetricContext(
```
matt-codecov, line 109: i don't think celery wants us to store per-request state on the task like this https://docs.celeryq.dev/en/latest/userguide/tasks.html#instantiation. we probably should either recreate the context or pass it around as a function argument, though i don't really like that :P

tasks/static_analysis_suite_check.py
```diff
@@ -45,6 +49,9 @@ async def run_async(
                 == StaticAnalysisSingleFileSnapshotState.CREATED.db_id,
             )
         )
+        metrics_context = MetricContext(
+            repo_id=suite.commit.repository.repoid, commit_id=suite.commit.id
```
matt-codecov, line 53: i am probably overoptimizing, but if you have a repository db model, you can pass in `repo.owner.ownerid` so `MetricContext` doesn't have to query it later

maybe i should write a factory function that takes db models and prepopulates with them

tasks/label_analysis.py
```diff
@@ -410,17 +419,39 @@ def calculate_final_result(
                 commit=commit_sha,
             ),
         )
+        self.metrics_context.log_simple_metric(
+            "label_analysis.tests_saved_count", len(all_report_labels)
+        )
+        self.metrics_context.log_simple_metric(
+            "label_analysis.requests_with_requested_labels",
+            float(requested_labels is not None),
+        )
```
matt-codecov, line 428: i think you should still be able to use `attempt_log_simple_metric()` here despite this not being in an async function? as long as you're running inside an asyncio loop it should let you. i guess if you're invoking this from a synchronous test it will get mad that there isn't a runtime

matt-codecov (approved):
excited!!!!! only one actual comment, the rest is just me thinking out loud

tasks/label_analysis.py
giovanni-guidini: Oh that is a very good observation.

What do you mean by "recreate the context"? We are doing that, no? Since the entrypoint for the task is `run_async` and we create a new context every time.

matt-codecov: recreate in each function where we need it, i mean

PR #229, opened 2024-01-05
add temporary whitelisted org to test future ghm flow

adrian-codecov:
For the longest time, we haven't been charging github marketplace customers the proper way. We intend to notify GHM customers in the near future about the change, but in the meantime we're wanting to test with whitelist org. 

- I'm adding a conditional in the decoration details to bypass any customer's that fall into the whitelisted org

### Legal Boilerplate


services/tests/test_decoration.py
```diff
@@ -376,6 +376,83 @@ def test_get_decoration_type_not_pr_plan(self, dbsession, mocker, enriched_pull)
         assert decoration_details.reason == "Org not on PR plan"
         assert decoration_details.should_attempt_author_auto_activation is False
 
+    def test_get_decoration_type_users_plan(self, dbsession, mocker, enriched_pull):
```
adrian-codecov, line 379: This is a temporary test, this was the least effort way to replicate the logic for enriched pulls

joseph-sentry (approved):

joseph-sentry:
maybe add a comment above the whitelisted_orgs if statement briefly explaining why it's there

PR #234, opened 2024-01-16
Add test results finisher

joseph-sentry:
Depends on: #197 

## Summary
This PR adds functionality for the test results PR comment feature, and contains the necessary improvements to the test result processor task to achieve this.

The first change to the processor is that it's no longer writing to the db anymore. This is because the processor tasks run concurrently, and concurrent writes to the DB were making it difficult to ensure that there multiple Test objects weren't being created at the same time. The writes to the DB to persist `Test` and `TestInstance` objects were moved to the finisher. 

The second change to the processor is that it now returns a processed list of "Testrun dicts", which contains some metadata (`env` and `run_number`) and a list of "testruns" which the finisher will use to decide what `Test` and `TestInstance` objects to create.

The `TestResultsFinisher` task is created in this PR. This task is responsible for gathering the results of the processing step, adding new `Test` and `TestInstance` objects, updating existing `TestInstance` objects that need to be updated, and finally either writing the test results PR comment or triggering the notify task to write the coverage PR comment.

The upload task was updated to use a chord to run the processor tasks concurrently and to pass their results to the finisher task that will run when all the processor tasks are completed.

The notify task now checks if there are any failed tests that exist on a report attached to the commit it is trying to notify for, if there exist failed tests, it will no longer notify.

### CI Re-run 
This is the edge case where a user might not make any changes to the code, but re-runs the CI. This may happen if there is a flaky test that sometimes fails. We must handle this because when this happens we want to update the outcomes of the tests that ran previously.  The `run_number` metadata field that the processor passes to the finisher is used to handle this edge case. The `run_number` corresponds to the `build_code` in the upload. This helps us distinguish between different runs of the same workflow. The `run_number` tells the finisher that if a `TestInstance` exists on the current commit and maps to the same `Test` as the one we are currently examining, we should look at the `run_number` to determine which one should be used. The notify task being called if there are no failures in the test result finisher, is also part of the solution to this edge case, because of the case where during the re run, the coverage notifier runs before the test result finisher 

### Same test, different env
This is a case where a user may be running the same test in the same CI run multiple times under a different environment eg. running a tests under multiple python versions. We need to be able to distinguish between these tests. We do this by adding the `env` field to the `Test` object which comprises a hash of the `flag_names` and the `job_code` of upload where the `Test` was first created. The uniqueness constraint on `Test` objects includes the `testsuite`, `name`, `repoid` and `env` fields which means there can only be one `Test` with the same name, testsuite and env for each repo.  
Subsequent uploads that contain the same test being run with the same env will map to the existing `Test` object in the DB.

## Possible improvements
- Checkpoint logger integration
- Addition of metrics on top of the checkpoint logger
- Streaming parsing improvements mentioned here: https://github.com/codecov/worker/pull/197#pullrequestreview-1812283749
- Adding a roll up of the test result data for a given CommitReport (similar to `ReportLevelTotals`)
- Moving `TestInstance` data to GCS in sqlite files

### DIFF
--- a/tasks/notify.py
+++ b/tasks/notify.py
@@ -15,7 +15,7 @@ from sqlalchemy.orm.session import Session
 from test_results_parser import Outcome
 
 from app import celery_app
-from database.enums import CommitErrorTypes, Decoration
+from database.enums import CommitErrorTypes, Decoration, ReportType
 from database.models import Commit, Pull
 from database.models.core import GITHUB_APP_INSTALLATION_DEFAULT_NAME
 from helpers.checkpoint_logger import from_kwargs as checkpoints_from_kwargs
@@ -37,6 +37,7 @@ from services.repository import (
     fetch_and_update_pull_request_information_from_commit,
     get_repo_provider_service,
 )
+from services.test_results import latest_test_instances_for_a_given_commit
 from services.yaml import get_current_yaml, read_yaml_field
 from tasks.base import BaseCodecovTask
 from tasks.upload_processor import UPLOAD_PROCESSING_LOCK_NAME
@@ -141,21 +142,12 @@ class NotifyTask(BaseCodecovTask, name=notify_task_name):
         commit = commits_query.first()
 
         # check if there were any test failures
-        latest_test_instances = (
-            db_session.query(TestInstance)
-            .join(Upload)
-            .join(CommitReport)
-            .filter(
-                CommitReport.commit_id == commit.id_,
-            )
-            .order_by(TestInstance.test_id)
-            .order_by(desc(Upload.created_at))
-            .distinct(TestInstance.test_id)
-            .all()
+        latest_test_instances = latest_test_instances_for_a_given_commit(
+            db_session, commit.id_
         )
 
         if any(
-            [test.outcome == int(Outcome.Failure) for test in latest_test_instances]
+            [test.outcome == str(Outcome.Failure) for test in latest_test_instances]
         ):
             return {
                 "notify_attempted": False,
--- a/tasks/notify.py
+++ b/tasks/notify.py
@@ -10,7 +10,9 @@ from shared.celery_config import (
 from shared.reports.readonly import ReadOnlyReport
 from shared.torngit.exceptions import TorngitClientError, TorngitServerFailureError
 from shared.yaml import UserYaml
+from sqlalchemy import desc
 from sqlalchemy.orm.session import Session
+from test_results_parser import Outcome
 
 from app import celery_app
 from database.enums import CommitErrorTypes, Decoration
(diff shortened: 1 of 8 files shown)


matt-codecov (commented):
started a group DM about a design tweak for performance reasons. need to sync on that before accept/request changes decision

tasks/notify.py
```diff
@@ -130,6 +138,30 @@ async def run_async_within_lock(
...
+
+        # check if there were any test failures
+        latest_test_instances = (
+            db_session.query(TestInstance)
+            .join(Upload)
+            .join(CommitReport)
+            .filter(
+                CommitReport.commit_id == commit.id_,
+            )
+            .order_by(TestInstance.test_id)
```
matt-codecov, line 150: any particular reason?

tasks/notify.py
```diff
@@ -130,6 +138,30 @@ async def run_async_within_lock(
...
+            .order_by(desc(Upload.created_at))
+            .distinct(TestInstance.test_id)
+            .all()
+        )
+
+        if any(
+            [test.outcome == int(Outcome.Failure) for test in latest_test_instances]
```
matt-codecov, line 157: i think `test.outcome` is a string now, is the int cast correct here?

tasks/test_results_finisher.py
```diff
@@ -0,0 +1,153 @@
...
+        test_instances = (
+            db_session.query(TestInstance)
+            .join(Upload)
+            .join(CommitReport)
+            .join(Commit)
+            .filter(Commit.id_ == commit.id_)
+            .order_by(TestInstance.test_id)
+            .order_by(desc(Upload.created_at))
+            .distinct(TestInstance.test_id)
+            .all()
```
matt-codecov, line 109: consider moving this query to a helper function on `TestInstance` or something since you use it in more than one place so that the logic definitely stays the same

tasks/test_results_finisher.py
```diff
@@ -0,0 +1,153 @@
...
+        if self.check_if_no_failures(test_instances):
+            self.app.tasks[notify_task_name].apply_async(
+                args=None,
+                kwargs=dict(
+                    repoid=repoid, commitid=commitid, current_yaml=commit_yaml.to_dict()
+                ),
+            )
+            return {"notify_attempted": False, "notify_succeeded": False}
+
+        success = None
```
matt-codecov, line 121: why is this here if it's overwritten two lines down?

tasks/test_results_finisher.py
```diff
@@ -0,0 +1,153 @@
...
+            (
+                testrun_list["successful"] is False
+                for result in previous_result
+                for testrun_list in result
+            )
+        )
+
+    def check_if_no_failures(self, testrun_list):
+        return all(
+            [instance.outcome != int(Outcome.Failure) for instance in testrun_list]
```
matt-codecov, line 148: same Q re `instance.outcome` being a string and `int()` maybe not being correct

matt-codecov (commented):
thanks for all the iteration on these changes, i think we will long-term be happy we took the time. just nits left, some old comments + a couple new ones

tasks/notify.py
joseph-sentry: it's how the sql works, because I'm using `distinct on` on the `test_id` to select only once test instance for each `test_id`.  `distinct on` requires the left most ordered field to be the same as the field we are `distinct on`: https://www.postgresql.org/docs/9.0/sql-select.html#SQL-DISTINCT

PR #236, opened 2024-01-19
feat: GithubAppInstallation model

giovanni-guidini:
These changes port https://github.com/codecov/codecov-api/pull/346 to the worker so we can use them here as well.

context: codecov/engineering-team#970


database/models/core.py
```diff
@@ -158,6 +169,49 @@ def __repr__(self):
         return f"Repo<{self.repoid}>"
 
 
+GITHUB_APP_INSTALLATION_DEFAULT_NAME = "codecov_app_installation"
+
+
+class GithubAppInstallation(CodecovBaseModel, MixinBaseClass):
```
adrian-codecov, line 175: Is there an easy way to test this E2E to prevent something like what happened in API?

adrian-codecov (approved):

database/models/core.py
giovanni-guidini: Unsure. I don't think we have the infra in place...

I tested locally, if that's worth anything (upload, PR comment, statuses and exploring files in the UI), but it's very much a "trust me" scenario and we know it's easy to make mistakes.

PR #239, opened 2024-01-23
feat: use GithubAppInstalltion in the worker

giovanni-guidini:
depends on: https://github.com/codecov/worker/pull/236

context: codecov/engineering-team#970

usage of the `GithubAppInstallation` model through the worker.
It still respects and doesn't affect deprecated usages of
`owner.integration_id` and `repo.using_integration`, but if a
ghapp exists for the owner it takes precedence over the legacy fields.

### Legal Boilerplate


tasks/notify.py
```diff
@@ -179,7 +180,24 @@ async def run_async_within_lock(
                 "Not sending notifications yet because we are waiting for CI to finish",
                 extra=dict(repoid=commit.repoid, commit=commit.commitid),
             )
-            if commit.repository.using_integration or commit.repository.hookid:
+            ghapp_default_installations = list(
```
adrian-codecov, line 183: I see this affects the notify task for coverage uploads. Would it be tough to add the GH App integration for the bundle analysis notifier?  https://github.com/codecov/worker/blob/c248c6f78f6ac8d4ef954ea49f394241da46b060/services/bundle_analysis.py#L268

adrian-codecov: (If unfamiliar, BA uses a different notifier cause the normal notifier relied too much on coverage data being present, hence Scott opted to do a separate notifier wrapper class)

services/bots.py
```diff
@@ -13,15 +14,46 @@
...
+        if repository:
+            if app_installation.is_repo_covered_by_integration(repository):
+                return app_installation.installation_id
+            # The repo we want to get a token for is not covered by the installation
+            return None
+        else:
+            # Getting owner installation - not tied to any particular repo
+            return app_installation.installation_id
+    # DEPRECATED FLOW - begin
+    if owner.integration_id and deprecated_using_integration:
```
adrian-codecov, line 39: Could this code run? How do customers with their integration id tied to their owner start using the new GH App class? Would that be "auto-populated" after syncing?

services/bots.py
```diff
@@ -13,15 +14,46 @@
...
+    *,
+    repository: Optional[Repository] = None,
+    installation_name: str = GITHUB_APP_INSTALLATION_DEFAULT_NAME,
+) -> Optional[int]:
+    default_app_installation_filter = filter(
+        lambda obj: obj.name == installation_name, owner.github_app_installations or []
+    )
+    # filter is an Iterator, so we need to scan matches
+    # in practice we only consider the 1st one
+    for app_installation in default_app_installation_filter:
```
joseph-sentry, line 29: will this always contain one or zero installations? or is there a possibility of there being multiple?

giovanni-guidini: Currently there will always be 0 or 1 installations I believe. That comes from Github when the user installs the codecov gh app to their profile. I don't think it's possible for a user in GH to have multiple installations of the same app in their profile.

Assume in the future there are many such installations for some reason.
The idea of `installation_name` is to identify app's functionalities in such a way that they are interchangeable.
Although we'd probably want to be able to pass a list of names and select one option randomly... but that's for the future and there might be other implementation choices available.

services/bots.py
giovanni-guidini: Yes, the code can run if there's no `GithubAppInstallation` object for a given owner.

The `GH App` objects are populated by the API when we receive the installation webhook.
I don't think there are plans for backfilling this info, so it would only be filled for an existing user if they make changes to their installation.

tasks/notify.py
giovanni-guidini: When making these changes I looked for references of `.using_integration`. There was no such ref in the bundle analysis notifier.

I guess it was decided that that task doesn't need to check this for the timeout on the retry. Personally I think that this implementation here is a bit "too much", but was respecting the logic that's in place.

adrian-codecov (approved):

PR #240, opened 2024-01-23
deps: update cryptography package version

joseph-sentry:
Fixes: https://github.com/codecov/internal-issues/issues/159


matt-codecov (commented):
you updated `shared` with this same version constraint in https://github.com/codecov/shared/pull/106. i think if you update `shared` here you should get this constraint without needing to hardcode it in `requirements.in`? same goes for `codecov-api` which has its own issue: https://github.com/codecov/internal-issues/issues/178

matt-codecov (approved):

PR #242, opened 2024-01-26
Attempt to fix adjust base behavior rounding bug

joseph-sentry:
This PR creates a log in the `_apply_adjust_base_behavior` function and attempts to fix the following bug.

There's a bug where we compare the base adjusted coverage to the
head coverage. The issue occurs when the base adjusted coverage has more
digits after the decimal point than the head coverage.

The head coverage that we fetch in this function will always have 5
digits after the decimal, but the base adjusted coverage does not have
that same restriction.

The decimal library will not round the number with more digits after
the decimal point when doing a comparison.

Currently, in our code we are basically doing the following:

`Decimal('90.001').quantize('0.00') <= Decimal('90.001')`

which will evaluate to false.

If we are rounding the head coverage to 5 digits after the decimal point
we must do the same for the base adjusted coverage for it to be a fair
comparison.

### DIFF
--- a/services/notification/notifiers/mixins/status.py
+++ b/services/notification/notifiers/mixins/status.py
@@ -207,7 +207,13 @@ class StatusProjectMixin(object):
                 partials_removed=partials_removed,
             ),
         )
-        if base_adjusted_coverage <= head_coverage:
+        # the head coverage is rounded to five digits after the dot, using shared.helpers.numeric.ratio
+        # so we should round the base adjusted coverage to the same amount of digits after the dot
+        # Decimal.quantize: https://docs.python.org/3/library/decimal.html#decimal.Decimal.quantize
+        quantized_base_adjusted_coverage = base_adjusted_coverage.quantize(
+            Decimal("0.00000")
+        )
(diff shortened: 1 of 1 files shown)


services/notification/notifiers/mixins/status.py
```diff
@@ -196,7 +196,24 @@ def get_sum_from_lists(
...
+                base_adjusted_coverage=base_adjusted_coverage,
+                head_coverage=head_coverage,
+                hits_removed=hits_removed,
+                misses_removed=misses_removed,
+                partials_removed=partials_removed,
+            ),
+        )
+        # the head coverage is rounded to five digits after the dot, using shared.helpers.numeric.ratio
+        # so we should round the base adjusted coverage to the same amount of digits after the dot
+        # Decimal.quantize: https://docs.python.org/3/library/decimal.html#decimal.Decimal.quantize
```
giovanni-guidini, line 212: Very nice comment. Thanks for adding it here.

giovanni-guidini (approved):
LGTM

PR #244, opened 2024-01-29
965 add languages during repo sync

adrian-codecov:
This PR adjusts the list of languages that need to be adjusted. This change makes the language sync applicable to every repository that was synced for that organization.

- Adds correct list of repoids
- Adjusts case for null repoids list

### Legal Boilerplate


tasks/sync_repo_languages.py
```diff
@@ -65,9 +66,15 @@ async def run_async(
         except TorngitError:
             log.warning(
                 "Unable to find languages for this repository",
-                dict(repository_id=repository.repoid),
+                extra=dict(repository_id=repository.repoid),
```
joseph-sentry, line 69: [nit] i would have the key in the extra dict be repoid here as well

tasks/sync_repos.py
```diff
@@ -73,20 +73,20 @@ async def run_async(
...
                     with metrics.timer(f"{metrics_scope}.sync_repos"):
-                        await self.sync_repos(
+                        synced_repoids = await self.sync_repos(
                             db_session, git, owner, username, using_integration
                         )
 
                 self.sync_repos_languages(
-                    ownerid=owner.ownerid,
-                    db_session=db_session,
+                    synced_repoids=synced_repoids or [],
```
joseph-sentry, line 89: [nit] since you're setting synced_repoids above to `[]` i think the `or []` is not needed

JerrySentry (approved):

tasks/sync_repo_languages.py
adrian-codecov: Can do!

tasks/sync_repos.py
adrian-codecov: I added it here cause the sync repos fn has an explicit `return`, returning None, that I didn't want to change to `return []` so I rather added it here. I did get an error if I didn't supply a non-None value

PR #246, opened 2024-01-30
refactor: refactor FlushRepoTask

giovanni-guidini:
These changes refactor the FlushRepoTask for 2 reasons:

1. Progress reports through the task.
    logging in the task was virtually inexistent short of the "task called" one
    but there was no indication on the progress of the task
2. Performance metrics through sentry_sdk.trace
    this will allow us to understand where the bottlenecks of this task are and
    provide better insight for improving it

### Legal Boilerplate

### DIFF
--- a/tasks/flush_repo.py
+++ b/tasks/flush_repo.py
@@ -1,8 +1,11 @@
 import logging
+from dataclasses import dataclass
+from typing import Optional
 
-from sqlalchemy import null
+import sentry_sdk
 
 from app import celery_app
+from database.engine import Session
 from database.models import (
     Branch,
     Commit,
@@ -32,16 +35,28 @@ from tasks.base import BaseCodecovTask
 log = logging.getLogger(__name__)
 
 
+@dataclass
+class FlushRepoTaskReturnType(object):
+    error: Optional[str] = None
+    deleted_commits_count: int = 0
+    delete_branches_count: int = 0
+    deleted_pulls_count: int = 0
+    deleted_archives_count: int = 0
+
+
 class FlushRepoTask(BaseCodecovTask, name="app.tasks.flush_repo.FlushRepo"):
-    async def run_async(self, db_session, *, repoid: int, **kwargs):
-        log.info("Deleting repo contents", extra=dict(repoid=repoid))
-        repo = db_session.query(Repository).filter_by(repoid=repoid).first()
+    @sentry_sdk.trace
+    def _delete_archive(self, repo: Repository) -> int:
         archive_service = ArchiveService(repo)
         deleted_archives = archive_service.delete_repo_files()
+        log.info(
+            "Deleted archives from storage",
+            extra=dict(repoid=repo.repoid, deleted_archives_count=deleted_archives),
+        )
+        return deleted_archives
 
-        commit_ids = db_session.query(Commit.id_).filter_by(repoid=repoid)
-
-        # comparisons
+    @sentry_sdk.trace
+    def _delete_comparisons(self, db_session: Session, commit_ids, repoid: int) -> None:
         commit_comparison_ids = db_session.query(CompareCommit.id_).filter(
             CompareCommit.base_commit_id.in_(commit_ids)
             | CompareCommit.compare_commit_id.in_(commit_ids)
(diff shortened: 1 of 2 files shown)


tasks/flush_repo.py
```diff
@@ -32,16 +35,28 @@
...
+
+
 class FlushRepoTask(BaseCodecovTask, name="app.tasks.flush_repo.FlushRepo"):
-    async def run_async(self, db_session, *, repoid: int, **kwargs):
-        log.info("Deleting repo contents", extra=dict(repoid=repoid))
-        repo = db_session.query(Repository).filter_by(repoid=repoid).first()
+    @sentry_sdk.trace
+    def _delete_archive(self, repo: Repository) -> int:
         archive_service = ArchiveService(repo)
         deleted_archives = archive_service.delete_repo_files()
```
adrian-codecov, line 51: Should this fn be async? I suppose we don't necessarily care to wait for the files to be deleted, but we don't know for certain if the call succeeded otherwise right?

tasks/flush_repo.py
```diff
@@ -55,11 +70,10 @@ async def run_async(self, db_session, *, repoid: int, **kwargs):
...
         ).delete(synchronize_session=False)
         db_session.commit()
+        log.info("Deleted comparisons", extra=dict(repoid=repoid))
 
-        # reports
-        report_ids = db_session.query(CommitReport.id_).filter(
-            CommitReport.commit_id.in_(commit_ids)
-        )
+    @sentry_sdk.trace
+    def _delete_reports(self, db_session: Session, report_ids, repoid: int):
```
adrian-codecov, line 76: Ty for all these helper methods, a lot more organized + easy to read

tasks/flush_repo.py
```diff
@@ -84,62 +101,113 @@ async def run_async(self, db_session, *, repoid: int, **kwargs):
...
+        with sentry_sdk.start_span("query_report_ids"):
+            report_ids = db_session.query(CommitReport.id_).filter(
+                CommitReport.commit_id.in_(commit_ids)
+            )
+        self._delete_reports(db_session, report_ids, repoid)
+        self._delete_uploads(db_session, report_ids, repoid)
+
+        self._delete_commit_details(db_session, commit_ids, repoid)
+
+        # TODO: Component comparison
```
adrian-codecov, line 197: Would we still leave the commit and component comparisons existing in the DB? I think since commit comparisons rely on commit ids existing, that the delete_commits fn below would fail wouldnt it?

adrian-codecov (commented):
Some comments to take a peek at, hmu when it's ready for a 2nd review!

tasks/flush_repo.py
adrian-codecov: Yea agree, not part of this scope, just found it interesting that the archive service wasn't aysnc in the first place, makes sense

tasks/flush_repo.py
giovanni-guidini: Yes this is true. It's not the only piece of data that we should be removing but aren't (`LabelAnalysisRequestProcessingError` has the same issue).

But I wanted to be very intentional with these changes and limit them to a _refactor_.
This for a couple of reasons that include:
1. Checking all the missing pieces of data, adding them in and adding tests for all of that consumes a lot more time.
2. These changes are meant to give more information so we can make better decisions on how to improve the feature overall. We might come to the conclusion that we need to tear the whole thing apart and start from scratch. In this case the extra effort I'd have to put to add the missing details would be lost.
3. Currently there's no "epic" or "plan" on how to tackle this flow. This needs to be discussed and prioritized. I personally think that this knowledge that some data is not being deleted when it should needs to factor in on the "how big this project is gonna be?"

So while you are absolutely right and we should (and will) fix it I hope we can ignore the fact that this is partially broken for now. It is one of the reasons I left the comment there.
Also [this comment](https://github.com/codecov/engineering-team/issues/1031#issuecomment-1917575228) captures some of the data that we know is missing, in case you're worried it'll be lost.

adrian-codecov: Yeah big time, didn't mean to imply we needed to add those as yeah this should be contained to a refactor, was mostly trying to see if what I said made sense, still learning the ropes here 😅. And ty writing the things we're missing somewhere too 👌

adrian-codecov (approved):

PR #247, opened 2024-01-31
feat: adjust repoids for orgs using app integration

adrian-codecov:
This PR is to adjust retrieving repoids for customers using app integration. It previously was returning a list of repos instead of the repoids themselves. This PR adds that functionality + tests

### Legal Boilerplate


tasks/sync_repos.py
```diff
@@ -179,7 +181,7 @@ def process_repos(repos):
             extra=dict(repoids=total_missing_repos),
         )
 
-        return total_missing_repos
```
matt-codecov, line 182: is this variable still used at all?

matt-codecov (approved):
looks fine but why do we want to do this?

tasks/sync_repos.py
adrian-codecov: `total_missing_repos` was there prior to my implementation, I only see it being used in the logs atm, and it's different from reposid as it returns a list of dictionaries that looks like this `[{owner: {...}. repo: {...}}, {...}]`

The need to do this in the first place is that we want to sync languages for a bunch of repositories for Bundle Analysis. I was previously doing repos from a specific owner, but that missed repos from multiple owners. The sync_repos functions, with or without integration, already capture all the repos from N owners we're syncing,, so it made sense to pass those along to the syncing function rather than figuring out all the repos/owners affected through a query

PR #251, opened 2024-02-01
Improve test result PR comment

joseph-sentry:
This changes the PR comment to organize rows such that tests with the same failure message will be displayed in the same row. Also, if a test has multiple flags with the same failure message, it will display them in the same line.

### DIFF
--- a/services/test_results.py
+++ b/services/test_results.py
@@ -218,8 +218,13 @@ class TestResultsNotifier:
             "| {0} | <pre>{1}</pre> |".format(
                 (
                     "<br>".join(
-                        self.insert_breaks(f"{test_name}[{','.join(test_env_list)}]")
-                        for test_name, test_env_list in failed_test_to_env_list.items()
+                        self.insert_breaks(
+                            f"{test_name}[{','.join(sorted(test_env_list))}]"
+                        )
+                        for test_name, test_env_list in sorted(
+                            failed_test_to_env_list.items(),
+                            key=lambda failure: failure[0],
+                        )
                     )
                 ),
                 failure_message.replace("\n", "<br>"),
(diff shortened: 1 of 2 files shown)


joseph-sentry:
> @joseph-sentry , do you have a sample screenshot of how this would look?



image

services/test_results.py
```diff
@@ -200,14 +200,23 @@ def build_message(self, url, test_instances):
...
+                            key=lambda failure: failure[0],
+                        )
+                    )
+                ),
                 failure_message.replace("\n", "<br>"),
             )
-            for failed_test_name, failure_message in sorted(
-                failures.items(), key=lambda failure: failure[0]
+            for failure_message, failed_test_to_env_list in sorted(
+                failures.items(), key=lambda x: x[0]
```
adrian-codecov, line 219: Might be missing something but couldn't you use `failures.keys()` and `x` instead of x[0]? same for the loop above

adrian-codecov (commented):
Otherwise lgtm

adrian-codecov (approved):

services/test_results.py
joseph-sentry: do you mean to do something like: 
```
self.insert_breaks(
  f"{test_name}[{','.join(sorted(failed_test_to_env_list[test_name]))}]"
)
for test_name in sorted(
  failed_test_to_env_list.keys(),
)

adrian-codecov: Rereading this now, your approach is fine, I was under the impression you only needed the keys rather than the key/value, so I was suggesting to use only the keys to make it more readable

adrian-codecov (approved):

PR #252, opened 2024-02-05
fix(upload): don't use checkpoints logger unless report type is coverage

joseph-sentry:
Previously, checkpoints was being used when it wasn't necessarily defined in the Upload task

### DIFF
--- a/tasks/upload.py
+++ b/tasks/upload.py
@@ -391,7 +391,7 @@ class UploadTask(BaseCodecovTask, name=upload_task_name):
         report_code = upload_context.report_code
 
         checkpoints = None
-        if report_type == "coverage":
+        if report_type == ReportType.COVERAGE:
             try:
                 checkpoints = checkpoints_from_kwargs(UploadFlow, kwargs)
                 checkpoints.log(UploadFlow.PROCESSING_BEGIN)
--- a/tasks/upload.py
+++ b/tasks/upload.py
@@ -305,7 +305,8 @@ class UploadTask(BaseCodecovTask, name=upload_task_name):
                     "Not retrying since there are likely no jobs that need scheduling",
                     extra=dict(commit=commitid, repoid=repoid, report_type=report_type),
                 )
-                checkpoints.log(UploadFlow.NO_PENDING_JOBS)
+                if checkpoints:
+                    checkpoints.log(UploadFlow.NO_PENDING_JOBS)
                 return {
                     "was_setup": False,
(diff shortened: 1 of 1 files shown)


tasks/upload.py
```diff
@@ -383,17 +385,21 @@ async def run_async_within_lock(
...
-        except ValueError as e:
-            log.warning(f"CheckpointLogger failed to log/submit", extra=dict(error=e))
-
         repoid = upload_context.repoid
         commitid = upload_context.commitid
         report_type = upload_context.report_type
         report_code = upload_context.report_code
 
+        checkpoints = None
+        if report_type == ReportType.COVERAGE:
```
giovanni-guidini, line 394: some line above you used `ReportType.COVERAGE.value`.
Please make sure that this comparison is accurate.

tasks/upload.py
```diff
@@ -580,7 +587,7 @@ def schedule_task(
             )
         elif commit_report.report_type == ReportType.TEST_RESULTS.value:
             res = self._schedule_test_results_processing_task(
```
giovanni-guidini, line 589: [nit] add TODO here to include flows specific of Test_RESULTS

tasks/upload.py
```diff
@@ -712,10 +718,6 @@ def _schedule_test_results_processing_task(
                 )
                 processor_task_group.append(sig)
         if processor_task_group:
-            checkpoint_data = None
-            if checkpoints:
-                checkpoints.log(UploadFlow.INITIAL_PROCESSING_COMPLETE)
```
giovanni-guidini, line 717: I don't see initial processing complete anymore... has it been moved to another task or are we not using it?

I'm not comfortable approving the PR without knowing why this has been removed. So I will defer approval to someone that knows better

joseph-sentry: this is happening in the test_results scheduling and that's why I removed it, since I don't initialize the `checkpoints` anyways when the report type is `test_results`

tasks/upload.py
joseph-sentry: I see why this is confusing. In this line, we are using the `report_type` from the `UploadContext` which is the enum itself. In the line above that you are referring to, we are using the `report_type` passed as an arg to the task, which is a `str`

tasks/upload.py
giovanni-guidini: I see. Thanks for the update here.

giovanni-guidini (approved):

PR #256, opened 2024-02-07
feat: create and use save_commit_measurement task

adrian-codecov:
We're creating a `save_commit_measurement` task to replace any calls to the `save_commit_measurement` function, as specified by Dana here, https://github.com/codecov/engineering-team/issues/682. This PR:

- Adds new Shared version
- Adds new `SaveCommitMeasurementTask`, which basically calls the `save_commit_measurment` function in `services.timeseries`
- Adds tests for this ^ task
- Calls task in `upload_finisher` and `timeseries_backfill` tasks
- Adjusts the tests for these ^ fns

### Legal Boilerplate


tasks/tests/integration/test_timeseries_backfill.py
```diff
@@ -43,36 +51,46 @@ async def test_backfill_dataset_run_async(dbsession, mocker, mock_storage):
...
     assert res == {"successful": True}
-
-    coverage_measurement = (
-        dbsession.query(Measurement)
-        .filter_by(
-            name=MeasurementName.coverage.value,
-            commit_sha=commit.commitid,
-        )
-        .one_or_none()
+    mocked_app.tasks[
```
adrian-codecov, line 64: I'm unsure if this makes sense here since this is an integration test. We used to call the function and wait for it here, https://github.com/codecov/worker/pull/256/files#diff-902a7b3fa8f6ee724af5423139f45596d96edeb267f72ab8aa8cd47a441c467dR38, but now we're dispatching tasks to do the logic, so the previous test logic I removed didn't await for the changes to take place. Would it make sense here to await for each task to finish then? (don't think so), or is this test adjustment good?

tasks/save_commit_measurements.py
```diff
@@ -0,0 +1,66 @@
...
+
+        if commit is None:
+            return {"successful": False, "error": "no_commit_in_db"}
+
+        try:
+            # TODO: We should improve on the error handling/logs inside this fn
+            save_commit_measurements(commit=commit, dataset_names=dataset_names)
+            return {"successful": True}
+        except Exception as e:
+            log.error(
```
giovanni-guidini, line 51: [nit] If you use `log.exception` we get the stack trace. It's really useful for debugging.
And you don't have to add `error=e` in the extra.

I recommend you change this.

tasks/tests/unit/test_save_commit_measurements.py
```diff
@@ -0,0 +1,45 @@
+import pytest
+
+from database.tests.factories.core import CommitFactory, OwnerFactory, RepositoryFactory
+from tasks.save_commit_measurements import SaveCommitMeasurementsTask
+
+
+class TestSaveCommitMeasurements(object):
```
giovanni-guidini, line 7: [nit] Add a test for the failure too, where the task raises some exception

giovanni-guidini (commented):
Changes itself look fine, but please check the comments.

The ticket also hints that timescale fails with some frequency. In this case this task would just fail to register the measurements. How important are they? Shouldn't we have some simple retry mechanism for this task for the case that timescale failures are temporary?

tasks/tests/unit/test_save_commit_measurements.py
adrian-codecov: Ah yes, I missed this 🤦, ty

tasks/save_commit_measurements.py
adrian-codecov: Thanks about this! This type of advice is really useful as I'm onboarding :)

Any reason why the error doesn't log a stack trace?

giovanni-guidini: Idk... probably some design decision from the devs of the log library we use 🤷

tasks/save_commit_measurements.py
```diff
@@ -0,0 +1,65 @@
...
+            .filter(Commit.repoid == repoid, Commit.commitid == commitid)
+            .first()
+        )
+
+        if commit is None:
+            return {"successful": False, "error": "no_commit_in_db"}
+
+        try:
+            # TODO: We should improve on the error handling/logs inside this fn
+            print("asdfasdf")
```
giovanni-guidini, line 48: sneaky print

tasks/save_commit_measurements.py
```diff
@@ -0,0 +1,65 @@
...
+            save_commit_measurements(commit=commit, dataset_names=dataset_names)
+            print("asdfasdf123")
```
giovanni-guidini, line 50: sneaky print

giovanni-guidini (approved):
LGTM. Please remove the prints before merging tho

PR #260, opened 2024-02-08
fix: remove n+1 repo flags query

joseph-sentry:
previously, we were querying the repo flags for each upload, but that was unnecessary as we can just query once and maintain a dict for the flags that we insert

### DIFF
--- a/services/report/__init__.py
+++ b/services/report/__init__.py
@@ -178,6 +178,10 @@ class BaseReportService:
 class ReportService(BaseReportService):
     metrics_prefix = "services.report"
 
+    def __init__(self, current_yaml: UserYaml):
+        super().__init__(current_yaml)
+        self.flag_dict = None
+
     def has_initialized_report(self, commit: Commit) -> bool:
         """Says whether a commit has already initialized its report or not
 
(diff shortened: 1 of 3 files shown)


services/report/__init__.py
```diff
@@ -302,25 +306,28 @@ def _attach_flags_to_upload(self, upload: Upload, flag_names: Sequence[str]):
         db_session = upload.get_db_session()
         repoid = upload.report.commit.repoid
 
-        existing_flag_dict = self.get_existing_flag_dict(db_session, repoid)
+        if self.flag_dict is None:
```
adrian-codecov, line 309: Will this not be None every time? So doing `self.flag_dict = self.fetch_repo_flags` right off the bat. Wdyt also on the fetch_repo_flags updating the flag_dict variable there rather than returning a value?

services/test_results.py
```diff
@@ -69,25 +74,28 @@ def _attach_flags_to_upload(self, upload: Upload, flag_names: Sequence[str]):
         db_session = upload.get_db_session()
         repoid = upload.report.commit.repoid
 
-        existing_flag_dict = self.get_existing_flag_dict(db_session, repoid)
+        if self.flag_dict is None:
+            self.fetch_repo_flags(db_session, repoid)
```
adrian-codecov, line 78: Comment above applies here too

services/report/__init__.py
joseph-sentry: it should only be `None` once, because if it's `None` `fetch_repo_flags` runs and updates it to be something other than `None`

adrian-codecov (approved):

PR #263, opened 2024-02-09
fix: ignore insallation for owner some times

giovanni-guidini:
This is a problem with the 'sync repos' button in the UI. It only affects getting the token for a OWNER (if it's for a repo it's correct). Sadly this affects `sync_teams` and `sync_repos` tasks.

If the owner decides to sync we should use the owner's token, even if there is an installation app for that owner. The extra logic helps us do that.

### DIFF
--- a/tasks/sync_teams.py
+++ b/tasks/sync_teams.py
@@ -24,7 +24,9 @@ class SyncTeamsTask(BaseCodecovTask, name=sync_teams_task_name):
         assert owner, "Owner not found"
         service = owner.service
 
-        git = get_owner_provider_service(owner, using_integration=False)
+        git = get_owner_provider_service(
+            owner, using_integration=False, ignore_installation=True
+        )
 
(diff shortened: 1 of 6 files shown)


tasks/sync_teams.py
```diff
@@ -24,7 +24,9 @@ async def run_async(self, db_session, ownerid, *, username=None, **kwargs):
         assert owner, "Owner not found"
         service = owner.service
 
-        git = get_owner_provider_service(owner, using_integration=False)
+        git = get_owner_provider_service(
+            owner, using_integration=False, ignore_installation=True
```
adrian-codecov, line 28: Just checking if this `ignore_installation=True` meant to be `ignore_installation=(not using_integration)`?

giovanni-guidini: yes, it's supposed to be the opposite in this case.

adrian-codecov (approved):

trent-codecov (approved):

PR #264, opened 2024-02-10
Add backfill task for github app installation 

adrian-codecov:
### Legal Boilerplate


tasks/backfill_gh_app_installations.py
```diff
@@ -0,0 +1,144 @@
...
+            )
+            return {"successful": True, "reason": "no installation needed"}
+
+        # Check if the owner exists
+        owner: Owner = (
+            db_session.query(Owner).filter_by(ownerid=ownerid, service=service).first()
+        )
+
+        if not owner:
+            log.exception(
```
joseph-sentry, line 77: I think we should use `log.error` instead of `log.exception` it seems `exception` is reserved for use from within an `except` clause.

adrian-codecov: Gotchaa, I got some feedback before from Gio to prefer error from exception in this PR, https://github.com/codecov/worker/pull/256#discussion_r1481451117. I'm still unclear when we should use each or the other, do you have a rule of thumb to know which one to use?

tasks/backfill_gh_app_installations.py
```diff
@@ -0,0 +1,144 @@
...
+from app import celery_app
+from celery_config import backfill_gh_app_installations
+from database.models.core import GithubAppInstallation, Owner, Repository
+from services.owner import get_owner_provider_service
+from tasks.base import BaseCodecovTask
+
+log = logging.getLogger(__name__)
+
+
+class BackfillGHAppInstallationsTask(
```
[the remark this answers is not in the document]
adrian-codecov: One time thing in a prod pod 👀

tasks/backfill_gh_app_installations.py
```diff
@@ -0,0 +1,149 @@
...
+            owners: List[Owner] = (
+                db_session.query(Owner).filter_by(service="github").all()
+            )
+
+        # Loop through all owner_ids
+        for owner in owners:
+            ownerid = owner.ownerid
+
+            # Check if owner any sort of integration
+            if not owner.integration_id:
```
trent-codecov, line 83: We want to still check if they have records in the new style. Unless that is impossible to happen currently? We want to sync repo ids if they have new style records regardless if this older field is set.

tasks/backfill_gh_app_installations.py
```diff
@@ -0,0 +1,149 @@
...
+                log.info("No installation work needed", extra=dict(ownerid=ownerid))
+                continue
+
+            # Check if owner has a gh app installation entry
+            gh_app_installation: GithubAppInstallation = (
```
trent-codecov, line 88: Perhaps pull this out of the loop and query for all in the list of owners you are running for.

tasks/backfill_gh_app_installations.py
```diff
@@ -0,0 +1,149 @@
...
+                    new_repo_service_ids.add(service_id)
+            log.info(
+                "Added the following repo service ids to this gh app installation",
+                extra=dict(
+                    ownerid=ownerid,
+                    installation_id=gh_app_installation.installation_id,
+                    new_repo_service_ids=list(new_repo_service_ids),
+                ),
+            )
+            gh_app_installation.repository_service_ids = list(
```
trent-codecov, line 56: I think by definition we only want the new ones. If GH doesn't tell us we can access the repo and Codecov doesn't have the repo in the db, then we don't want it in the list. I think the union with current is redundant. Thoughts?

tasks/backfill_gh_app_installations.py
```diff
@@ -0,0 +1,149 @@
...
+        db_session: Session,
+        ownerid: int,
+        owner_service: torngit.base.TorngitBaseAdapter,
+        gh_app_installation: GithubAppInstallation,
+    ):
+        repos = await owner_service.list_repos_using_installation()
+        if repos:
+            # Fetching all repos service ids we have for that owner in the DB
+            repo_service_ids_in_db = [
+                repo.service_id
```
trent-codecov, line 32: Unsure on the syntax here, will this be n+1 or does it roll it up?

trent-codecov (changes requested):
Looks good overall. One thing to change/discuss.

tasks/backfill_gh_app_installations.py
adrian-codecov: You're right, this makes sense, will change this to `if not owner.integration_id and not gh_app_installation`

tasks/backfill_gh_app_installations.py
adrian-codecov: So the question is, can anything be in current that's not in new. I think most cases would be redundant, like `{1}.union({1,2,3}) = {1,2,3}`. We could get something like `{1}.union({2,3}) = {1,2,3}`, but by definition that would mean a repo that was part of an org (or selected by the app) that is no longer, so yeah I think you'd only want the new, not new + old. I'll make a change

tasks/backfill_gh_app_installations.py
adrian-codecov: Yeah this should be 1 query, this is the same as

```
repo_list = db_session.query(Repository.service_id).filter_by(ownerid=ownerid).all() // 1 query

for repo in repo_list:
  repo_service_ids_in_db.append(repo.service_id)
```

tasks/backfill_gh_app_installations.py
```diff
@@ -0,0 +1,192 @@
...
+
+        for gh_app_installation in gh_app_installations:
+            # Check if gh app has 'all' repositories selected
+            installation_id = gh_app_installation.installation_id
+
+            owner = owners_dict[gh_app_installation.ownerid]
+            ownerid = owner.ownerid
+
+            owner_service = get_owner_provider_service(
+                owner=owner, using_integration=True
```
adrian-codecov, line 92: I had to get all the owners + ownerid cause I needed them per gh app installation to use this service

tasks/backfill_gh_app_installations.py
```diff
@@ -0,0 +1,198 @@
...
+
+    async def backfill_owners_with_integration_without_gh_app(
+        self, db_session: Session, owner_ids: List[int] = None
+    ):
+        print("backfill new apps")
+        owners_with_integration_id_without_gh_app_query = (
+            db_session.query(Owner)
+            .join(
+                GithubAppInstallation,
+                Owner.ownerid != GithubAppInstallation.ownerid,
```
trent-codecov, line 127: This seems like it would be
```sql
select * from owners join GithubAppInstallation on owner.ownerid != GithubAppInstallation.ownerid
```
That would return all of the installations for other owners. 

I assume the goal is to find owners without the installation. If so you could use a left join and where null.
```sql
select * from owners left join GithubAppInstallation on owners.ownerid = GithubAppInstallation.owner_id where  GithubAppInstallation.owner_id is null
```

trent-codecov (changes requested):
One query question/fix.

tasks/backfill_gh_app_installations.py
```diff
@@ -0,0 +1,198 @@
...
+            log.info("Successful backfill", extra=dict(ownerid=ownerid))
+
+    async def run_async(
+        self,
+        db_session: Session,
+        owner_ids: Optional[List[int]] = None,
+        *args,
+        **kwargs
+    ):
+        print("starting test")
```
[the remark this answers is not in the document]
adrian-codecov: 💀

tasks/backfill_gh_app_installations.py
```diff
@@ -0,0 +1,198 @@
...
+        # Backfill gh apps we already have
+        await self.backfill_existing_gh_apps(db_session=db_session, owner_ids=owner_ids)
+
+        # Backfill owners with legacy integration + adding new gh app
+        await self.backfill_owners_with_integration_without_gh_app(
```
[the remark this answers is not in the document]
adrian-codecov: Sounds good, we can break this into it's own task if we need to if we find a proper case to run this in a cron or more periodically. I'll leave this as is for now and change it later when we get to this use case, unless you think we should change this here

tasks/backfill_gh_app_installations.py
adrian-codecov: I think what we want is this slightly modified query from yours, 

``` sql
select * from owners left join GithubAppInstallation on owners.ownerid = GithubAppInstallation.owner_id where  GithubAppInstallation.owner_id is null and owner.integration_id is not null
```

Difference being to filter by owner.integration_id != null, which went from "owners without gh app" to "owners without gh app with an installation id", cause we don't want to backfill for customers without an integration id to start with

tasks/backfill_gh_app_installations.py
trent-codecov: Sounds good to me

tasks/backfill_gh_app_installations.py
trent-codecov: This sounds good.

trent-codecov (approved):

PR #265, opened 2024-02-12
fix: apply notify task after lock is released

joseph-sentry:
there is the possibility of an issue when the
notify task is queued up before the test result finisher releases the notification lock due to the fact that a notify will drop itself if it fails to lock, so we should make sure that the test result finisher is not holding the notification lock before it queues up the notify task

[Sentry Issue](https://codecov.sentry.io/issues/4947303451/?project=-1&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=1h&stream_index=2)

Fixes: https://github.com/codecov/internal-issues/issues/364

### DIFF
--- a/tasks/test_results_finisher.py
+++ b/tasks/test_results_finisher.py
@@ -34,6 +34,7 @@ ESCAPE_FAILURE_MESSAGE_DEFN = [
     Replacement(["\r"], "", EscapeEnum.REPLACE),
     Replacement(["\n"], "<br>", EscapeEnum.REPLACE),
 ]
+QUEUE_NOTIFY_KEY = "queue_notify"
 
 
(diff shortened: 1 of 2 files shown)


giovanni-guidini:
There's 1 more change I'd like to see in this PR, which is changing 

`99: except LockError as err:` 

to be 

`99: except LockRetry as err:`

You see, that is when the notification task realizes that there's possibly another notification task happening. However when line 87 was changed to use `lock_manager` this broke. Previously it used `redis` directly (which raised `LockError`), but the `lock_manager` raises the `LockRetry` error.

So we should handle the exception better.

giovanni-guidini (approved):
Approving on the basis that the other change I commented on has nothing to do with these changes so should probably be on another PR.
This seems fine. Apologies for suggesting unrelated changes 😅

PR #272, opened 2024-02-16
feat: use default codecov installation when user has no bot or installation

joseph-sentry:
Fixes: https://github.com/codecov/engineering-team/issues/1170

### DIFF
--- a/services/bots.py
+++ b/services/bots.py
@@ -117,10 +117,28 @@ def get_token_type_mapping(repo: Repository):
             "Repository has no good bot for admin, but still continuing operations in case it is not doing an admin call anyway",
             extra=dict(repoid=repo.repoid),
         )
+
+    if get_config("github", "integration", "id") and get_config(
+        "github",
+        "integration",
+        "default_installation_id",  # this is different from the integration id,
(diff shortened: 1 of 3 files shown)


services/bots.py
```diff
@@ -117,10 +117,28 @@ def get_token_type_mapping(repo: Repository):
             "Repository has no good bot for admin, but still continuing operations in case it is not doing an admin call anyway",
             extra=dict(repoid=repo.repoid),
         )
+
+    if get_config("github", "integration", "id") and get_config(
```
trent-codecov, line 121: I think you need a "if service is github" here otherwise it will try to do this for bb/gl

services/bots.py
```diff
@@ -117,10 +117,32 @@ def get_token_type_mapping(repo: Repository):
...
             extra=dict(repoid=repo.repoid),
         )
+
+    if (
+        repo.service == "github"
+        and get_config("github", "integration", "id")
+        and get_config(
+            "github",
+            "integration",
+            "default_installation_id",  # this is different from the integration id,
```
giovanni-guidini, line 127: is this something new that you are adding or a piece of config that already exists?

PR #273, opened 2024-02-17
fix: try junit parser if file extension is xml

joseph-sentry:


tasks/test_results_processor.py
```diff
@@ -224,7 +226,7 @@ def parse_single_file(
 
         return res
 
-    def match_report(self, file_bytes):
+    def match_report(self, filename: str, file_bytes):
```
adrian-codecov, line 229: Is it easy to type file_bytes?

tasks/test_results_processor.py
```diff
@@ -161,11 +161,12 @@ def process_individual_arg(self, upload: Upload, repository) -> List[Testrun]:
 
         testrun_list = []
 
-        for file in data["test_results_files"]:
-            file = file["data"]
+        for file_dict in data["test_results_files"]:
+            filename = file_dict["filename"]
+            file = file_dict["data"]
```
adrian-codecov, line 166: What's your POV on dataclasses vs dicts?

adrian-codecov (approved):

tasks/test_results_processor.py
joseph-sentry: we're loading this from json and by default `json.loads` returns a dict, but it might be worth looking for a solution where it's loaded into a class

adrian-codecov: :bufo-ack:

Yeah I like dataclasses cause it's nice for the IDE for autocomplete, plus we know what the response is, but yea for the future

PR #279, opened 2024-02-22
feat: Configure multi apps per task per owner

giovanni-guidini:
Allows for owners to define configuration of what app
to use for a specific task. Owners can have multiple apps
with the same name that are considered for the task,
as long as they are configured properly.

The catch here is that currently there's no way to configure
these apps unless you do it manually. We will use that for testing.

### Legal Boilerplate


giovanni-guidini:
Depends on https://github.com/codecov/codecov-api/pull/421

matt-codecov (approved):

PR #282, opened 2024-02-23
fix: turn off django autocommit, start committing when tasks finish

matt-codecov:
sqlalchemy has autocommit turned off, let's make django behave the same way

### Legal Boilerplate


tasks/tests/unit/test_base.py
```diff
@@ -223,6 +229,98 @@ def test_wrap_up_dbsession_invalid_nothing_works(self, mocker):
...
+        task = celery_app.tasks[registered_task.name]
+        task.apply()
+
+        assert mock_django_rollback.call_args_list == [call()]
+        assert mock_dbsession_rollback.call_args_list == [call()]
+
+        assert mock_django_commit.call_args_list == [call()]
+        assert mock_wrap_up.call_args_list == [call(dbsession)]
+
+    def test_run_sqlalchemy_error_rollback(self, mocker, dbsession, celery_app):
```
adrian-codecov, line 302: Should this test assert something about sqlalchemy? Or should it say django?

tasks/base.py
```diff
@@ -159,6 +161,46 @@ def apply_async(self, args=None, kwargs=None, **options):
...
+        try:
+            django_transaction.commit()
+        except Exception as e:
+            log.warning(
+                "Django transaction failed to commit.",
+                exc_info=True,
+                extra=dict(e=e),
+            )
+
+        if timeseries_enabled():
```
adrian-codecov, line 174: Wouldn't we be committing to timeseries a lot here for non-timeseries transactions if this is the condition? I read this as "if you have timeseries setup, and you come to this fn, you'll commit", but we commit to timeseries very infrequently so wouldn't we be overcommiting?

tasks/tests/unit/test_base.py
matt-codecov: `SQLAlchemyError` is the base class. `IntegrityError` is also an example of `SQLAlchemyError` but it has a separate `except` branch so it's a separate test

will add a comment or something explaining

tasks/base.py
matt-codecov: https://github.com/codecov/worker/pull/283 will make base task log a few metrics to timescale for each task run, so we will need to commit the timeseries db

if we do away with the timescale metrics and just use alloy we could maybe skip the timeseries commit, but then again, committing an empty transaction is probably not expensive

adrian-codecov (approved):

PR #283, opened 2024-02-23
re-enable logging basetask metrics to sql models

matt-codecov:
worker is officially able to use django models now. uncommenting this usage of django models so basetask metrics will be sent to sql https://github.com/codecov/worker/blob/bf4fe4432e362109aaa8577420f774b890db6905/tasks/base.py#L218-L229

before deploying to prod, deploy to staging and check that data flows

### Legal Boilerplate


giovanni-guidini (approved):
Very excited to get these metrics!

PR #284, opened 2024-02-23
refactor: refactor feature with new interface

daniel-codecov:
Refactors the instances of `Feature` to be updated with this [pr](https://github.com/codecov/shared/pull/130/files). The overrides and variants no longer live in code, but rather the database and can be modified via django admin. For these two features, they will be seeded by this [migration](https://github.com/codecov/shared/pull/130/files#diff-0b7e216fec1adeb9819362f7c079ba848f947cc81e9892ffa95cc0d427202013) so we don't need to do anything when this code gets deployed. (Their proportions were already 0, so technically they don't affect anything anyways)

We also need to use a new commit hash for shared once shared is merged


matt-codecov (commented):
you will also need to update worker's pip requirements to use the new version of `shared`

if you've pushed your latest shared changes to github in a branch, you can get the SHA of the top commit on your branch. once you've merged your shared changes into the main branch, you want to grab the merge commit's SHA

with the SHA in hand, edit `requirements.in` and change the commit sha in the line for shared to the new one. then run `pip-compile requirements.in` to regenerate `requirements.txt`. then `make build` to rebuild your docker container, restart your docker compose setup, and off to the races. same process for `codecov-api`

matt-codecov (approved):

PR #287, opened 2024-02-26
Fix test result PR comment

joseph-sentry:
This PR adds escaping for certain characters when making the test result PR comment.

We have to do this to properly display the failure message table.

### DIFF
--- a/helpers/string.py
+++ b/helpers/string.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 from enum import Enum
 from typing import List
 
@@ -8,11 +9,11 @@ class EscapeEnum(Enum):
     REPLACE = "replace"
 
 
+@dataclass
 class Replacement:
-    def __init__(self, chars: str, output: str, method: EscapeEnum):
-        self.chars = chars
-        self.output = output
-        self.method = method
+    chars: str
+    output: str
+    method: EscapeEnum
 
 
 class StringEscaper:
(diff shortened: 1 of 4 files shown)


helpers/string.py
```diff
@@ -0,0 +1,54 @@
...
+from typing import List
+
+
+class EscapeEnum(Enum):
+    APPEND = "append"
+    PREPEND = "prepend"
+    REPLACE = "replace"
+
+
+class Replacement:
```
giovanni-guidini, line 11: [nit]
I believe you can use a [@dataclass](https://docs.python.org/3/library/dataclasses.html) for this. That way you don't have to specify the `__init__` function, just the fields.

```python
@dataclass
class Replacement:
    chars: str
    output: str
    method: EscapeEnum
```

helpers/string.py
```diff
@@ -0,0 +1,54 @@
...
+    def __init__(self, chars: str, output: str, method: EscapeEnum):
+        self.chars = chars
+        self.output = output
+        self.method = method
+
+
+class StringEscaper:
+    """
+    Class to use to escape strings using format defined
```
giovanni-guidini, line 20: Amazing docs 👏
thanks for that

helpers/string.py
```diff
@@ -0,0 +1,54 @@
...
+        to each char in Replacement.chars using the char in output
+
+        for example:
+            escape_def = [
+                Replacement(chars="<>", output="\\", method=EscapeEnum.PREPEND),
+                Replacement(chars="pre", output="div", method=EscapeEnum.REPLACE)
+            ]
+            StringEscaper(escape_def)
+
+            replace("<pre></pre>")
```
giovanni-guidini, line 37: If you apply the `replace` function for every _char_ in this `Replacement` definition wouldn't the output actually be `\<divdivdiv\>\</divdivdiv\>` ?

helpers/tests/unit/test_string.py
```diff
@@ -0,0 +1,13 @@
+from helpers.string import EscapeEnum, Replacement, StringEscaper
+
+
+def test_string_escaper():
```
giovanni-guidini, line 4: I suggest you include the example in the documentation here

helpers/string.py
joseph-sentry: good point, that was a bad example, but this also brings up the point that maybe this should include an option for specifying strings to replace, append, or prepend to

giovanni-guidini (approved):

PR #290, opened 2024-02-27
fix: re-enable sql task metrics

matt-codecov:
https://github.com/codecov/worker/pull/289 should fix the problems i saw when i deployed these metrics to staging

enabled in https://github.com/codecov/worker/pull/283, reverted in https://github.com/codecov/worker/pull/286, back on now

sorry for the churn on this. in the future i'll deploy to staging manually to test for things like this before merging

### Legal Boilerplate


giovanni-guidini (approved):
✨ 3rd time is the charm ✨

PR #301, opened 2024-03-04
fix: remove feature flags

daniel-codecov:
Temporarily remove feature flags since it is querying django models which will be blocking the event loop in asyncio.  

### Legal Boilerplate

### DIFF
--- a/rollouts/__init__.py
+++ b/rollouts/__init__.py
@@ -10,7 +10,20 @@ LIST_REPOS_GENERATOR_BY_OWNER_ID = Feature(
 
 # Eventually we want all repos to use this
 # This flag will just help us with the rollout process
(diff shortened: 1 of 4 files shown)


joseph-sentry (approved):

rollouts/__init__.py
```diff
@@ -10,7 +10,20 @@
...
+#     0.0,
+# )
+
+
+class TempFeature:
+    def __init__(self):
+        pass
+
+    def check_value(self, id, default=False):
+        if id == 16621196 or id == 16273544:
```
daniel-codecov, line 24: gio's overrides

matt-codecov (approved):

rollouts/__init__.py
thomasrockhu-codecov: lolol, perhaps would be useful to a comment in the code as to what this is

PR #302, opened 2024-03-05
fix: add feature flags back

daniel-codecov:
It basically just reverts this [PR](https://github.com/codecov/worker/pull/301/files) since the new removed django async deploy is now stable. Also fixes the way we mock async iterators for sync repo tests

### Legal Boilerplate


matt-codecov (approved):
probably don't want https://github.com/codecov/shared/pull/146 and instead want https://github.com/codecov/worker/pull/305 (or https://github.com/codecov/worker/pull/304)

PR #304, opened 2024-03-06
remove most of the spurious async/awaits from worker tasks

matt-codecov:
tl;dr: remove most of the `async`/`await` from worker tasks. where async functions remain (e.g. functions in shared) run them with `async_to_sync`. https://github.com/codecov/worker/pull/305 is a faster unblock until we can follow through on this change if preferred

our use of async in worker is mostly spurious. it's basically:
```
async def run_async():
    await foo()
    await bar()
    await baz()

asyncio.run(run_async())
```
when `run_async()` runs `await foo()`, it can yield control to some other task. but there's no other task going on in the event loop, and `run_async()` can't continue to start `bar()` until after `foo()` totally completes. so there's nothing else to run and we wind up running `foo()`, `bar()`, and `baz()` synchronously

django is unhappy with us if we try to use its ORM from inside an async event loop. since our base task runs everything in an async function, it'll be unhappy quite a lot. this PR removes that top-level async and most of the async throughout tasks.

it's possible we _could_ benefit from `async`/`await`, but we would have to use it more like the following pattern:
```
def run_impl():
    yaml_future = get_commit_yaml()
    pr_info_future = get_pr_info()
    yaml, pr = async_to_sync(asyncio.gather)(yaml_future, pr_info_future)
```
we would have to move logic around so that all of our async functions are ready to be run at the same time and then use asyncio to run them concurrently.

### useful commands for updating/recreating this PR
run from root of worker repo
i recommend installing `watch` and `rg` with `brew install watch ripgrep`

```
# replace all `async def` with `def`
find tasks -type f -name '*.py' -exec sed -E -i '' 's/async def/def/' {} +

# delete all `@pytest.mark.asyncio` lines in tests
find tasks/tests -type f -name '*.py' -exec sed -E -i '' '/@pytest.mark.asyncio/d' {} +

# rename `run_async` to `run_impl`
find tasks -type f -name '*.py' -exec sed -E -i '' 's/run_async/run_impl/' {} +

# rename `process_async` to `process_impl`
find tasks -type f -name '*.py' -exec sed -E -i '' 's/run_async/run_impl/' {} +

# replace `await self.` with `self.`
find tasks -type f -name '*.py' -exec sed -E -i '' 's/await self\./self./' {} +

# display a list of all usages of "await", auto-updated every 2 seconds
watch -c -n 2 "rg -j1 'await' tasks --color always | less -R"

# print first usage of `async_to_sync` in every file
# useful to find all files which use it without importing it in one go
find tasks -type f -name '*.py' -exec awk 's != FILENAME && /async_to_sync/{print FILENAME; print; s=FILENAME}' {} +
```

process was basically:
- run the commands starting with `find` to get rid of most async usage in worker + tests
- run the `watch` command in a separate terminal to view a live-updating list of places that still have `await`s
- changed all the `await`s to use `async_to_sync`
- ran the tests, fixed whatever came up

### Legal Boilerplate

### DIFF
--- a/tasks/sync_pull.py
+++ b/tasks/sync_pull.py
@@ -1,3 +1,4 @@
+import asyncio
 import json
 import logging
 import os
@@ -167,14 +168,12 @@ class PullSyncTask(BaseCodecovTask, name=pulls_task_name):
         commits = None
         db_session.commit()
         try:
-            commits_future = repository_service.get_pull_request_commits(pull.pullid)
-            base_ancestors_tree_future = repository_service.get_ancestors_tree(
+            commits = async_to_sync(repository_service.get_pull_request_commits)(
+                pull.pullid
+            )
+            base_ancestors_tree = async_to_sync(repository_service.get_ancestors_tree)(
                 enriched_pull.provider_pull["base"]["branch"]
             )
-            commits, base_ancestors_tree = async_to_sync(
-                asyncio.gather(commits_future, base_ancestors_tree_future)
-            )
-
             commit_updates_done = self.update_pull_commits(
                 enriched_pull, commits, base_ancestors_tree
             )
--- a/tasks/sync_pull.py
+++ b/tasks/sync_pull.py
@@ -6,6 +6,7 @@ from datetime import datetime
 from typing import Any, Dict, List, Sequence
 
 import sqlalchemy.orm
+from asgiref.sync import async_to_sync
 from redis.exceptions import LockError
 from shared.celery_config import notify_task_name, pulls_task_name
 from shared.reports.types import Change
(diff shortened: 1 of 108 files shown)


tasks/sync_pull.py
```diff
@@ -166,8 +168,10 @@ async def run_async_within_lock(
         commits = None
         db_session.commit()
         try:
-            commits = await repository_service.get_pull_request_commits(pull.pullid)
-            base_ancestors_tree = await repository_service.get_ancestors_tree(
+            commits = async_to_sync(repository_service.get_pull_request_commits)(
+                pull.pullid
+            )
+            base_ancestors_tree = async_to_sync(repository_service.get_ancestors_tree)(
```
matt-codecov, line 174: this is an example of where asyncio could be used successfully:
```
commits_future = repository_service.get_pull_request_commits(pull.pullid)
ancestor_tree_future = repository_service.get_ancestors_tree(enriched_pull.provider_pull["base"]["branch"])
commits, base_ancestors_tree = async_to_sync(asyncio.gather)(commits_future, ancestor_tree_future)
```
i didn't though because doing that makes a test fail. as-written, the commits call raises an exception and the ancestors tree call is never run. they both run if done as above, but this test passes a `set()` instead of a `dict()` for `provider_pull` so the test fails when trying to subscript it. besides the failure, it's a slight difference in behavior that doesn't belong here

adrian-codecov (approved):

PR #308, opened 2024-03-07
Shorten paths

joseph-sentry:
Add path shortening to the failure messages in the test results comment

This works by doing a regex to find all file paths in the failure message using a regex, then shortening them and replacing their occurrences in the failure message with the shortened version


helpers/string.py
```diff
@@ -61,3 +63,47 @@ def replace(self, replacement_target):
...
+
+    """
+
+    matches = file_path_regex.findall(string)
+    for match_tuple in matches:
+        file_path = match_tuple[0]
+        split_file_path = file_path.split("/")
+
+        # if the file_path has more than 3 components we should shorten it
+        if len(split_file_path) > 3:
```
adrian-codecov, line 98: Thoughts on making this 3 a variable describing its purpose for readability? So something like MAX_COMPONENT_LENGTH or something like that?

adrian-codecov (approved):
Little comment, otherwise approved

PR #310, opened 2024-03-07
Fix test results PR comment

joseph-sentry:
Depends on #311 
Fixes: https://github.com/codecov/engineering-team/issues/1296

- Vertically stack file description

- Take into account existing breaks in insert_breaks function

### DIFF
--- a/helpers/string.py
+++ b/helpers/string.py
@@ -67,6 +67,11 @@ class StringEscaper:
 
 MAX_PATH_COMPONENTS = 3
 
+
+# matches file paths with an optional line number and column at the end:
+# /Users/josephsawaya/dev/test-result-action/demo/calculator/calculator.test.ts:10:31
+# /Users/josephsawaya/dev/test-result-action/demo/calculator/calculator.test.ts
+# Users/josephsawaya/dev/test-result-action/demo/calculator/calculator.test.ts
 file_path_regex = regex.compile(
     r"((\/*[\w\-]+\/)+([\w\.]+)(:\d+:\d+)*)",
 )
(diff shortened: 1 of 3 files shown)


helpers/string.py
```diff
@@ -61,3 +63,47 @@ def replace(self, replacement_target):
                         string, replacement.output
                     )
         return replacement_target
+
+
+file_path_regex = regex.compile(
+    r"((\/*[\w\-]+\/)+([\w\.]+)(:\d+:\d+)*)",
```
thomasrockhu-codecov, line 69: would be nice if there were a comment of examples of valid regexes

services/test_results.py
```diff
@@ -172,6 +172,20 @@ async def notify(
             )
             return False, "torngit_error"
 
+    def format_single_test(self, test_name, test_env_list):
+        testsuite_section = f"Testsuite:<br>{test_name.split('//////////')[0]}"
```
giovanni-guidini, line 176: is `'//////////'` something that we put there? Seems very random to have this many '/' in a row...

Also [nit] maybe create a variable for this value, especially if it's something we put there. It could come with a comment explaining why this exists...

giovanni-guidini (approved):

services/test_results.py
joseph-sentry: We did put it there, it's because I was trying to deduplicate failure messages, I wanted to have multiple tests that map to the same failure message in the same row, so I had to use a dict, and for that I had to put the testsuite and testname together as the key in the dict, but I needed a way to separate them back out so I joined using `/////////` when placing them as the key in the dict and then split it back out here.

I'm planning on removing this in my next change to the notifier.

PR #315, opened 2024-03-12
fix: check if PreProcessUpload is running

giovanni-guidini:
These changes add a verification for `PreProcessUpload` task already running in another worker pod. This is to avoid a report being created twice for the same commit.

Imagine the following scenario:
* You make 2 concurrent uploads (calling `create-report` twice for the same commit).
* Task A will get the lock and create a new report;
* Task A releases the lock and task B immediately gets it;
* Task A is commiting the transaction (outside the lock), at the same time task B passed [this point](https://github.com/codecov/worker/blob/e41ba11d68e37ffc7892e93954f68a7b0175cf93/services/report/__init__.py#L278) and is creating a new report;
* Task B creates a new report and saves it again;

I'm not sure if the report will have more sessions than it should, but the database will show more `reports_upload` instances than it should

To avoid that we simply check if there's a task already running. Because this task really only needs to run once, if one of them runs that's enough.

You may think this is not possible, but I did stumble upon this scenario when investigating https://l.codecov.dev/SPUjD3 (see screenshot, check link for full investigation) well.... at least I think it was this scenario... maybe there's a `dbsession.flush()` missing somewhere

image

### DIFF
--- a/tasks/preprocess_upload.py
+++ b/tasks/preprocess_upload.py
@@ -45,8 +45,10 @@ class PreProcessUpload(BaseCodecovTask, name="app.tasks.upload.PreProcessUpload"
             "Received preprocess upload task",
             extra=dict(repoid=repoid, commit=commitid, report_code=report_code),
         )
-        lock_name = f"preprocess_upload_lock_{repoid}_{commitid}"
+        lock_name = f"preprocess_upload_lock_{repoid}_{commitid}_{report_code}"
         redis_connection = get_redis_connection()
+        # This task only needs to run once per commit (per report_code)
+        # To generate the report. So if one is already running we don't need another
         if self._is_running(redis_connection, lock_name):
             log.info(
                 "PreProcess task is already running",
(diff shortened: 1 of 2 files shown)


joseph-sentry (approved):
I'm approving because I agree that this is a scenario that could occur and that this would fix it. 

Do you think an alternative solution would be to add a `dbsession.commit` to the code that runs within the lock?

tasks/preprocess_upload.py
```diff
@@ -70,7 +76,12 @@ def run_impl(
                     lock_name=lock_name,
                 ),
             )
-            return {"preprocessed_upload": False}
+            return {"preprocessed_upload": False, "reason": "unable_to_acquire_lock"}
```
joseph-sentry, line 79: I would also maybe add the check to see if the lock is taken here in order to decided whether to log.warning("unable to acquire lock") or log.info("PreProcess task is already running"), because it's possible that the `LockError` is caused by another task having already taken it

joseph-sentry (commented):
I think it would also make sense to change the blocking_timeout on the lock to None, since if we fail to get the lock we don't actually want to keep trying, we can fail immediately

joseph-sentry (approved):

PR #316, opened 2024-03-12
feat: add FailureNormalizer service

joseph-sentry:
The first step towards flaky test detection
requires us to normalize failure messages to
ignore random differences in failure messages
such as uuids generated every test run that
would differ between failure messages in different test runs.


services/failure_normalizer.py
```diff
@@ -0,0 +1,77 @@
+from typing import Dict
+
+import regex
+
+predefined_dict_of_regexes_to_match = {
+    "UUID": r"[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}",
+    "DATETIME": r"(?:19|20)[0-9]{2}-(?:(?:0?[0-9])|(?:1[012]))-(?:(?:0?[0-9])|1[0-9]|2[0-9]|3[01])T(?:0[0-9]|1[0-9]|2[0-3]):(?:[0-5][0-9]):(?:[0-5][0-9])(?:Z|(?:-(?:0[0-9]:[03]0)))?",
+    "DATETIME2": r"(?:19|20)[0-9]{2}(?:(?:0?[0-9])|(?:1[012]))(?:(?:0?[0-9])|1[0-9]|2[0-9]|3[01])T(?:0[0-9]|1[0-9]|2[0-3])(?:[0-5][0-9])(?:[0-5][0-9])(?:Z|(?:-(?:0[0-9]:[03]0)))?",
```
giovanni-guidini, line 8: What are we to gain from breaking up the message into `DATETIME` and `DATETIME2` ?

(as opposed to combining both messages into "DATETIME")

services/failure_normalizer.py
```diff
@@ -0,0 +1,77 @@
...
+class FailureNormalizer:
+    """
+    Class for normalizing a failure message
+
+    Takes a dict of strings to strings where the key is the replacement string and the value is
+    the regex we want to match and replace occurences of and a boolean that toggles whether we
+    should ignore the predefined list of patterns in the constructor
+
+    The normalize_failure_message method on the class takes a string as a parameter
+    it will remove all occurences of a match of the regexes specified in the list
```
giovanni-guidini, line 27: [nit] 'it will remove all occurences' it's not really accurate, is it? It _replaces_ them, rather than _removing_

services/failure_normalizer.py
```diff
@@ -0,0 +1,77 @@
...
+        if not ignore_predefined:
+            dict_of_regex_strings = (
+                predefined_dict_of_regexes_to_match | dict_of_regex_strings
+            )
+
+        for key, regex_string in dict_of_regex_strings.items():
+            compiled_regex = regex.compile(regex_string, flags=flags)
+            self.dict_of_regex[key] = compiled_regex
+
+    def normalize_failure_message(self, failure_message: str):
```
giovanni-guidini, line 72: I'd be interested in a timer metric for this function

giovanni-guidini (commented):
Looking promising!
I'd like to just understand the double keys in the dict before approving, please.

services/failure_normalizer.py
joseph-sentry: Technically I don't think a datetime format can selectively use dashes between the numbers so I broke down the datetime regexes into one regex for datetimes with dashes and one regex for datetimes without dashes, I did the same thing for time as well, there's a way to combine both of them but I think it's actually more simple to have it be this way, or else the regex may become too long or too complicated.

what might be a good idea is to allow the values in the dict to be lists of regexes instead of just one regex, so we can avoid having to make multiple keys, but then allowing users to customize this existing dict becomes complicated (do we append to the list, do we replace the list entirely? Right now the behaviour is that if a user hasa key in their dict that matches a key in the predefined dict, it will replace the value in the predefined dict with theirs).

giovanni-guidini: I like turning the values into lists and having less keys.

I feel that combining user provided regexes into the list would be the best idea, and we can put those first, so if it's gonna match it'll match the user defined regex first.

services/tests/test_failure_normalizer.py
```diff
@@ -0,0 +1,126 @@
...
+    f = FailureNormalizer(user_dict)
+    s = f.normalize_failure_message(test_string)
+
+    assert (
+        s
+        == "UUID UUID_string DATE UUID HEXNUMBER DATETIME DATETIME  TIME  DATETIME URL LINENO LINENO :: HEXNUMBER"
+    )
+
+
+def test_failure_normalizer_overwrite_predefined():
```
giovanni-guidini, line 41: This test is not being executed because it has the same name as the test below.
Either merge the tests or change the name of one of them

services/tests/test_failure_normalizer.py
```diff
@@ -0,0 +1,126 @@
...
+message="",
+commitid="abf6d4df662c47e32460020ab14abf9303581429",
+s = b'\\x592f6b514678496f4333336a54314f71774c744f7934524d4479517778715270446678487459344769777458454a584d632b61633349432f35636c52635659473330782f7a496b7a5053542b426333454d614c5635673d3d'
+altchars = None, validate = False
+""",
+            """mocker = &lt;pytest_mock.plugin.MockFixture object at HEXNUMBER&gt;
+mock_configuration = &lt;shared.config.ConfigHelper object at HEXNUMBER&gt;
+chain = mocker.patch("tasks.upload.chain")
+storage_path = (
+    "vNO/repos/testing/UUID/bundle_report.sqlite"
```
giovanni-guidini, line 98: The FILEPATH regex didn't pickup these paths because the custom `order_to_process` in the test doesn't include it

giovanni-guidini (approved):
Please update the tests before merging

PR #319, opened 2024-03-13
perf: experiment with bulk instert test results

giovanni-guidini:
These changes have 2 obejctives:
1. (duh) try to speed up time of writing tests and test instances in the DB.
    This is inspired by https://docs.sqlalchemy.org/en/13/faq/performance.html#i-m-inserting-400-000-rows-with-the-orm-and-it-s-really-slow
    The idea is to bulk_update and let the DB handle the concurrency and the conflicts.

2. Test using Sentry metrics + Features for perf potential improvements.
    Although I suspect the time benefit will be enough for us to prefer the bulk_insert technique I don't know what the benefit is (potentially). And there's the drawback of using extra memory, which I'm also not sure how much extra memory it is.
    So I want to use this opportunity to explore this setup of running perf experiments :D

### Legal Boilerplate

### DIFF
--- a/tasks/test_results_processor.py
+++ b/tasks/test_results_processor.py
@@ -3,6 +3,7 @@ import json
 import logging
 import zlib
 from io import BytesIO
+from sys import getsizeof
 from typing import List
 
 from sentry_sdk import metrics
@@ -10,6 +11,7 @@ from shared.celery_config import test_results_processor_task_name
 from shared.config import get_config
 from shared.yaml import UserYaml
 from sqlalchemy.dialects.postgresql import insert
+from sqlalchemy.orm import Session
 from test_results_parser import (
     ParserError,
     Testrun,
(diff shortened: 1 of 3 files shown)


tasks/test_results_processor.py
```diff
@@ -148,6 +138,116 @@ def process_individual_upload(
...
+            insert_test_instances = insert(TestInstance.__table__).values(
+                test_instance_data
+            )
+            db_session.execute(insert_test_instances)
+            db_session.flush()
+        # Memory outside the time metrics to not disturb the counter
+        # Obviously this is a very rough estimate of sizes. We are interested more
+        # in the difference between the insert approaches. SO this should be fine.
+        # And these aux memory structures take the bulk of extra memory we need
+        memory_used += getsizeof(test_data) // 1024
```
joseph-sentry, line 212: I tested this out locally and the memory_used i was getting was 0, I'm not sure if this is just a local thing I ran into or if there's a problem with this approach, I'm okay with trying it out to see if it works out in prod, since that's just my experience running it locally

joseph-sentry (approved):
changes look good, I had trouble verifying the memory usage gathering locally but I think we can see if that problem persists in prod

tasks/test_results_processor.py
giovanni-guidini: Depending on the number of tests that you tested against it would be small enough in size that it would return 0. After all it's Kb and the integer division will round down.

Before adding the `// 1024` I was getting value for those calls, so... maybe that 🤷

PR #331, opened 2024-03-15
feat: add message when test_result is setup

giovanni-guidini:
To help users confirm that test_results is setup correctly we want to display
a message near the header saying that that is the case and all tests passed.

These changes accomplish that by creating a new `NotificationContext` in the
`ComparisonProxy` and carrying the info to the notifier.

### Legal Boilerplate

### DIFF
--- a/services/notification/notifiers/mixins/message/sections.py
+++ b/services/notification/notifiers/mixins/message/sections.py
@@ -103,6 +103,13 @@ class NewFooterSectionWriter(BaseSectionWriter):
 
 
 class NewHeaderSectionWriter(BaseSectionWriter):
+    def _possibly_include_test_result_setup_confirmation(self, comparison):
+        if comparison.all_tests_passed():
+            yield ("")
+            yield (
+                ":heavy_check_mark: Test ingestion set up successfully. No failed tests found. :relaxed:"
+            )
+
     async def do_write_section(self, comparison, diff, changes, links, behind_by=None):
         yaml = self.current_yaml
         base_report = comparison.project_coverage_base.report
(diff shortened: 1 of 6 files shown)


services/notification/notifiers/mixins/message/sections.py
```diff
@@ -303,6 +317,12 @@ async def do_write_section(self, comparison, diff, changes, links, behind_by=Non
                     "Changes have been made to critical files, which contain lines commonly executed in production. [Learn more](https://docs.codecov.com/docs/impact-analysis)"
                 )
 
+        if comparison.all_tests_passed():
+            yield ("")
+            yield (
+                ":heavy_check_mark: Test ingestion set up successfully. No failed tests found. :relaxed:"
+            )
```
rohan-at-sentry, line 324: Looks like this message will fire every time the `all_test_passed` check is true.  @giovanni-guidini  can you confirm?

giovanni-guidini: That is accurate.
Keep in mind that if the setup for test_results is complete but not all tests passed then we don't notify with this notifier.

rohan-at-sentry: in that case... after first time setup ever time all test past, we'll show a message titled "Test ingestion set up successfully..."

Which is a little weird.. 

Maybe we alter that message to say "All tests successful" instead. wdyt @giovanni-guidini @Adal3n3

rohan-at-sentry: @Adal3n3 @giovanni-guidini and @rohan-at-sentry aligned

Tweak the message to 

✅ All tests successful. No failed tests found :relaxed:

joseph-sentry (approved):

PR #336, opened 2024-03-18
fix: don't eager merge sessions into report

giovanni-guidini:
context: https://l.codecov.dev/xvGZaM

Because we eagerly merge sessions when processing a report the following scenario can happen:
1. You upload an empty report with a flag set to carryforward
2. There will be a new session for that upload (even though it's empty)
3. In the next commit the session will be carried forward

If no real upload was made to clean up old carriedforward sessions from empty uploads
they would accumulate in the report (because `_adjust_sessions` only runs *after* the
processing is done, and *doesn't run* if it's an empty upload). What prompted this
investigation is that a customer was fund with thousands of carryforward session in
some of their commits.

So I'm exploiting the fact that merges into the report need to be made from within a lock
and getting the ID for the session _without_ actually adding it to the original report.

If the report is successful we add it and run `_adjust_sessions`. If the report was empty,
we just never add it to the report, as if it didn't existed.

OBS.: Even though I use the litle exploit I believe that passing the `sessionid` to
`ReportBuilder` is unecessary. We can likely not get a sessionid at all when processin the report
and it would still work. I decided to leave this for it's own change tho.

Also I decided to bump the severity of the messages that indicate a report is empty.
Users are not expected to upload empty reports, so it's at least sensible it should be
a warning.

### Legal Boilerplate

### DIFF
--- a/services/report/tests/unit/test_process.py
+++ b/services/report/tests/unit/test_process.py
@@ -197,8 +197,8 @@ class TestProcessRawUpload(BaseTestCase):
                 session=Session(flags=["fruits"]),
                 flags=[],
             )
-        assert len(original_report.sessions) == 2
-        assert sorted(original_report.flags.keys()) == ["fruits", "unit"]
+        assert len(original_report.sessions) == 1
+        assert sorted(original_report.flags.keys()) == ["unit"]
         assert original_report.files == ["file_1.go", "file_2.py"]
         assert original_report.flags["unit"].totals == ReportTotals(
             files=2,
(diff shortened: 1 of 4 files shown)


giovanni-guidini:
PS about tests changed in this PR:

I started by creating a new test to verify the behavior observed on the investigation could be reproduced.
In fixing it the other test - that showed what happens when you upload an empty report - broke. It was left there so we wouldn't accidentally break this behavior (thanks Thiago for thinking forward like that 🙏 )

As I am not breaking the behavior accidentally, but very much intentionally, I altered the test to match the new behavior. And left my test too because it was there at this point 😅

services/report/tests/unit/test_process.py
```diff
@@ -1229,3 +1204,52 @@ def test_process_raw_upload_with_carryforwarded_flags(self):
...
+    def test_process_raw_upload_carryforward_flag_empty_upload(self):
+        """Tests that an upload with carryforward flag doesn't create a new session
+        in the original report for the flag it was uploaded with.
+
+        That is, a new session is created in the commit's report only if the upload
+        was properly processed
+        """
+        original_report = self._generate_sample_report()
+        original_sessions = original_report.sessions
+        expected_sessions = {
```
joseph-sentry, line 1217: In the previous behaviour, would we have seen 3 sessions instead of 2 here, with two of them being for the flag "somethingold"?

giovanni-guidini: Yes. You can see a similar behavior in the test I altered. The removed parts are the outcome of the old behavior, where an extra session would be present.

joseph-sentry: ah I see now, thanks for pointing that out

joseph-sentry (approved):
lgtm

PR #337, opened 2024-03-18
Update timeseries related tasks

JerrySentry:
- Modify backfill batch sizing for component measurements
- Add option to delete only Measurements (not Dataset) when using TimeseriesDeleteTask 

### Legal Boilerplate


services/timeseries.py
```diff
@@ -227,17 +228,23 @@ def repository_flag_ids(repository: Repository) -> Mapping[str, int]:
     return {repo_flag.flag_name: repo_flag.id for repo_flag in repo_flags}
```
JerrySentry, line 230: Before this change the batch size is always determined by number of flags even if the repo has no flags. Now it will batch by number of flags if trying to backfill flag coverage, batch by number of components if trying to backfill component coverage, and don't further batch (ie default to `backfill_max_batch_size`) if trying to backfill by generic coverage.

services/timeseries.py
```diff
@@ -248,3 +255,15 @@ def delete_repository_data(repository: Repository):
         owner_id=repository.ownerid,
         repo_id=repository.repoid,
     ).delete()
+
+
```
JerrySentry, line 259: Add this new function that is similar to `delete_repository_data` except it only deletes specific `Measurements` (by its name and measurable_id), this will be used when we want to only delete a specific trendline, for example, delete all datapoints of component_id="python3".

adrian-codecov (approved):

PR #338, opened 2024-03-18
feat: add FlakeDetector class

joseph-sentry:
This class is given a dbsession, repoid and an optional failure normalizer Returns a list of tuples of flaky test id and the type of flake for that id

### DIFF
--- a/services/flake_detection.py
+++ b/services/flake_detection.py
@@ -162,7 +162,7 @@ class FlakeDetector:
         return False
 
     @trace
-    def detect_flakes(self) -> List[Tuple[str, str]]:
+    def detect_flakes(self) -> dict[str, FlakeType]:
         """
         Detect flaky tests on a given repo based on the test instances
         gathered in the query in the constructor
--- a/services/flake_detection.py
+++ b/services/flake_detection.py
@@ -93,81 +93,71 @@ class FlakeDetector:
             .order_by(TestInstance.test_id)
             .all()
         )
-        memory_used = getsizeof(self.test_instances_ordered_by_test)
+        memory_used_kb = getsizeof(self.test_instances_ordered_by_test) // 1024
         metrics.gauge(
             key="flake_detection.populate.aux_memory_used",
-            value=memory_used,
+            value=memory_used_kb,
             unit="kilobytes",
         )
 
-    def check_if_failed_on_default(self, instance):
+    def check_if_failed_on_default(
+        self, curr_test_context: FlakeDetectionContext, instance: TestInstance
+    ):
         if instance.branch == self.default_branch:
-            self.curr_test_context.is_curr_flake = True
+            curr_test_context.is_curr_flake = True
             self.resulting_flakes[
-                self.curr_test_context.curr_test_id
+                curr_test_context.curr_test_id
             ] = FlakeType.FailedInDefaultBranch
 
-            metrics.incr(
-                "flake_detection.detect_flakes.flake_detected",
-                1,
-                tags={"flake_type": str(FlakeType.FailedInDefaultBranch)},
-            )
             return True
         return False
 
-    def check_if_consecutive_diff_outcomes(self, instance):
+    def check_if_consecutive_diff_outcomes(
+        self, curr_test_context: FlakeDetectionContext, instance: TestInstance
+    ):
         # ignore skips
         if instance.outcome != str(Outcome.Skip):
-            existing_outcome_on_commit = self.curr_test_context.commit_to_outcome.get(
+            existing_outcome_on_commit = curr_test_context.commit_to_outcome.get(
                 instance.commitid, None
             )
             if (
                 existing_outcome_on_commit is not None
                 and existing_outcome_on_commit != instance.outcome
             ):
-                self.curr_test_context.is_curr_flake = True
+                curr_test_context.is_curr_flake = True
                 self.resulting_flakes[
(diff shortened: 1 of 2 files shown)


services/tests/test_flake_detector.py
```diff
@@ -0,0 +1,203 @@
...
+    create_test_instance(
+        dbsession, test_id, uploadid, str(Outcome.Failure), "failure message"
+    )
+    create_test_instance(dbsession, test_id, uploadid2, str(Outcome.Pass), None)
+
+    fd = FlakeDetector(dbsession, repoid)
+
+    flaky_tests = fd.detect_flakes()
+
+    assert flaky_tests == [(test_id, FlakeType.ConsecutiveDiffOutcomes)]
```
giovanni-guidini, line 135: Apparently there are 4 outcomes (failure, skip, error, pass)

Suppose you are building a library that's platform dependent so you have certain tests that are only executed in certain platforms. So they will be `skip` in some and `pass` in others. Does this count as a flaky test?

services/tests/test_flake_detector.py
```diff
@@ -0,0 +1,203 @@
...
+    )
+
+    fd = FlakeDetector(dbsession, repoid)
+
+    flaky_tests = fd.detect_flakes()
+
+    assert flaky_tests == [(test_id, FlakeType.UnrelatedMatchingFailures)]
+
+
+def test_flake_matching_failures_on_related_branches(dbsession):
```
giovanni-guidini, line 172: This is a good test 👍

giovanni-guidini (commented):
I like the general shape of this. Left comments.

services/flake_detection.py
```diff
@@ -0,0 +1,148 @@
...
+                Commit.commitid,
+            )
+            .join(Test)
+            .join(Upload, TestInstance.upload_id == Upload.id_)
+            .join(CommitReport, Upload.report_id == CommitReport.id_)
+            .join(Commit, CommitReport.commit_id == Commit.id_)
+            .join(Repository, Repository.repoid == Test.repoid)
+            .filter(
+                Repository.repoid == repoid,
+                Upload.created_at >= (datetime.now() - timedelta(days=30)),
```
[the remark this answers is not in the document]
joseph-sentry: I think the join via upload is necessary regardless, to be able to get the commit information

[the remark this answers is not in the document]
joseph-sentry: for the time window issue I think this would require getting fancy with the query if we want to put a limit of results returned for each test id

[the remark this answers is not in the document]
giovanni-guidini: Ok. Best to not touch it now then. We optimize when we have to since it's gonna be some effort

services/flake_detection.py
```diff
@@ -0,0 +1,235 @@
...
+
+        TODO: Differentiate between random flakes and infra flakes. Infra flakes
+        are flakes that are caused by a dependency or a temporary outage.
+        They can be defined as: matching failure messages on unrelated branches
+        clustered at a point in time, this test should only be failing for that
+        range in time.
+
+        Returns a list of tuples of flaky test id and type of flake
+        """
+        self.curr_test_context = FlakeDetectionContext()
```
giovanni-guidini, line 198: [very nit] This is a personal preference, you can ignore it completely if you don't feel the same way.

Personally I'd have made `curr_test_context` just a variable of this function and passed it around. It's easier to write tests to individual functions that use the state (easier to mock the context), although only slightly.

But when reading the code I don't have to look into where this context is defined in the class (and it's not in the `__init__` method, so it's hard to find) if it's just another arg, with it's type right there.

One last reason is because of the garbage collection. Having this a local var will allow the context to the cleaned faster when the function finishes (if the class lives longer).

You don't have to change anything, just something to think about.

services/flake_detection.py
```diff
@@ -0,0 +1,235 @@
...
+                        self.curr_test_context.reset(instance.test_id)
+
+                    # if we've already determined the current test to be a flake
+                    # we don't have to keep examining instances of this test
+                    if self.curr_test_context.is_curr_flake:
+                        continue
+
+                    # check if failed on default branch
+                    # should probably automatically create an issue here
+                    if self.check_if_failed_on_default(instance):
```
giovanni-guidini, line 222: [nit] 

Move the metrics from inside the aux functions to inside each of the `if`s.
You'd still know what tag to emit, and the `for` won't look so... empty, the poor thing.
Also each function would be more focused on the conditions that mark a test as flake.

You can also change the `if...if...if` to `if...elif...elif` and keep the same effect of stopping earlier (if you have the metrics in the for) without all these `continue`s.

services/tests/test_flake_detector.py
```diff
@@ -0,0 +1,207 @@
...
+            repoid=repoid,
+            testsuite=testsuite,
+            name=name,
+            flags_hash="",
+        )
+        dbsession.add(t)
+        dbsession.flush()
+        return id_
+    else:
+        name = str(uuid4())
```
giovanni-guidini, line 67: It's weird to have uncovered lines in your test files. Please either add tests that use these lines or refactor this portion to remove the unused lines

giovanni-guidini (approved):
I think the general shape of this looks good. Any other comment would be minor.

But please address the comment on the uncovered lines on the test file before merging 🙏  thanks

PR #339, opened 2024-03-19
refactor(test results notify): to prepare for adding flaky test info

joseph-sentry:


tasks/test_results_finisher.py
```diff
@@ -129,15 +131,15 @@ def process_impl_within_lock(
 
         escaper = StringEscaper(ESCAPE_FAILURE_MESSAGE_DEFN)
 
+        failures = []
```
giovanni-guidini, line 134: maybe also remove the definition of `failures` in line 130? It seems to be replaced by this one.

giovanni-guidini (approved):
Very good job. Looks way cleaner and I quite like the aux dataclasses you have to make it easier to pass information around.

PR #345, opened 2024-03-25
chore: Add some additional logs to sync repos

ajay-sentry:
Looking to add some additional logs for easier parsing and triaging of future sync repo issues

### Legal Boilerplate


Gguidini (commented):
Please keep the log messages static. It makes it a lot easier to search for them.
To log extra info use the `extra` dict

codecov.yml
```diff
@@ -21,13 +21,13 @@ component_management:
     - component_id: actual_code
```
adrian-codecov, line 21: Why was this file changed?

services/bots.py
```diff
@@ -85,6 +98,15 @@ def get_repo_appropriate_bot_token(
     repo: Repository,
     installation_name_to_use: Optional[str] = GITHUB_APP_INSTALLATION_DEFAULT_NAME,
 ) -> Tuple[Dict, Optional[Owner]]:
+
+    log.info(
+        "Get repo appropriate bot token",
```
adrian-codecov, line 103: Are both this token + gh app installation safe things to log?

adrian-codecov (commented):
Some quick questions, otherwise good on my end

codecov.yml
ajay-sentry: These were some black (or prettier?) updates, I can try to unstage / revert these changes but they shouldn't affect anything

services/bots.py
ajay-sentry: I believe so, I think the token isn't generated til here:
https://github.com/codecov/worker/blob/70645fde638d559494b93f4ee9693847d52446ad/services/bots.py#L165-L170

where installation_name_to_use I think is a GH specific string:
https://github.com/codecov/worker/blob/305c1eb744e3cf0cf3165528a8a5ea83fdb5603a/services/notification/notifiers/base.py#L53

adrian-codecov (approved):

PR #346, opened 2024-03-25
Add Flake models and refactor flake detection

joseph-sentry:
Refactors flake detection to use FlakeDetectionEngine and symptom detectors

### DIFF
--- a/services/flake_detection.py
+++ b/services/flake_detection.py
@@ -17,7 +17,7 @@ from database.models.reports import CommitReport, Test, TestInstance, Upload
 
 log = getLogger(__name__)
 
-FlakeDetectionResult = dict[Test, dict[TestInstance, list[FlakeSymptomType]]]
+FlakeDetectionResult = dict[Test, set[FlakeSymptomType]]
 
 
 class SymptomDetector:
@@ -242,7 +242,7 @@ class FlakeDetectionEngine:
                     for symptom_detector in self.symptom_detectors:
                         symptom_detector.ingest(instance)
 
-            results = defaultdict(lambda: defaultdict(list))
+            results = defaultdict(set)
 
             with metrics.timing(
                 "flake_detection.detect_flakes.detection",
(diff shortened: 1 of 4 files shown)


services/tests/test_flake_detector.py
```diff
@@ -77,6 +83,7 @@ def create_test_instance(dbsession, test_id, upload_id, outcome, failure_message
     )
     dbsession.add(ti)
     dbsession.flush()
+    return ti
```
thomasrockhu-codecov, line 86: https://i.makeagif.com/media/5-22-2015/Ld82mY.gif

services/flake_detection.py
```diff
@@ -1,67 +1,170 @@
...
-    commit_to_outcome = dict()
+    def detect(self) -> dict[str, TestInstance]:
+        raise NotImplementedError()
 
-    def reset(self, test_id):
-        self.curr_test_id = test_id
-        self.is_curr_flake = False
-        self.flake_dict.clear()
-        self.commit_to_outcome.clear()
+    @cached_property
```
giovanni-guidini, line 42: Let's remember that having a class-wide cached property is not a good idea cause the cache is shared among all instances of the class... I think that was the gotcha.

I do think that each of these classes should return a single `FlakeSymtomType` so this seems fine. Probably it's what you intended, but just stating it again

services/flake_detection.py
```diff
@@ -1,67 +1,170 @@
...
 
+class BaseSymptomDetector:
+    """
+    blueprint class for a single symptom of flakiness detector
 
-class FlakeType(Enum):
-    FAILED_IN_DEFAULT_BRANCH = "failed_in_default_branch"
-    CONSECUTIVE_DIFF_OUTCOMES = "consecutive_diff_outcomes"
-    UNRELATED_MATCHING_FAILURES = "unrelated_matching_failures"
+    classes that implement this interface are expected to:
```
giovanni-guidini, line 24: [very nit] I think this can be broken up into docstrings in the abstract methods directly... I feel that this could very easily become obsolete and people forget to update it.

services/flake_detection.py
giovanni-guidini: That being said, I don't think we gain anything by having this cached.... in all your implementations it's just a constant being returned. Maybe I'd remove this cached_property, but nit in any case.

giovanni-guidini (approved):

PR #347, opened 2024-03-27
Daniel/parallel upload processing experiment

daniel-codecov:
### Legal Boilerplate

### DIFF
--- a/services/report/__init__.py
+++ b/services/report/__init__.py
@@ -41,6 +41,7 @@ from helpers.exceptions import (
     RepositoryWithoutValidBotError,
 )
 from helpers.labels import get_labels_per_session
+from rollouts import PARALLEL_UPLOAD_PROCESSING_BY_REPO
 from services.archive import ArchiveService
 from services.report.parser import get_proper_parser
 from services.report.parser.types import ParsedRawReport
@@ -519,7 +520,7 @@ class ReportService(BaseReportService):
             get_current_env() == Environment.local
             or commit.repoid in report_builder_repo_ids
         )
-        if not new_report_builder_enabled:
+        if not new_report_builder_enabled:  # sometimes this runs
             return self.get_existing_report_for_commit_from_legacy_data(
                 commit, report_class=report_class, report_code=report_code
             )
@@ -791,7 +792,7 @@ class ReportService(BaseReportService):
 
     @sentry_sdk.trace
     def build_report_from_raw_content(
-        self, master: Optional[Report], upload: Upload
+        self, master: Optional[Report], upload: Upload, parallel_idx=None
     ) -> ProcessingResult:
         """
             Processes an upload on top of an existing report `master` and returns
@@ -854,6 +855,7 @@ class ReportService(BaseReportService):
                     flags,
                     session,
                     upload=upload,
+                    parallel_idx=parallel_idx,
                 )
             log.info(
                 "Successfully processed report",
@@ -914,9 +916,15 @@ class ReportService(BaseReportService):
         db_session = upload_obj.get_db_session()
         session = processing_result.session
         if processing_result.error is None:
-            upload_obj.state = "processed"
-            upload_obj.state_id = UploadState.PROCESSED.db_id
-            upload_obj.order_number = session.id
+            if PARALLEL_UPLOAD_PROCESSING_BY_REPO.check_value(
+                "jfdios"
+            ):  # TODO: temp string
+                upload_obj.state_id = UploadState.PARALLEL_PROCESSED.db_id
+                upload_obj.state = "parallel_processed"
+            else:
+                upload_obj.state_id = UploadState.PROCESSED.db_id
+                upload_obj.state = "processed"
+            upload_obj.order_number = session.id  # this value uses parallel_idx
             upload_totals = upload_obj.totals
             if upload_totals is None:
                 upload_totals = UploadLevelTotals(
--- a/services/report/__init__.py
+++ b/services/report/__init__.py
@@ -43,6 +43,7 @@ from helpers.exceptions import (
 from helpers.labels import get_labels_per_session
 from rollouts import PARALLEL_UPLOAD_PROCESSING_BY_REPO
 from services.archive import ArchiveService
+from services.redis import get_redis_connection
 from services.report.parser import get_proper_parser
 from services.report.parser.types import ParsedRawReport
 from services.report.raw_upload_processor import process_raw_upload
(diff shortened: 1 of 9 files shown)


services/report/__init__.py
```diff
@@ -519,7 +536,7 @@ def get_existing_report_for_commit(
             get_current_env() == Environment.local
             or commit.repoid in report_builder_repo_ids
         )
-        if not new_report_builder_enabled:
+        if not new_report_builder_enabled:  # sometimes this runs
```
daniel-codecov, line 539: sometimes this is used

daniel-codecov: which affects how the parallel report is built

services/report/__init__.py
```diff
@@ -1056,3 +1081,26 @@ def save_full_report(self, commit: Commit, report: Report, report_code=None):
...
+
+        # save incremental results to archive storage
+        # upload_finisher will combine
+        chunks = report.to_archive().encode()
+        _, files_and_sessions = report.to_database()
+
+        archive_service.write_chunks(
+            commitid,
+            chunks,
+            report_code=f"parallel/chunks",  # TODO: actual report_code won't work anymore
```
daniel-codecov, line 1099: use/create a function that just allows me to write a file with custom name

services/report/raw_upload_processor.py
```diff
@@ -124,7 +127,7 @@ def process_raw_upload(
             log.info(
                 "Customer is using joined=False feature", extra=dict(flag_used=flag)
             )
-            joined = False
+            joined = False  # TODO: make this work for parallel
```
daniel-codecov, line 130: I have no idea what this `joined` feature is. This is an example of what hopefully could've been surfaced from this verification experiment. 

I feel like there very well a lot of small features like this that I'm not able to reproduce/know of locally.

services/report/__init__.py
```diff
@@ -281,6 +283,21 @@ async def initialize_and_save_report(
                 # This means there is a report to carryforward
                 self.save_full_report(commit, report, report_code)
 
+                # Behind parallel processing flag, save the CFF report to GCS so the parallel variant of
+                # finisher can build off of it later. Makes the assumption that the CFFs takes the first
+                # 0 to i session ids where i is the number of CFFs to be carried forward.
```
daniel-codecov, line 288: typo: technically i - 1

matt-codecov (commented):
the session id stuff is really gnarly. great work figuring out a) that a solution was necessary and b) a solution haha

i have a few notes on the redis approach:
- ezpz: you should also include the repo id in the redis key name. otherwise two forks uploading for a shared commit will collide
- as-is, the longer the service runs, the more keys we'll accumulate in redis that are probably never going to be used again a couple hours after they're created. could you [add a TTL](https://redis-py-doc.readthedocs.io/en/master/#redis.Redis.pexpire)? maybe reset the TTL each place you call `incrby`. i think this TTL would also automatically take care the cleanup you're doing in `flush_repo`

services/report/parser/__init__.py
```diff
@@ -3,7 +3,9 @@
 from services.report.parser.version_one import VersionOneReportParser
 
 
-def get_proper_parser(upload: Upload):
+def get_proper_parser(upload: Upload, parallel_idx=None):
+    if parallel_idx is not None:
+        return LegacyReportParser()
```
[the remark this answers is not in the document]
daniel-codecov: during the old behaviour, processor task will [rewrite the raw uploaded coverage report](https://github.com/codecov/worker/blob/a01c215e950db16f121df51ce79a86420edf4c6c/tasks/upload_processor.py#L259) to a human readable format. Hence, when the parallel flow starts after serial flow has completely finished, it will have to parse the human readable format instead of the raw uploaded report.

Looking back, maybe it would be better for the experiment to use the raw uploaded report format as it is closer to what we would expect when we fully rollout parallel upload processing

tasks/flush_repo.py
```diff
@@ -151,10 +155,29 @@ def _delete_label_analysis(self, db_session: Session, commit_ids, repoid: int):
 
     @sentry_sdk.trace
     def _delete_commits(self, db_session: Session, repoid: int) -> int:
-        deleted_commits = db_session.query(Commit).filter_by(repoid=repoid).delete()
+        commits_to_delete = db_session.query(Commit).filter_by(repoid=repoid).all()
+        commit_ids_to_delete = [commit.commitid for commit in commits_to_delete]
+
+        # Reset parallel session id redis key for each commit
+        for commit in commits_to_delete:
```
daniel-codecov, line 162: even though there is a TTL, I think we should still reset redis keys explicitly here. Otherwise if a user tries to do uploads for a commit they just erased, the session IDs will be higher than they need to.

tasks/flush_repo.py
```diff
@@ -151,10 +155,32 @@ def _delete_label_analysis(self, db_session: Session, commit_ids, repoid: int):
...
+        commit_ids_to_delete = [commit.commitid for commit in commits_to_delete]
+
+        pipeline = get_redis_connection().pipeline()
+        for id in commit_ids_to_delete:
+            pipeline.set(
+                get_parallel_upload_processing_session_counter_redis_key(
+                    repoid=repoid, commitid=id
+                ),
+                0,
+            )
```
matt-codecov, line 170: i'm not positive but i think `set()` will undo any expiry that was set on the key. i think this method takes a keyword argument `ex` where you can pass in the expiry https://redis-py.readthedocs.io/en/stable/commands.html#redis.commands.core.CoreCommands.set

tasks/upload_finisher.py
```diff
@@ -325,6 +393,117 @@ def invalidate_caches(self, redis_connection, commit: Commit):
...
+                "upload_pk": partial_report["upload_pk"],
+            }
+
+        def merge_report(cumulative_report, obj):
+            incremental_report = obj["report"]
+            parallel_idx = obj["parallel_idx"]
+            assert len(incremental_report.sessions) == 1
+            session_id, session = cumulative_report.add_session(
+                incremental_report.sessions[parallel_idx]
+            )
```
matt-codecov, line 471: shouldn't this use `use_id_from_session`? otherwise if you insert something with `parallel_idx=15` before something with `parallel_idx=12` it'll get mixed up?

tasks/upload_processor.py
```diff
@@ -142,21 +175,26 @@ def process_impl_within_lock(
...
-        log.info(
-            "Obtained upload processing lock, starting",
-            extra=dict(
-                repoid=repoid,
-                commit=commitid,
-                parent_task=self.request.parent_id,
-                report_code=report_code,
-            ),
-        )
+        if not PARALLEL_UPLOAD_PROCESSING_BY_REPO.check_value(repoid) and in_parallel:
```
matt-codecov, line 182: so if we AREN'T in the experiment but we WERE told to run this task in parallel? when would that happen? do we need this second condition?

matt-codecov (approved):

tasks/flush_repo.py
daniel-codecov: Hmm... Are you saying this because it is not sufficient to set a key explicitly to 0? because it will still take up space? 

So what I need to do is set the key to 0, but still set a ttl aswell?

tasks/upload_finisher.py
daniel-codecov: good catch... I'm confused as to why this didn't raise problems earlier...

tasks/upload_processor.py
daniel-codecov: thanks for the catch, we don't need to second condition

daniel-codecov: ah forgot the parenthesis: it should be `not check_value() or not in_parallel` which is equivalent to `not (check_value() and in_parallel)`

daniel-codecov: the only time we don't take the lock is when `PARALLEL_UPLOAD_PROCESSING_BY_REPO.check_value(repoid) and in_parallel` is true

PR #348, opened 2024-03-27
update shared to latest

adrian-codecov:
This updates shared to use latest change that includes SHA after model migration

### Legal Boilerplate


nora-codecov (approved):

django_scaffold/management/commands/migrate.py
```diff
@@ -1,84 +1,2 @@
...
-            django_transaction.commit(database)
-        except:
-            log.info("Codecov migrations failed.")
-            raise
-        else:
-            log.info("Codecov migrations succeeded.")
-        finally:
-            lock.release()
+# We inherit the migrate command from Shared
+from shared.django_apps.legacy_migrations.management.commands.migrate import *
```
matt-codecov, line 2: if we actually install the new apps in `INSTALLED_APPS`, is this file still necessary at all?

matt-codecov (commented):
any reason not to install the apps in `INSTALLED_APPS` in this PR?

django_scaffold/management/commands/migrate.py
adrian-codecov: > any reason not to install the apps in INSTALLED_APPS in this PR?

Adding the apps led to those errors we were seeing locally. I intend to add those in a followup PR, in this one I just wanted to ensure the app wouldn't break if we updated to the latest shared.

> if we actually install the new apps in INSTALLED_APPS, is this file still necessary at all?

That is a good point. I was treating this the same way as the api changes, but it's not necessarily the case cause I'd install the legacy_app from scratch. I think what would make sense for this PR is to get rid of this change ^ and in the followup PR test if I need this at all, good catch. Lmk if that sounds good and I'll change this 1st thing in the morning

matt-codecov (approved):
i think we are on the same page about everything

PR #349, opened 2024-03-27
fix: pass iterable into any() instead of lambda

matt-codecov:
`'function' object is not iterable` issue in sentry has been showing up for a few hours

### Legal Boilerplate


nora-codecov (approved):
👍

adrian-codecov (approved):
Yeah I think you are correct, I think the intent of this fn is to say "if any of these gh_installations has `is_repo_covered_by_integration == True`, then return true". And this lambda doesn't return a list so what you say makes sense

PR #351, opened 2024-03-28
reclaim Owner.createstamp

giovanni-guidini:
These changes add a createstamp for Owners when they are created in more
scenarios than just logging in. This is a breaking behavior considering legacy
versions of Codecov.

The change itself if simple, but should be taken carefully since I'm directly
ignoring a comment in the code. So let's dive into the details.

Why are you doing this?

Context is codecov/engineering-team#1387.
We want to pin new behavior (patch-centric comment and statuses) only to new
users. new in this context mean "was created after a certain date".
To do that we need to know when users were created. Instead of creating another
column for created_at I am simply reusing this existing one.

Isn't it dangerous?

Well yes, there is a possibility that something unexpected will happen.
But the "legacy system" that the comment alluded to is no longer being used :D
(it was codecov-io). And I searched the current system and it doesn't seem to
be using the createdstamp for anything... so maybe we're fine.

Is there an alternative?

Well yes, creating a migration to have a repeated column.
Obviously it'd be a risky migration, so now we're just taking on a different risk.

related: https://github.com/codecov/codecov-api/pull/471

### DIFF
--- a/database/models/core.py
+++ b/database/models/core.py
@@ -52,8 +52,7 @@ class Owner(CodecovBaseModel):
     plan_activated_users = Column(
         postgresql.ARRAY(types.Integer), server_default=FetchedValue()
     )
-    # createstamp seems to be used by legacy to track first login
-    # so we shouldn't touch this outside login
+
     createstamp = Column(types.DateTime, server_default=FetchedValue())
     admins = Column(postgresql.ARRAY(types.Integer), server_default=FetchedValue())
     permission = Column(postgresql.ARRAY(types.Integer), server_default=FetchedValue())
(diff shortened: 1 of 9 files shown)


database/models/core.py
```diff
@@ -52,8 +52,7 @@ class Owner(CodecovBaseModel):
     plan_activated_users = Column(
         postgresql.ARRAY(types.Integer), server_default=FetchedValue()
     )
-    # createstamp seems to be used by legacy to track first login
-    # so we shouldn't touch this outside login
```
adrian-codecov, line 56: This seems also like a bad way to track login only in the first place

adrian-codecov (approved):
So previous to this, do owners only get that created_at when going through the login flow?

PR #356, opened 2024-04-02
feat: ramp up process for ghapp installations

giovanni-guidini:
These changes introduce a change in weight for installations recently updated.
Any change in a ghapp `updated_at` field will cause it to enter a "ramp up" phase
during which other suitable installations will be prefered for usage.

The idea is that changes to an installation can introduce config errors and cause
the ghapp to be unusable in it's current state. To avoid breaking the app we use
the ramp up process.

The rampup is based on `random.choices` and giving weights to installs based on how
long ago they were last updated. The rap up function is exponential and the whole
process stabilizes after 10 days (at which point all installations have the same
probability of being selected)

closes: codecov/engineering-team#1483

### Legal Boilerplate

### DIFF
--- a/services/bots.py
+++ b/services/bots.py
@@ -1,19 +1,37 @@
 import logging
-from random import shuffle
-from typing import Dict, Optional, Tuple
+import random
+from datetime import UTC, datetime
+from typing import Dict, List, Optional, Tuple
 
 from shared.config import get_config
 from shared.torngit.base import TokenType
 
 from database.models import Owner, Repository
-from database.models.core import GITHUB_APP_INSTALLATION_DEFAULT_NAME
+from database.models.core import (
+    GITHUB_APP_INSTALLATION_DEFAULT_NAME,
+    GithubAppInstallation,
+)
 from helpers.environment import is_enterprise
 from helpers.exceptions import OwnerWithoutValidBotError, RepositoryWithoutValidBotError
-from helpers.github_installation import get_installation_name_for_owner_for_task
 from services.encryption import encryptor
 from services.github import get_github_integration_token
 
 log = logging.getLogger(__name__)
+
+
+MAX_GITHUB_APP_SELECTION_WEIGHT = 1200
+
+
+def _get_installation_weight(installation: GithubAppInstallation) -> int:
+    """The weight for a given app installation.
+    Establishes an exponential ramp-up period for installations after being updated.
+    """
+    age = datetime.now(UTC) - installation.updated_at
+    if age.days >= 10:
+        return MAX_GITHUB_APP_SELECTION_WEIGHT
+    # seconds_in_hour = 3600
+    age_hours = (age.seconds // 3600) + age.days * 24
+    return age_hours + 2**age.days
 
 
 def get_owner_installation_id(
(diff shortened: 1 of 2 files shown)


hootener:
**Test Failures Detected**: Due to failing tests, we cannot provide coverage reports at this time.

### :x: Failed Test Results: 
Completed 1 tests with **`1 failed`**, 0 passed and 0 skipped.
<details><summary>View the full list of failed tests</summary>

| **Test Description** | **Failure message** |
| :-- | :-- |
| <pre>Testsuite:<br>pytest<br>Test name:<br>::<br>Envs:<br>- default<br></pre> | <pre>No failure message available</pre> |

services/bots.py
```diff
@@ -1,21 +1,39 @@
...
+
+
+def _get_installation_weight(installation: GithubAppInstallation) -> int:
+    """The weight for a given app installation.
+    Establishes an exponential ramp-up period for installations after being updated.
+    """
+    age = datetime.now(UTC) - installation.updated_at
+    if age.days >= 10:
+        return MAX_GITHUB_APP_SELECTION_WEIGHT
+    # seconds_in_hour = 3600
```
joseph-sentry, line 32: [nit] make sure to use this on line below

services/bots.py
```diff
@@ -1,21 +1,39 @@
...
+    age_hours = (age.seconds // 3600) + age.days * 24
+    return age_hours + 2**age.days
```
joseph-sentry, line 34: how did you decide what to do for this function?

services/bots.py
```diff
@@ -30,18 +48,34 @@ def get_owner_installation_id(
...
+        )
+        already_checked = dict()
         # filter is an Iterator, so we need to scan matches
-        for app_installation in default_app_installation_filter:
+        for app_installation in apps_to_consider:
+            if already_checked.get(app_installation.installation_id):
+                continue
+            already_checked[app_installation.installation_id] = True
             if repository:
                 if app_installation.is_repo_covered_by_integration(repository):
```
joseph-sentry, line 80: [unlikely edge case]

should the apps_to_consider also have the restriction that the repo is covered by integration, what if all the apps in the apps to consider aren't covering this repo, but there exists some app that does cover the repo in the original list of github apps?

joseph-sentry (approved):
looks good, left some comments (nothing too urgent)

services/bots.py
giovanni-guidini: 1st iteration it was going to be a linear function. The rampup was then gonna be about 50days, which is too long. And it makes more sense for it to be exponential, cause it starts slow and then ramps up fast.

So I discussed the requirements and acceptable parameters with Trent. We agreed on 1-2% in the first day and a rampo up period of 7-10 days. Then I plotted the graph in Desmos to figure out more or less how it was gonna look like.

Once I was satisfied I ran the test with choices many times to check the range of sleections for the new installation to see if it was within the parameters I wanted. Et voilà

services/bots.py
joseph-sentry: makes sense, it's just since the user seems to be in control of the repos that will have perms

PR #358, opened 2024-04-03
feat(test results): error message PR comment if no successful parse

joseph-sentry:
Fixes: https://github.com/codecov/test-results-action/issues/15

### DIFF
--- a/services/comparison/__init__.py
+++ b/services/comparison/__init__.py
@@ -29,7 +29,7 @@ class NotificationContext(object):
     """Extra information not necessarily related to coverage that may affect notifications"""
 
     all_tests_passed: bool
-    test_results_error: TestResultsProcessingError | None = False
+    test_results_error: TestResultsProcessingError | None = None
 
 
 class ComparisonProxy(object):
--- a/services/comparison/__init__.py
+++ b/services/comparison/__init__.py
@@ -12,7 +12,7 @@ from shared.torngit.exceptions import (
     TorngitObjectNotFoundError,
 )
 
-from database.enums import CompareCommitState
+from database.enums import CompareCommitState, TestResultsProcessingError
 from database.models import CompareCommit
 from helpers.metrics import metrics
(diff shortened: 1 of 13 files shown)


giovanni-guidini (dismissed):

services/comparison/__init__.py
```diff
@@ -29,6 +29,7 @@ class NotificationContext(object):
     """Extra information not necessarily related to coverage that may affect notifications"""
 
     all_tests_passed: bool
+    test_results_error: TestResultsProcessingError | None = False
```
giovanni-guidini, line 32: Shouldn't the default value be `None`? To follow the type annotation.

services/notification/notifiers/mixins/message/sections.py
```diff
@@ -105,10 +105,16 @@ async def do_write_section(self, comparison, diff, changes, links, behind_by=Non
 class NewHeaderSectionWriter(BaseSectionWriter):
     def _possibly_include_test_result_setup_confirmation(self, comparison):
         if comparison.all_tests_passed():
-            yield ("")
-            yield (
-                ":white_check_mark: All tests successful. No failed tests found :relaxed:"
-            )
+            if comparison.test_results_error():
```
giovanni-guidini, line 108: So... if there are no tests then we assume all tests passed? Crazy... Can we fix that? Maybe by checking that there's at least 1 test in the "all tests passed".
I didn't realize this was a possibility before 😅

tasks/notify.py
```diff
@@ -328,6 +328,11 @@ def run_impl_within_lock(
                     test_result_commit_report is not None
                     and test_result_commit_report.test_result_totals is not None
```
giovanni-guidini, line 329: Could you change this one too?... Doesn't make sense to have "all_tests_passed" when actually no tests passed :E

nora-codecov (approved):

PR #359, opened 2024-04-04
fix: Remove metrics that are failing

giovanni-guidini:
Fix issue https://l.codecov.dev/FSbOor

Why remove metrics? Because I realized that we're not actually hitting the database
on those calls. There's no `first()` or `all()`. It's just a filter.

### Legal Boilerplate


joseph-sentry:
what does this filter actually accomplish, if we aren't hitting the db, then what is the value of `report_ids`?

giovanni-guidini:
> what does this filter do? if we aren't hitting the db, what is the value of `report_ids`?

I believe it's like the Django QueryBuilder? In that you specify this filter, and use it later on in another filter. Then SQLAlchemy would probably expand it as a subfilter / subquery in the other query.

trent-codecov (approved):

PR #361, opened 2024-04-04
fix: make it usable without timeseries

zetaab:
currently its not possible to run worker without "timeseries" database handler. This PR makes it possible to run codecov without that.

fixes https://github.com/codecov/feedback/issues/318

### Legal Boilerplate


flarochelleupgrade:
Running into this with 24.4.1 as well; hoping this will go through soon! 🙏

zetaab:
cc @matt-codecov as you have implemented those parts of the code

matt-codecov (approved):

PR #363, opened 2024-04-04
feat: parallel upload processing verification task

daniel-codecov:
Adds another celery task that will be fired at the end of `UploadFinisher` task for parallel upload processing. The purpose of this task is to compare the results of the serial flow with the parallel flow to determine if the results generated from parallel upload processing are correct. The outcome of the verification will be logged, which is how we can search for them.

I've removed fields from the report_json that are not relevant to the correctness of the report (like timestamp) so that the diff is not affected. 

works off of this: https://github.com/codecov/worker/pull/347

### Legal Boilerplate


tasks/parallel_verification.py
```diff
@@ -0,0 +1,183 @@
...
+
+# To filter out values not relevant for verifying report content correctness. We
+# don't want these values to be the reason why there exists a diff, since it's
+# not actually related to coverage report contents
+def sort_and_stringify_report_json(d):
+    for session_id in d["sessions"].keys():
+        d["sessions"][session_id].pop("d", None)  # remove timestamp
+        d["sessions"][session_id].pop("e", None)  # remove env
+        d["sessions"][session_id].pop("p", None)  # remove state
+        d["sessions"][session_id].pop("se", None)  # remove session extras
```
matt-codecov, line 165: python passes function arguments by reference so this will actually modify the passed-in `d` variable, and if `d` is or contains a reference to a db model field, you could accidentally overwrite real data by doing this

your parallel and non-legacy cases are safe, but your legacy case could potentially remove these fields from the saved json of a real session. not sure - maybe django returns copies or the `ArchiveField` thing saves us or we'd have to explicitly tell django to save or something

to be safe, probably add `import copy` and add `d = copy.deepcopy(d)` before the for loop :v

tasks/parallel_verification.py
```diff
@@ -0,0 +1,183 @@
...
+
+        # round the coverage precentage to 2 decimals (because we're inconsistent somewhere)
+        if "t" in d["sessions"][session_id]:
+            d["sessions"][session_id]["t"][5] = str(
+                round(float(d["sessions"][session_id]["t"][5]), 2)
+            )
+
+    # prune out totals since they're calculated deterministically from report contents
+    for file_name in d["files"].keys():
+        d["files"][file_name] = d["files"][file_name][:3]
```
matt-codecov, line 175: would we not expect file totals to be identical?

matt-codecov (commented):
mostly great, a Q about stripping totals out for comparison

tasks/parallel_verification.py
daniel-codecov: we would... tbh I haven't figured out how to recalculate it yet

daniel-codecov: figured it out!

matt-codecov (approved):

PR #372, opened 2024-04-09
fix: except LockRetry instead of LockError

joseph-sentry:
Fixes: https://codecov.sentry.io/issues/5057432400/?project=-1&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=1&utc=true

We should be excepting a `LockRetry` here because that is what `LockManager` throws when it hits an error when trying to acquire the lock


giovanni-guidini (approved):
❤️

PR #373, opened 2024-04-09
add metric + experiment to see if longer CFF search range lowers errors

matt-codecov:
when looking for an ancestor commit to take our carryforward flag data from, we only search back 10 commits. if we go 10 commits without reports, or if the last 10 commits are still processing, CFF fails and the flags that were being CF'd will be dropped until they receive a new upload

this PR sets up an experiment to plug in different limits and see if any of them have an effect on carryforward success. it creates two metrics that we can watch:
- a "base not found" metric which should be lower in the test group
- a "carryforward success" metric which should be higher in the test group

### Legal Boilerplate

### DIFF
--- a/helpers/telemetry.py
+++ b/helpers/telemetry.py
@@ -2,6 +2,7 @@ import asyncio
 from datetime import datetime
 
 import django
+from asgiref.sync import sync_to_async
 from shared.django_apps.pg_telemetry.models import SimpleMetric as PgSimpleMetric
 from shared.django_apps.ts_telemetry.models import SimpleMetric as TsSimpleMetric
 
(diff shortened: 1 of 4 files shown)


helpers/telemetry.py
```diff
@@ -108,6 +109,10 @@ def populate(self):
 
         self.populated = True
 
+    @sync_to_async
+    def log_simple_metric_async(self, name: str, value: float):
+        self.log_simple_metric(name, value)
+
```
matt-codecov, line 115: turns out i don't use this and forgot to clean it up

nora-codecov (approved):

PR #376, opened 2024-04-10
feat: minor adjustments to gh backfill task

adrian-codecov:
Making some small adjustments:

- Use latest shared to enable jwt token
- Add 'name' to the github app (and app_id although it's not required)
- Refactor the "repository_selection = all" functionality to be used for brand new apps as well

### Legal Boilerplate


tasks/backfill_gh_app_installations.py
```diff
@@ -54,6 +60,28 @@ def add_repos_service_ids_from_provider(
...
+    ):
+        remote_gh_app_installation = async_to_sync(
+            owner_service.get_gh_app_installation
+        )(installation_id=gh_app_installation.installation_id)
+        repository_selection = remote_gh_app_installation.get(
+            "repository_selection", ""
+        )
+        if repository_selection == "all":
+            gh_app_installation.repository_service_ids = None
+            db_session.commit()
```
matt-codecov, line 77: i wouldn't expect this side effect from a function named `is_condition_true()`, consider renaming to something like `maybe_set_installation_to_all_repos()`

matt-codecov (approved):

PR #381, opened 2024-04-12
add logs + move experiment code to helpers

daniel-codecov:
Adds more logging for the parallel upload processing experiment. Also refactors the code in `tasks/upload.py` for readability. 

  Sentry/Codecov employees and contractors can delete or ignore the following.

-->

### Legal Boilerplate

### DIFF
--- a/helpers/parallel_upload_processing.py
+++ b/helpers/parallel_upload_processing.py
@@ -1,0 +1,64 @@
+import copy
+
+from shared.utils.sessions import SessionType
+
+from database.models.reports import Upload
+
+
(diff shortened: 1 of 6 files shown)


helpers/parallel_upload_processing.py
```diff
@@ -0,0 +1,64 @@
+import copy
+
+from shared.utils.sessions import SessionType
+
+from database.models.reports import Upload
+
```
daniel-codecov, line 6: copied from `tasks/upload.py`

nora-codecov (approved):

PR #384, opened 2024-04-16
fix carryforward metric, mock feature/metrics in most tests

matt-codecov:
`worker_service_report_carryforward_success` and `worker_service_report_carryforward_base_not_found` were broken because of a `synctoasync` thing. this fixes them

to satisfy tests, it also adds new test fixtures that will:
- mock `MetricContext.populate()` to skip db accesses and things a test may not be set up for
- mock `Feature.check_value()` to make it return defaults

these fixtures are on by default, but you can disable them by adding `@pytest.mark.real_feature` or `@pytest.mark.real_metric_context` to the test.

### Legal Boilerplate

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -298,3 +298,29 @@ def mock_checkpoint_submit(mocker, request):
         "submit_subflow",
         mock_submit,
     )
+
+
+@pytest.fixture(autouse=True)
+def mock_metric_context(mocker, request):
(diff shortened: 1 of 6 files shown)


joseph-sentry (approved):
It'd be good to document the explanation shared in slack in a comment somewhere around where attempt_log_simple_metric is defined

adrian-codecov (approved):

PR #387, opened 2024-04-16
feat: add ability to create random name for uploads missing a name

joseph-sentry:
Fixes: https://github.com/codecov/internal-issues/issues/133


tasks/upload.py
```diff
@@ -492,6 +493,7 @@ def run_impl_within_lock(
             self.retry(countdown=60, kwargs=kwargs)
         argument_list = []
 
+        name_creator = NameCreator()
```
nora-codecov, line 496: [nit] seems kinda random to instantiate `NameCreator` here - why not do this in the `if` statement where you need it, or earlier in the init method for this class?

nora-codecov (approved):

tasks/upload.py
joseph-sentry: I ended up going with a possibly more simple pattern ([The global object pattern](https://python-patterns.guide/python/module-globals/)). I create a "default" instance in the `helpers.name_creator` and import that same instance to be used everywhere.

joseph-sentry:
Closing this in favour of https://github.com/codecov/codecov-cli/pull/421

PR #391, opened 2024-04-17
1443 first pull adjustment

adrian-codecov:
We have a specific PR message for the first time you upload a PR comment. The original approach preconceived we'd only have coverage specific pulls, but that changed since since bundle analysis was introduced and the first PR experience can be broken. This change adjusts the logic to account for that, rename the method to make more sense and adjust necessary tests. I left some explanations as docstrings in the adjusted method for clarity

### Legal Boilerplate


joseph-sentry (approved):
good explanation in the comment 👍

PR #399, opened 2024-04-19
Add task group to metrics

giovanni-guidini:
We have task health metrics (success, failure, ...)
But we also have a lot of tasks. And the graphs in grafana don't look amazing with all those tasks there.
In particular I'd like to have general success / failure rates per task group.
"task_group" is the task_config_group (part of the task name).

This can be used so we know at any given time how's the general health for the group as a whole.

### Legal Boilerplate

### DIFF
--- a/tasks/base.py
+++ b/tasks/base.py
@@ -26,12 +26,12 @@ log = logging.getLogger("worker")
 REQUEST_TIMEOUT_COUNTER = Counter(
     "worker_task_counts_timeouts",
     "Number of times a task experienced any kind of timeout",
-    ["task"],
+    ["task", "task_group"],
 )
 REQUEST_HARD_TIMEOUT_COUNTER = Counter(
     "worker_task_counts_hard_timeouts",
     "Number of times a task experienced a hard timeout",
-    ["task"],
+    ["task", "task_group"],
 )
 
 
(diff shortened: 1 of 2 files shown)


tasks/base.py
```diff
@@ -114,20 +123,33 @@ def on_timeout(self, soft: bool, timeout: int):
 class BaseCodecovTask(celery_app.Task):
     Request = BaseCodecovRequest
 
-    def __init_subclass__(cls, name=None):
+    def __init_subclass__(cls, name="unknown_task"):
         cls.name = name
+        # All task names follow the format `app.[cron|task].<task_group>.<task>`
+        task_group = name.split(".")[-2] if name != "unknown_task" else "unknown_group"
```
joseph-sentry, line 129: does this give an error if the length of `name.split('.')` is 1?

tasks/base.py
```diff
@@ -42,49 +42,58 @@ def metrics_prefix(self):
 
     def on_timeout(self, soft: bool, timeout: int):
         res = super().on_timeout(soft, timeout)
+        task_group = (
+            self.name.split(".")[-2] if self.name is not None else "unknown_group"
```
joseph-sentry, line 46: [nit] this could be in a function since you are doing it somewhere else as well

matt-codecov: also are we positive all task names definitely fit this pattern?

matt-codecov (commented):
can sort of do this with label filters in grafana. filter where task matches some regex, or exactly matches something in a set

giovanni-guidini:
> can sort of do this with label filters in grafana. filter where task matches some regex, or exactly matches something in a set

Do you mean by that that I can extract a new label from the task name and use this new label in filters OR that I'd have to use 1 expression per task_group and edit all dashboards that do that if a new group is introduced?

If it's option 1 I'm OK with that. If it's option 2 I'd rather make my life simpler with this extra label

tasks/base.py
giovanni-guidini: Yes, but all tasks are supposed to have a `task_group` in their name (separated by `'.'`. Check [the celery config](https://github.com/codecov/shared/blob/main/shared/celery_config.py)

tasks/base.py
giovanni-guidini: They are supposed to fit this pattern. It's the pattern we use for the [task routes in celery config](https://github.com/codecov/shared/blob/main/shared/celery_config.py)

I can add a try/except just in case

tasks/base.py
```diff
@@ -35,16 +36,22 @@
 )
 
 
+def _get_task_group_from_name(task_name: Optional[str]) -> Optional[str]:
+    if task_name is None or ("." not in task_name):
+        return None
```
joseph-sentry, line 41: [nit] do we want this to return `None` or `"unknown_group"`

joseph-sentry (approved):
on last nit comment

tasks/base.py
giovanni-guidini: At first I thought "unknown_group" but after checking the existing dashboards for the tasks I noticed that it's possible to have the task_name there are `None`... so just being consistent with returning `None` as well

giovanni-guidini:
As mentioned by @matt-codecov it's possible to get what I wanted using the [label_replace](https://prometheus.io/docs/prometheus/latest/querying/functions/#label_replace) function in grafana.

So no need to send this extra tag in the metric

PR #400, opened 2024-04-19
  Update `sqlparse` to 0.5.0

michelletran-codecov:
This fixes [https://github.com/advisories/GHSA-2m57-hf25-phgg](https://github.com/advisories/GHSA-2m57-hf25-phgg).

`requirements.txt` was auto-generated by running:

`pip-compile --upgrade-package sqlparse`

The only dependency, `django`, doesn't have `sqlparse` pinned, so we should be able to upgrade this.

### Legal Boilerplate

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ test.integration:
 
 
 update-requirements:
-	pip install pip-tools==6.1.0
+	pip install pip-tools==7.4.1
 	pip-compile requirements.in
 
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -363,9 +363,7 @@ requests==2.31.0
 respx==0.20.2
     # via -r requirements.in
 rfc3986[idna2008]==1.4.0
-    # via
-    #   httpx
-    #   rfc3986
+    # via httpx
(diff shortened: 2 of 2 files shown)


joseph-sentry (approved):

Makefile
```diff
@@ -58,7 +58,7 @@ test.integration:
 
 
 update-requirements:
-	pip install pip-tools==6.1.0
+	pip install pip-tools==7.4.1
```
michelletran-codecov, line 61: Is there a reason for pinning `pip-tools`? 

`6.1.0` has a conflict with newer versions of `pip`, so I upgraded `pip-tools` instead. AFAICT, this make target is only used for development.

joseph-sentry: i think the goal is consistency across all of our machines, if we were running different enough versions of the same tool, the result would probably diverge. 

To be honest though, I don't use this make command and I'm not sure if the other devs do, and I've just been installing pip-tools unpinned when I need it :/

michelletran-codecov: Cool thanks! I'll leave this pinned to a new version for now then.

PR #405, opened 2024-04-23
hotfix: sync_repos using integration when the repo services is known is acting strange

giovanni-guidini:
These changes:

* fix a bug in the sync_repos with integration when the list of repos affected is known.
* makes the previous behavior (of listing all repos) active again even in case we have the list of affected repos. This is so that the app can fix itself if we have difference between the repos we think the installation covers and the repos we can acutally list using it.

### DIFF
--- a/tasks/sync_repos.py
+++ b/tasks/sync_repos.py
@@ -137,7 +137,7 @@ class SyncReposTask(BaseCodecovTask, name=sync_repos_task_name):
         db_session,
         git,
         owner: Owner,
-        repository_service_ids: Optional[List[Tuple[str, str]]],
+        repository_service_ids: Optional[List[Tuple[int, str]]],
     ):
         repoids_added = []
         # Casting to str in case celery interprets the service ID as a integer for some reason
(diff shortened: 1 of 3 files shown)


joseph-sentry (dismissed):

tasks/sync_repos.py
```diff
@@ -161,12 +161,15 @@ async def sync_repos_affected_repos_known(
...
+                repository_service_ids=repository_service_ids,
             ),
         )
 
         # Get info from provider on the repos we don't have
         repos_to_search = [
-            x[1] for x in repository_service_ids if x[0] in missing_repo_service_ids
+            x[1]
+            for x in repository_service_ids
+            if str(x[0]) in missing_repo_service_ids
```
matt-codecov, line 172: i don't see where you are changing the type of this input in this PR. was this broken the whole time?

matt-codecov (approved):
fixes https://github.com/codecov/engineering-team/issues/1519

tasks/sync_repos.py
giovanni-guidini: I believe it was broken this whole time, yes.
I know, my bad 😓

PR #406, opened 2024-04-23
fallback to assuming repo is covered

giovanni-guidini:
I suspect that we might have a little bug in some situations when syncing repos using integration.

Let's assume that suspicion is true and we might have repos not properly sinced in the installation.
These changes aim to not break the app in a specific scenario:
- Org installs Codecov App (not selecting all repos)
- Org admin logs in and sets up a new repo
- For some reason we fail to sync the repo and the installation list of repos. Maybe we missed a webhook? Would be the only possibility really... but maybe the admin logging in would actually make the repo appear in the DB, even missing the webhook.
- App is still broken cause the ORG never logged in, and the repo is not marked as "covered" by the installation

The changes here would fix that scenario by ignoring the repo when selecting the token to communicate with GH.
Ideally we fix the sync process and remove this fix, of course.

Notice that if the owner HAS a token or bot than we use that instead.

### Legal Boilerplate

### DIFF
--- a/services/repository.py
+++ b/services/repository.py
@@ -19,6 +19,7 @@ from database.models.core import (
     GITHUB_APP_INSTALLATION_DEFAULT_NAME,
     GithubAppInstallation,
 )
+from helpers.exceptions import RepositoryWithoutValidBotError
 from helpers.token_refresh import get_token_refresh_callback
 from services.bots import (
     get_owner_installation_id,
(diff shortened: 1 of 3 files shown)


services/repository.py
```diff
@@ -63,7 +64,35 @@ def get_repo_provider_service(
...
+        )
+    except RepositoryWithoutValidBotError:
+        installation_info_without_considering_repo = get_owner_installation_id(
+            repository.owner,
+            False,
+            repository=None,
+            ignore_installation=False,
+            installation_name=installation_name_to_use,
+        )
+        print("INSTALLATION INFO (no repo)", installation_info_without_considering_repo)
```
joseph-sentry, line 80: the content in these print statements can still remain, but they should probably be `log.info`

joseph-sentry (approved):
just one comment that you can fix, the change looks good, the logic is starting to get a little complicated so it may make sense to move some things into their own functions and document the idea behind how we're handling errors using comments later

services/repository.py
giovanni-guidini: Nah, I forgot to remove those :E

services/repository.py
```diff
@@ -51,19 +52,62 @@ def get_repo_provider_service(
...
         ignore_installation=False,
         installation_name=installation_name_to_use,
     )
-    token, token_owner = get_repo_appropriate_bot_token(repository, installation_info)
+    try:
+        token, token_owner = get_repo_appropriate_bot_token(
+            repository, installation_info
+        )
+    except RepositoryWithoutValidBotError:
+        # GitHub exclusive bit -- START
```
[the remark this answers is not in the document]
giovanni-guidini: It's not a problem if we don't check cause it would just raise the exception again, but yeah, makes sense to do that as save a few operations... and have a clearer distinction of logic in the code.

services/repository.py
```diff
@@ -87,9 +131,11 @@ def get_repo_provider_service(
             secret=get_config(service, "client_secret"),
         ),
         on_token_refresh=get_token_refresh_callback(token_owner),
-        fallback_installations=installation_info.get("fallback_installations")
-        if installation_info
-        else None,
+        fallback_installations=(
+            installation_info.get("fallback_installations")
+            if installation_info
+            else None
```
adrian-codecov, line 137: So is to have a fallback option, the gh app we think doesn't have that repo selected correct? 

If there is a fallback, and it indeed doesn't have the repo in that fallback, would it eventually error out?

giovanni-guidini: This was a lint change. It's not directly related to the change here.

We only have fallback installations if the user in question has multiple configured apps (dedicated cloud). In this case you would have installations that cover the repo (more than 1 actually) and not raise RepoBotNotConfigured.

adrian-codecov: Oh I see that makes sense, okok ty

adrian-codecov (approved):

giovanni-guidini:
Now that the sync_repos bug is fixed and we ran the backfill task for gh apps this seems unecessary

PR #408, opened 2024-04-24
fix the sync_integration fallback process with the data from the git provider

giovanni-guidini:
[fix: Properly check if installation is rate limited](https://github.com/codecov/worker/pull/408/commits/17fcf31af26421fe5f7fcca4feda7d756f9ab5cc) 
17fcf31
This fixes how we check if an intallation is rate limited and updates
how we pass info the the Torngit adapter (to comply with the expectations of torngit.GitHub).


requirements.txt
```diff
@@ -1,8 +1,8 @@
 #
-# This file is autogenerated by pip-compile with Python 3.10
+# This file is autogenerated by pip-compile with Python 3.11
```
nora-codecov, line 2: worker is still on python 3.10 right? could you re-do the pip-compile in 3.10, just in case

matt-codecov (approved):

PR #409, opened 2024-04-24
fix: Flake detection failure messages

joseph-sentry:
- If failure message is None ignore the test instance in the UnrelatedMatchesDetector, the information is just not there so we can't use this test instance

- Normalize failure messages in the ingest in UnrelatedMatchesDetector, we should have been doing this already

### DIFF
--- a/services/flake_detection.py
+++ b/services/flake_detection.py
@@ -120,10 +120,10 @@ class TestDictObject:
 
 
 class UnrelatedMatchesDetector(BaseSymptomDetector):
-    def __init__(self, failure_normalizer=None):
+    def __init__(self, failure_normalizer: FailureNormalizer | None = None):
         self.state = defaultdict(lambda: defaultdict(TestDictObject))
         self.res = defaultdict(list)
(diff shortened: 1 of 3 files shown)


services/flake_detection.py
```diff
@@ -119,19 +120,26 @@ def __init__(self):
 
 
 class UnrelatedMatchesDetector(BaseSymptomDetector):
-    def __init__(self):
+    def __init__(self, failure_normalizer=None):
         self.state = defaultdict(lambda: defaultdict(TestDictObject))
         self.res = defaultdict(list)
+        self.failure_normalizer: FailureNormalizer | None = failure_normalizer
```
giovanni-guidini, line 126: Why put the type annotation in the attribute instead of on the function argument?
The attribute type can be derived from the argument, not the other way around.

giovanni-guidini (commented):
Smol question 🙏

services/flake_detection.py
joseph-sentry: very good point

adrian-codecov (approved):

PR #413, opened 2024-04-24
Gh app backfill task extra

adrian-codecov:
Last time this was ran the tasks ran out of memory. I'm wanting to do 2 things to help with this:

1) add yield_per -> helps fetch smaller amount of data per iteration
2) separate the two backfill scenarios into two respective tasks -> more control about what's ran and what isnt

Logic remains the same, mostly moving files and tests arounds, renaming files/tasks, etc

- Created 2 new tasks, backfill_existing_gh_app_installations and backfill_owners_withou_gh_app. It separates the original task into 2 pieces.
  - Added yield_per() in both tasks
- Removed the original backfill_gh_apps_task in favor of the two above
- Reshuffled tests into their respective test files
- Centralized shared logic into helpers/backfills.py file

### Legal Boilerplate


helpers/backfills.py
```diff
@@ -0,0 +1,70 @@
...
+                new_repo_service_ids.append(service_id)
+        log.info(
+            "Added the following repo service ids to this gh app installation",
+            extra=dict(
+                ownerid=ownerid,
+                installation_id=gh_app_installation.installation_id,
+                new_repo_service_ids=new_repo_service_ids,
+            ),
+        )
+        gh_app_installation.repository_service_ids = new_repo_service_ids
```
giovanni-guidini, line 45: Won't this overwrite what might be already there? Or are you sure at this point that the `repository_service_ids` is an empty list (when it shouldn't be)?

helpers/backfills.py
```diff
@@ -0,0 +1,70 @@
...
+    owner_service,
+    gh_app_installation: GithubAppInstallation,
+):
+    remote_gh_app_installation = async_to_sync(owner_service.get_gh_app_installation)(
+        installation_id=gh_app_installation.installation_id
+    )
+    repository_selection = remote_gh_app_installation.get("repository_selection", "")
+    if repository_selection == "all":
+        gh_app_installation.repository_service_ids = None
+        db_session.commit()
```
giovanni-guidini, line 62: Do we want to use `flush` here? I think `commit` ends a transaction in the DB...

helpers/backfills.py
```diff
@@ -0,0 +1,70 @@
...
+        log.info(
+            "Selection is set to all, no installation is needed",
+            extra=dict(ownerid=gh_app_installation.ownerid),
+        )
+        del remote_gh_app_installation
```
giovanni-guidini, line 67: Feels unnecessary to explicitly delete this here given that it's a local variable... Returning from the function should remove it from the stack already, right?

tasks/backfill_existing_gh_app_installations.py
```diff
@@ -0,0 +1,122 @@
...
+from helpers.backfills import (
+    add_repos_service_ids_from_provider,
+    maybe_set_installation_to_all_repos,
+)
+from services.owner import get_owner_provider_service
+from tasks.base import BaseCodecovTask
+
+log = logging.getLogger(__name__)
+
+yield_amount = 100
```
giovanni-guidini, line 18: [very nit] use all_caps for constants (e.g. `YIELD_AMOUNT`)

tasks/backfill_existing_gh_app_installations.py
```diff
@@ -0,0 +1,122 @@
...
+        gh_app_installations: List[
+            GithubAppInstallation
+        ] = gh_app_installations_query.yield_per(yield_amount)
+
+        # I need to make reference of these in the gh installation app, Ill take suggestions here
+        owners_dict = {owner.ownerid: owner for owner in owners}
+
+        for gh_app_installation in gh_app_installations:
+            # Check if gh app has 'all' repositories selected
+            owner = owners_dict[gh_app_installation.ownerid]
```
giovanni-guidini, line 54: Shouldn't you be able to get the owner directly from the `GithubAppInstallation` instance?

https://github.com/codecov/worker/blob/bcec3478e2a27bb7950f40388cf191834fb2d5a3/database/models/core.py#L193

helpers/backfills.py
adrian-codecov: Yeah this won't overwrite the things we have, see here for example: https://github.com/codecov/worker/blob/163e7c9f933163ecfa4fa094863c4b5f68ddeadb/tasks/backfill_existing_gh_app_installations.py#L61-L62. We first check if the selection is all, else get independent repos for each app. Same happens for the other task, https://github.com/codecov/worker/blob/163e7c9f933163ecfa4fa094863c4b5f68ddeadb/tasks/backfill_owners_without_gh_app_installations.py#L83-L84

helpers/backfills.py
adrian-codecov: That's true, I copy pasted the previous iteration that was trying to help with memory but not necessary atm

tasks/backfill_existing_gh_app_installations.py
adrian-codecov: Good catch, will get rid of this

helpers/backfills.py
giovanni-guidini: What I meant is, for example, if some installation `G` has `repository_service_ids={1200}`. Then this logic would overwrite the current selection `{1200}`.

But it doesn't really matter cause you are listing _all_ the repos the installation has access to, which would include `1200`.

tasks/backfill_owners_without_gh_app_installations.py
```diff
@@ -0,0 +1,136 @@
...
+                    extra=dict(ownerid=ownerid),
+                )
+                gh_app_installation = GithubAppInstallation(
+                    owner=owner,
+                    installation_id=owner.integration_id,
+                    app_id=get_config("github", "integration", "id"),
+                    name=GITHUB_APP_INSTALLATION_DEFAULT_NAME,
+                )
+                db_session.add(gh_app_installation)
+                db_session.commit()
```
giovanni-guidini, line 75: I think you should wait to commit this change only after the repository selection is complete. Otherwise if there are issues with the repo selection we'd have a github installation with bad repo selection

giovanni-guidini (approved):
LGTM, left a comment though

giovanni-guidini (approved):

PR #417, opened 2024-04-26
Update `idna` package to 3.7

michelletran-codecov:
This mitigates [GHSA-jjg7-2v4v-x38h](https://github.com/advisories/GHSA-jjg7-2v4v-x38h) vulnerability.

I ran `pip-compile --upgrade-package idna` to generate this. The dependencies also don't pin the exact version, and 3.7 is within the ranges of downstream dependencies.

Link to ticket: https://github.com/codecov/internal-issues/issues/412

### Legal Boilerplate


requirements.txt
```diff
@@ -188,7 +188,7 @@ httpx==0.23.1
     #   shared
 identify==2.5.30
     # via pre-commit
-idna==2.10
+idna==3.7
```
adrian-codecov, line 191: Do you know if the next time someone runs pip-compile, if this will regress to 2.10? Or once 3.7, it will persist?

adrian-codecov: Confirmed this wouldn't be the case

adrian-codecov (approved):

PR #418, opened 2024-04-29
chore: extend UploadFlow with new ending

giovanni-guidini:
It was a suspition that we might be overcounting beginnings in the UploadFlow.
I believe one of the exit paths was not being covered in the `Upload` task.

These changes add a new ending to the UploadFlow. Namely, for the scenario where
an UploadTask finishes without enqueuing other tasks because there are no reports
waiting to be processed.

Also updating the `getutcnow` function that is deprecated and fixing some typos.

### Legal Boilerplate


michelletran-codecov (dismissed):
Just a small nit.

matt-codecov (approved):

PR #422, opened 2024-04-30
fix: always select fallback if available

giovanni-guidini:
Because `random.choices` makes selections with replacement and we were making 3 selections only
there was a chance that an app would vanish from the choices (from not being selected).
This is bad because it hides this app as a possible fallbakc option.

Particularly if the choice is marked currently as "rate limited" we would fail with exception
instead of falling back in the other app. We want the fallback.

These changes go around this "choice with selection" by removing the selected app from the
population after each step. It's maybe slower (in practice the population will be small enough
that this doesn't really matter), but makes sure that we always select all available apps.
And still respectes the weighted selection.

### Legal Boilerplate

### DIFF
--- a/services/bots.py
+++ b/services/bots.py
@@ -38,14 +38,17 @@ def _get_installation_weight(installation: GithubAppInstallation) -> int:
         return MAX_GITHUB_APP_SELECTION_WEIGHT
     seconds_in_hour = 3600
     age_hours = (age.seconds // seconds_in_hour) + age.days * 24
-    return age_hours + 2**age.days
+    # Prevent clock differences from making the weight negative
+    return max(1, age_hours + 2**age.days)
 
 
 def _get_apps_from_weighted_selection(
     owner: Owner, installation_name: str, repository: Optional[Repository]
 ) -> List[GithubAppInstallation]:
-    ghapp_installations_filter: List[GithubAppInstallation] = list(
-        filter(
+    # Map GithubAppInstallation.id --> GithubAppInstallation
+    ghapp_installations_filter: Dict[int, GithubAppInstallation] = {
+        obj.id: obj
+        for obj in filter(
             lambda obj: (
                 obj.name == installation_name
                 and obj.is_configured()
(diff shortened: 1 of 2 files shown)


services/bots.py
```diff
@@ -38,14 +38,17 @@ def _get_installation_weight(installation: GithubAppInstallation) -> int:
         return MAX_GITHUB_APP_SELECTION_WEIGHT
     seconds_in_hour = 3600
     age_hours = (age.seconds // seconds_in_hour) + age.days * 24
-    return age_hours + 2**age.days
+    # Prevent clock differences from making the weight negative
+    return max(1, age_hours + 2**age.days)
 
 
 def _get_apps_from_weighted_selection(
```
michelletran-codecov, line 45: I think it might be useful to add some function documentation around what it's returning because it looks like we're significantly changing the functionality. If I understand the code correctly, we're trying to get an ordered list of apps that's random based on the weighting (so that we don't _always_ use the highest weighted one). The return is no longer just a "random subset of the apps" but an ordered list of apps (with some randomness).

services/bots.py
```diff
@@ -58,23 +61,33 @@ def _get_apps_from_weighted_selection(
...
-    apps_to_consider = (
-        random.choices(ghapp_installations_filter, weights=weights, k=3)
-        if len(ghapp_installations_filter) > 0
-        else []
-    )
+    # We pick apps one by one until all apps have been selected
+    # Obviously apps with a higher weight have a higher change of being selected as the main app (1st selection)
+    # But it's important that others are also selected so we can use them as fallbacks
+    apps_to_consider = []
+    if keys:
```
michelletran-codecov, line 80: Can we remove this `if` check here (i.e. the `while` loop condition should fail if the list is empty)?

michelletran-codecov (commented):
Just a few comments. I think we should add some documentation to this function as it's doing some functionality that's a bit more complex than just `get_apps_from_weighted_selection`.

services/bots.py
giovanni-guidini: That is true... good callout here.

services/bots.py
giovanni-guidini: Sure, I'll refactor this

services/bots.py
```diff
@@ -42,23 +42,43 @@ def _get_installation_weight(installation: GithubAppInstallation) -> int:
...
-                obj.name == installation_name
-                and obj.is_configured()
-                and (
-                    # If there is a repo we want only the apps that cover said repo
-                    (repository and obj.is_repo_covered_by_integration(repository))
-                    # If there is no repo we still need some true value
-                    or (not repository)
-                )
-            ),
+            lambda obj: _can_use_this_app(obj, installation_name, repository),
```
michelletran-codecov, line 81: This is much more readable now! Thanks!

services/bots.py
```diff
@@ -77,21 +97,23 @@ def _get_apps_from_weighted_selection(
...
+        # which we are trying to avoid here. So we remove the key selected and its weight from the population.
+        key_idx = keys.index(selected_app_id)
+        keys.pop(key_idx)
+        weights.pop(key_idx)
+        selections += 1
     if installation_name != GITHUB_APP_INSTALLATION_DEFAULT_NAME:
         # Add the default app as the last fallback if the owner is using a different app for the task
         default_apps = filter(
-            lambda obj: obj.name == GITHUB_APP_INSTALLATION_DEFAULT_NAME,
+            lambda obj: _can_use_this_app(
```
michelletran-codecov, line 114: Just double checking that this is OK. `_can_use_this_app` has extra logic that's not in the original functionality of this code, but those extra checks are required to be able to use the app anyways(?)

michelletran-codecov (approved):
Generally LGTM! I just have a non-blocking question.

services/bots.py
giovanni-guidini: I actually noticed when refactoring that the repo verification was missing from this filter. I think it is necessary to have that too, because if the repo is not covered we can't use the app.

Of course, considering the default installation, we expect all repos (that Codecov has access to) to be covered, but it's better for the logic to be consistent.

giovanni-guidini: About the `is_configured` check is a similar story. We generaly expect the default app to be configured (otherwise it wouldn't have that name actually), but if it's not configured we shouldn't use it.

PR #423, opened 2024-04-30
Remove unused import with ruff

michelletran-codecov:
In particular, I was interested in F401. This should clean up all unused imports in the repo. We have to be careful with celery tasks because they don't get picked up if they are not "imported" anywhere. I've added an annotation on the `__init__` files that import these "unused" tasks.

I've also switched to using ruff for "black" linting, which behaves a bit differently.

### Legal Boilerplate


database/tests/unit/test_events.py
```diff
@@ -1,8 +1,6 @@
 import os
 
-import pytest
-
-import database.events
+import database.events  # noqa: F401
```
adrian-codecov, line 3: Soo, I searched this up and this F401 is an unused imported module, https://www.flake8rules.com/rules/F401.html#:~:text=A%20module%20has%20been%20imported,the%20import%20should%20be%20removed. Do we want to get rid of this line as part of running ruff? Or do we just want to signal it?

(I think we should remove it, maybe it was missed?)

helpers/pathmap/__init__.py
```diff
@@ -1,3 +1,4 @@
+# ruff: noqa: F401
```
adrian-codecov, line 1: Same here about F401

services/notification/notifiers/checks/__init__.py
```diff
@@ -1,3 +1,4 @@
+# ruff: noqa: F401
```
adrian-codecov, line 1: Same here, but now wondering if this should be treated differently for init.py for imports, so this one seems okay?

adrian-codecov (approved):
Quick questions on the imports question but otherwise lgtm!

services/notification/notifiers/checks/__init__.py
michelletran-codecov: Ah, I was trying to be conservative (especially with these imports in `__init__`). I'll try removing some of these and see how it goes.🤞

database/tests/unit/test_events.py
adrian-codecov: Interesting.. that's werid haha, I guess ruff can tell? Maybe there's methods from database.events that are silently used, but ya that's odd

services/notification/notifiers/checks/__init__.py
michelletran-codecov: Hmm... these paths are being imported elsewhere in the code. I would have to update the downstream references if we want to get rid of them here. Since this provides a shorter import path for code importing these modules, so I don't think there's any harm leaving them as is.

adrian-codecov (approved):

PR #426, opened 2024-05-01
Update test instances to use local commitid and branch fields

joseph-sentry:
We want to move the commitid and branch information for a test instance to directly on the object, so that when we query we don't have to join with the commits table.

### DIFF
--- a/tasks/backfill_test_instances.py
+++ b/tasks/backfill_test_instances.py
@@ -1,0 +1,53 @@
+import logging
+
+from shared.django_apps.reports.models import TestInstance
+from sqlalchemy import desc
+
+from app import celery_app
+from celery_config import backfill_test_instances_task_name
+from tasks.base import BaseCodecovTask
+
+log = logging.getLogger(__name__)
+
+
+class BackfillTestInstancesTask(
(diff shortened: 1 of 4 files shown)


tasks/backfill_test_instances.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+
+        test_instance_filter = TestInstance.objects.select_related(
+            "upload__report__commit"
+        ).filter(
+            branch=None,
+            commitid=None,
+        )
+        num_test_instances = test_instance_filter.count()
+        all_test_instances = test_instance_filter.all()
```
giovanni-guidini, line 28: Depending on how many test instances we have this could lead to OOM errors, right? I think there are millions of test instances, after all.
Consider using the `yield_per` here, so we don't get everything into memory at once.

giovanni-guidini: You can use the `yield_per` value as the chunk size and simplify the code a bit.

giovanni-guidini (approved):
please check the comments before merging

tasks/backfill_test_instances.py
giovanni-guidini: Interesting. Today I learned.

PR #439, opened 2024-05-08
feat: add TestResultsFlow for measuring time to notification

joseph-sentry:


helpers/checkpoint_logger/flows.py
```diff
@@ -55,3 +55,17 @@ class UploadFlow(BaseFlow):
     NOTIF_TOO_MANY_RETRIES = auto()
     NOTIF_STALE_HEAD = auto()
     NOTIF_ERROR_NO_REPORT = auto()
+
+
+@failure_events("TEST_RESULTS_ERROR")
+@success_events("TEST_RESULTS_BEGIN")
+@subflows(
+    ("notification_latency", "TEST_RESULTS_BEGIN", "TEST_RESULTS_NOTIFY"),
+    ("notification_latency", "TEST_RESULTS_BEGIN", "FLAKE_DETECTION_NOTIFY"),
```
giovanni-guidini, line 64: should the subflows have the same name?

giovanni-guidini (approved):
It might be interesting to add another checkpoint in the beginning of `TestResultsFinisherTask.process_impl_within_lock` like `TEST_RESULTS_PROCESSING_BEGIN`.

THe subflow `TEST_RESULTS_BEGIN --> TEST_RESULTS_FINISHER_BEGIN` would give us a rough estimate of processing time. (I think that the _finisher_ task only runs _after_ all the processor tasks finish, and that we don't current have a metric for that)

What do you think?


But in general lgtm 👍 
Very exciting to see more Flows

helpers/checkpoint_logger/flows.py
```diff
@@ -55,3 +55,24 @@ class UploadFlow(BaseFlow):
     NOTIF_TOO_MANY_RETRIES = auto()
     NOTIF_STALE_HEAD = auto()
     NOTIF_ERROR_NO_REPORT = auto()
+
+
+@failure_events("TEST_RESULTS_ERROR")
+@success_events("TEST_RESULTS_BEGIN")
+@subflows(
+    ("notification_latency", "TEST_RESULTS_BEGIN", "TEST_RESULTS_NOTIFY"),
```
giovanni-guidini, line 63: what's the difference between "notification_latency" and "test_results_notification_latency" subflows?

joseph-sentry: that was a mistake :P

giovanni-guidini (approved):

PR #440, opened 2024-05-08
feat: change to get new tokenless working

joseph-sentry:
We don't want to update the commit using github's information because that will rename the branch to a different format than what we're aiming for in tokenless

### DIFF
--- a/services/repository.py
+++ b/services/repository.py
@@ -253,9 +253,19 @@ async def update_commit_from_provider_info(repository_service, commit):
             # so we append the branch name with the fork slug
             branch_name = commit_updates["head"]["branch"]
             # TODO: 'slug' is in a `.get` because currently only GitHub returns that info
+
             if commit_updates["head"].get("slug") != commit_updates["base"].get("slug"):
-                branch_name = commit_updates["head"]["slug"] + ":" + branch_name
+                # get rid of repo name in head.slug
+                try:
+                    username = commit_updates["head"]["slug"].split("/")[0]
+                    branch_name = username + ":" + branch_name
+                except:
+                    log.error(
+                        "Error updating branch name for commit originating from fork",
+                        extra=dict(repoid=commit.repoid, commit=commit.commitid),
+                    )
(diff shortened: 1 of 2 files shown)


services/repository.py
```diff
@@ -253,9 +253,11 @@ async def update_commit_from_provider_info(repository_service, commit):
             # so we append the branch name with the fork slug
             branch_name = commit_updates["head"]["branch"]
             # TODO: 'slug' is in a `.get` because currently only GitHub returns that info
-            if commit_updates["head"].get("slug") != commit_updates["base"].get("slug"):
-                branch_name = commit_updates["head"]["slug"] + ":" + branch_name
-            commit.branch = branch_name
+            # if this PR is from a fork don't update the branch here
```
giovanni-guidini, line 256: Why is this the case? Can you maybe share some examples, please?

joseph-sentry: updated the PR to make more sense, basically with the old behaviour this would update the branch name to be:
`<username>/<repo name>:<branch name>` with this change it will be `<username>:<branch name>`

adrian-codecov (approved):
Did we change the branch format we were expecting? Or did we save the slug all this time and realize until now we only wanted the username?

services/repository.py
```diff
@@ -253,9 +253,19 @@ async def update_commit_from_provider_info(repository_service, commit):
...
             # TODO: 'slug' is in a `.get` because currently only GitHub returns that info
+
             if commit_updates["head"].get("slug") != commit_updates["base"].get("slug"):
-                branch_name = commit_updates["head"]["slug"] + ":" + branch_name
+                # get rid of repo name in head.slug
+                try:
+                    username = commit_updates["head"]["slug"].split("/")[0]
+                    branch_name = username + ":" + branch_name
+                except:
+                    log.error(
```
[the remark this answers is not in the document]
joseph-sentry: it's honestly just a visual change, i don't think users need to know what the repo name of the fork is, so it's kinda just in the way

PR #441, opened 2024-05-08
1576 use users measurements

adrian-codecov:
Finally. This is the PR where we get to use the measurements partition table on worker - API one here https://github.com/codecov/codecov-api/pull/539.

Basically, we're inserting upload records on the user measurements table when we insert an upload. For this repo, that happens on the Legacy endpoint where we create an upload record preemptively. We're replacing the setting/reading with the insert_user_measurements and query_user_measurements methods.

On top of that, I'm getting rid of the previous implementation/query that took ages as that's no longer needed. I'm deleting, replacing or adjusting the tests we previously used to adjust to the new implementation.

### Legal Boilerplate


services/report/__init__.py
```diff
@@ -328,6 +330,24 @@ def create_report_upload(
         flags = normalized_arguments.get("flags")
         flags = flags.split(",") if flags else []
         self._attach_flags_to_upload(upload, flags)
+
```
adrian-codecov, line 333: This was done here and not in the base class to only target Coverage type reports

requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.12
+# This file is autogenerated by pip-compile with Python 3.9
```
nora-codecov, line 2: oops, gonna have to redo this one!

adrian-codecov: 💀 a battle fought but done!

nora-codecov (approved):

PR #443, opened 2024-05-08
clean up unnecessary extra db queries in MetricContext

matt-codecov:
clean up unnecessary extra db queries in MetricContext

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/codecov/worker/pull/443).
* __->__ #443
* #442

### DIFF
--- a/helpers/telemetry.py
+++ b/helpers/telemetry.py
@@ -7,7 +7,7 @@ from asgiref.sync import sync_to_async
 from shared.django_apps.pg_telemetry.models import SimpleMetric as PgSimpleMetric
 
 from database.engine import get_db_session
-from database.models.core import Commit, Owner, Repository
+from database.models.core import Commit, Repository
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


helpers/telemetry.py
```diff
@@ -66,31 +62,26 @@ def __init__(
...
-                self.owner_id = owner.ownerid
+                repo = (
+                    dbsession.query(Repository)
+                    .filter(Repository.repoid == self.repo_id)
+                    .first()
+                )
+                self.owner_id = repo.ownerid
 
-            if self.commit_sha:
+            if self.commit_sha and not self.commit_id:
```
adrian-codecov, line 84: Could you have a sha but not an id?

adrian-codecov (approved):

helpers/telemetry.py
matt-codecov: i think so, don't remember where i saw this. i think it was maybe like, a webhook or GH API response or something that didn't include our DB id but did include the sha

PR #447, opened 2024-05-15
Round the coverage numbers before saving to database

michelletran-codecov:
This ensures that we use the user specified rounding strategy before saving to the database. 
This change does _not_ update the database to increase precision; that is too much effort, and this is mean to fix the common default case (of rounding down to 2 decimal places).
There might be a future PR to actually migrate the database for more precision, in which case this code will still be useful for setting that more precise number into the database.

Quick fix for the common default case in https://github.com/codecov/engineering-team/issues/1614

### Legal Boilerplate

### DIFF
--- a/database/models/reports.py
+++ b/database/models/reports.py
@@ -1,5 +1,6 @@
 import logging
 import uuid
+from decimal import Decimal
 from functools import cached_property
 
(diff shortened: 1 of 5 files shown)


joseph-sentry (approved):
lgtm

PR #452, opened 2024-05-15
make telemetry helper resilient to exceptions

matt-codecov:
https://github.com/codecov/internal-issues/issues/461


giovanni-guidini (approved):
Are we fine with having the catch-all exceptions?
Out of curiosity, does having `log.exception` creates a sentry issue?

PR #454, opened 2024-05-16
feat: count difference of uploaded sessions

giovanni-guidini:
ticket: https://github.com/codecov/engineering-team/issues/1622

These changes introduce a new capabiliy to `ComparisonProxy`. Namely, to get the
difference in the number of uploads reported per flag between HEAD and BASE.

The idea is to be able to give more actionable message in the PR comment on the following scenario:
1. I have uploaded 5 reports to my base commit
2. I upload 1 report to head commit
3. I have NOT covered all lines of my patch
4. I'm seeing project coverage go WAY down

Because typically coverage is uploaded via CI, and CI doesn't change often, a different
number of reports present could indicate issues in the coverage results.

We have to ignore carryforward sessions because different commits might upload different
sets of reports.

### Legal Boilerplate

### DIFF
--- a/services/comparison/__init__.py
+++ b/services/comparison/__init__.py
@@ -306,19 +306,17 @@ class ComparisonProxy(object):
                 # We ignore carryforward sessions
                 # Because not all commits would upload all flags (potentially)
                 # But they are still carried forward
-                if session.session_type == SessionType.uploaded:
+                if session.session_type != SessionType.carriedforward:
                     if session.flags == []:
                         session.flags = [""]
                     for flag in session.flags:
-                        existing = per_flag_dict.get(flag)
-                        if existing:
-                            existing[curr_counter] += 1
-                        else:
-                            new_entry = ReportUploadedCount(
+                        dict_value = per_flag_dict.get(flag)
+                        if dict_value is None:
+                            dict_value = ReportUploadedCount(
                                 flag=flag, base_count=0, head_count=0
                             )
-                            new_entry[curr_counter] += 1
-                            per_flag_dict[flag] = new_entry
+                        dict_value[curr_counter] += 1
+                        per_flag_dict[flag] = dict_value
         self._cached_reports_uploaded_per_flag = list(per_flag_dict.values())
         return self._cached_reports_uploaded_per_flag
 
--- a/services/comparison/__init__.py
+++ b/services/comparison/__init__.py
@@ -1,13 +1,14 @@
 import asyncio
 import logging
 from dataclasses import dataclass
-from typing import List, Optional
+from typing import Dict, List, Optional
 
 from shared.reports.changes import get_changes_using_rust, run_comparison_using_rust
 from shared.reports.types import Change
 from shared.torngit.exceptions import (
     TorngitClientGeneralError,
 )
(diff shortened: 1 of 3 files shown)


services/comparison/tests/unit/test_reports_uploaded_count_diff.py
```diff
@@ -0,0 +1,173 @@
...
+        comparison=Comparison(
+            head=FullCommit(report=head_report, commit=None),
+            project_coverage_base=FullCommit(report=base_report, commit=None),
+            patch_coverage_base_commitid=None,
+            enriched_pull=None,
+        )
+    )
+    # Python Dicts preserve order, so we can actually test this equality
+    # See more https://stackoverflow.com/a/39537308
+    assert comparison_proxy.get_reports_uploaded_count_per_flag_diff() == expected_diff
```
michelletran-codecov, line 158: These tests almost look the same as the ones on line 65 (including the parametrized valuses). Can we just add a the results and the extra assert call for `get_reports_uploaded_count_per_flag_diff`?

services/comparison/tests/unit/test_reports_uploaded_count_diff.py
```diff
@@ -0,0 +1,173 @@
...
+
+import pytest
+from shared.reports.resources import Report
+from shared.utils.sessions import Session, SessionType
+
+from services.comparison import ComparisonProxy
+from services.comparison.types import Comparison, FullCommit, ReportUploadedCount
+
+
+@pytest.mark.parametrize(
```
michelletran-codecov, line 11: When adding parametrize tests, I like to [add ids](https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids) for each test case to more clearly communicate what the test is checking.

michelletran-codecov (commented):
Just a few comments

services/comparison/tests/unit/test_reports_uploaded_count_diff.py
giovanni-guidini: this is so cool... I didn't know about that. Thanks for sharing, will add and ID on those

services/comparison/__init__.py
```diff
@@ -280,6 +283,65 @@ async def get_impacted_files(self):
...
+        head_report = self.comparison.head.report
+        ops = [(base_report, "base_count"), (head_report, "head_count")]
+        for curr_report, curr_counter in ops:
+            for session in curr_report.sessions:
+                # We ignore carryforward sessions
+                # Because not all commits would upload all flags (potentially)
+                # But they are still carried forward
+                if session.session_type == SessionType.uploaded:
+                    if session.flags == []:
+                        session.flags = [""]
```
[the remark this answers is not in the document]
giovanni-guidini: A session with no flag is one that is uploaded without using the (optional) `--flag` argument in the CLI or uploader. They are not incommon if you have only a simple project and don't separates your tests.

I think it's extremely unlikely that `""` is used as an actual flag because it doesn't carry any information in it. In essence there's little difference uploading with a `""` flag and no flag. I don't think we _need_  to handle it separately.
A good callout though.

PR #457, opened 2024-05-21
refactor: Comment conditions

giovanni-guidini:
These are proposed changes to move much of the verifying logic of "can we send this comment?"
to a dedicated `NotificationCondition` class.

This is inspired by the authentication classes of rest framework (like we use in api), but
more explicit. My hope is that we can expand this for other notifiers as well, and hopefully
reuse some of the notify conditions accross notifiers.

But I'd like to test the waters (== team's appetite for these changes) before commiting
a bunch more time in refactoring more code.

### Legal Boilerplate

### DIFF
--- a/services/notification/notifiers/comment/conditions.py
+++ b/services/notification/notifiers/comment/conditions.py
@@ -1,0 +1,141 @@
+import logging
+from abc import ABC, abstractmethod
+
+from shared.torngit.exceptions import (
+    TorngitClientError,
+)
+
+from services.comparison import ComparisonProxy
+from services.notification.notifiers.base import (
+    AbstractBaseNotifier,
+    NotificationResult,
+)
+
+log = logging.getLogger(__name__)
+
+
+class NotifyCondition(ABC):
+    is_async_condition: bool = False
+    failure_explanation: str
+
+    @abstractmethod
+    def check_condition(
+        notifier: AbstractBaseNotifier, comparison: ComparisonProxy
+    ) -> bool:
+        pass
+
+    def on_failure_side_effect(
+        notifier: AbstractBaseNotifier, comparison: ComparisonProxy
+    ) -> NotificationResult:
+        return NotificationResult()
+
+
+class AsyncNotifyCondition(NotifyCondition):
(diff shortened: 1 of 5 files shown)


nora-codecov (approved):
I like it!
Only request would be to add docstrings to the base `NotifyCondition` and `NotificationResult` to explain what those are and how we use them, so that it's faster to comprehend.

services/notification/notifiers/comment/conditions.py
```diff
@@ -0,0 +1,150 @@
...
+    @abstractmethod
+    async def check_condition(
+        notifier: AbstractBaseNotifier, comparison: ComparisonProxy
+    ) -> bool:
+        pass
+
+    async def on_failure_side_effect(
+        notifier: AbstractBaseNotifier, comparison: ComparisonProxy
+    ) -> NotificationResult:
+        pass
```
matt-codecov, line 54: should this return a default like the non-async one does?

matt-codecov (approved):
this is way more readable, great idea

if you want to go the extra mile, could move/adapt the tests for each condition from where they are in test_comment.py to a new test_conditions.py that tests the conditions directly. won't block on that request though

services/notification/notifiers/comment/conditions.py
```diff
@@ -0,0 +1,150 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class NotifyCondition(ABC):
+    """Abstract class that defines the basis of a NotifyCondition.
+
+    NotifyCondition are conditions that need to be met in order for a notification to be sent
+    from Codecov to a git provider.
+    NotifyConditions can have a side effect that is called when the condition fails.
```
michelletran-codecov, line 22: nit:

```suggestion
    NotifyCondition specifies the conditions that need to be met in order for a notification to be sent
    from Codecov to a Git provider.
    NotifyCondition can have a side effect that is called when the condition fails.
```

michelletran-codecov (approved):
I really like the change! It's much more readable now. 🌟 Thanks for the refactor!

PR #458, opened 2024-05-21
Increase the number of decimal places for the AbstractLevelTotals

michelletran-codecov:
After the migration to increase the number of decimal places for these tables, we should be able to use them to 5 decimal places. We just need to also update the models used in worker for consistency.

Migration happens in `shared`: https://github.com/codecov/shared/pull/224

ref: https://github.com/codecov/engineering-team/issues/1614

Diff for the `shared` update: https://github.com/codecov/shared/compare/codecov:f6c2c3852530192ab0c6b9fd0c0a800c2cbdb16f%5E%5E%5E%5E%5E...codecov:ae6e12f3cf43188a6d6fa0156ccb78252d3405a8

### Legal Boilerplate

### DIFF
--- a/database/models/reports.py
+++ b/database/models/reports.py
@@ -190,7 +190,7 @@ class ReportDetails(CodecovBaseModel, MixinBaseClass):
 
 class AbstractTotals(MixinBaseClass):
     branches = Column(types.Integer)
-    coverage = Column(types.Numeric(precision=7, scale=2))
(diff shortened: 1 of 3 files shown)


adrian-codecov (approved):
Lgtm! Just in case, this should be merged after the migrations from this https://github.com/codecov/shared/pull/224/files are applied 👌

PR #463, opened 2024-05-22
Logic to remove checks from team plan customer

adrian-codecov:
Customers with team plan only get patch status/checks, regardless of their yaml settings. This PR is to add this logic as it was never done when first creating the team plan. Ticket here https://github.com/codecov/engineering-team/issues/1660.

I introduced 2 new variables that indicate whether we need a status and a check. We are reusing `_should_use_checks_notifier` to check for status I just added a small piece to make it work with this logic. We are adding `_should_use_status_notifier` which returns true unless it's a team plan without a `patch` status/check - we've always returned status otherwise. Let me know if there's any confusion with this approach

### Legal Boilerplate

### DIFF
--- a/services/notification/__init__.py
+++ b/services/notification/__init__.py
@@ -13,9 +13,10 @@ from typing import Iterator, List
 from celery.exceptions import CeleryError, SoftTimeLimitExceeded
 from shared.config import get_config
 from shared.helpers.yaml import default_if_true
+from shared.plan.constants import TEAM_PLAN_REPRESENTATIONS
 from shared.yaml import UserYaml
 
-from database.models.core import GITHUB_APP_INSTALLATION_DEFAULT_NAME
+from database.models.core import GITHUB_APP_INSTALLATION_DEFAULT_NAME, Owner
 from helpers.metrics import metrics
 from services.comparison import ComparisonProxy
 from services.decoration import Decoration
@@ -60,8 +61,11 @@ class NotificationService(object):
         if checks_yaml_field is False:
             return False
 
-        owner = self.repository.owner
+        owner: Owner = self.repository.owner
         if owner.service not in ["github", "github_enterprise"]:
+            return False
+
+        if owner.plan in TEAM_PLAN_REPRESENTATIONS:
             return False
 
         app_installation_filter = filter(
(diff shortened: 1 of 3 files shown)


joseph-sentry (approved):

services/notification/__init__.py
```diff
@@ -101,39 +116,52 @@ def get_notifiers_instances(self) -> Iterator[AbstractBaseNotifier]:
...
+            should_use_checks_notifier = self._should_use_checks_notifier(
+                status_type=key
+            )
+
+            # Introduced to gate team plan not having statuses nor checks based on conditionals.
+            if not should_use_status_notifiers and not should_use_checks_notifier:
+                continue
+
+            # We always send statuses, so there's currently no case of checks without status.
+            if should_use_status_notifiers:
```
joseph-sentry, line 131: for readability maybe the logic following here should be encapsulated in a function

services/notification/__init__.py
```diff
@@ -101,39 +116,52 @@ def get_notifiers_instances(self) -> Iterator[AbstractBaseNotifier]:
...
-                    status_notifier=status_notifier_class(
+            should_use_status_notifiers = self._should_use_status_notifier(
+                status_type=key
+            )
+            should_use_checks_notifier = self._should_use_checks_notifier(
+                status_type=key
+            )
+
+            # Introduced to gate team plan not having statuses nor checks based on conditionals.
+            if not should_use_status_notifiers and not should_use_checks_notifier:
```
joseph-sentry, line 127: i'm wondering if this is meant to skip the comment and slack notifiers, if not, or if they will be skipped anyways, then we can remove this if statement

adrian-codecov: We should have the comment, clarifying about the slack one, but how would removing this help?

To clarify the logic here, everything remains the same, the only change introduced by this ticket is "if you're in a team plan, and you have a "changes" or "project" status checks, we don't show them, we only show "patch" specific comments" (the ticket was originally "skip everything" but its now "skip some things". There was a fn `_should_use_checks_notifier` that has some logic in it already that decides whether you show checks or not - I added a small case inside to do the team plan changes. We also always sent statuses, which changes with this ticket, so I added a fn `_should_use_status_notifier` which just adds the team plan related logic to it, otherwise returns true. 

So I don't see a way to get rid of this if ^ here.

PR #465, opened 2024-05-23
Rework backfill task to create sub-tasks

adrian-codecov:
This just reworks the existing GH app backfill task to distribute the backfill to subtasks. Previous implementation seemed to run out of memory, hence making several subtasks.

### Legal Boilerplate

### DIFF
--- a/celery_config.py
+++ b/celery_config.py
@@ -107,7 +107,9 @@ daily_plan_manager_task_name = "app.cron.daily.PlanManagerTask"
 
 # Backfill GH Apps
 backfill_existing_gh_app_installations_name = "app.tasks.backfill_existing_gh_app_installations.BackfillExistingGHAppInstallationsTask"
+backfill_existing_individual_gh_app_installation_name = "app.tasks.backfill_existing_individual_gh_app_installation.BackfillExistingIndividualGHAppInstallationTask"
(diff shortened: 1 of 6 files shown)


celery_config.py
```diff
@@ -107,7 +107,9 @@ def init_celery_tracing(*args, **kwargs):
 
 # Backfill GH Apps
 backfill_existing_gh_app_installations_name = "app.tasks.backfill_existing_gh_app_installations.BackfillExistingGHAppInstallationsTask"
+backfill_existing_individual_gh_app_installation_name = "app.tasks.backfill_existing_individual_gh_app_installation.BackfillExistingIndividualGHAppInstallationTask"
```
giovanni-guidini, line 110: this is more for future reference, but very nit

The task name is (or should be) formatted as
`app.[tasks|cron].<TaskConfigGroup>.<task name>`

so all of the tasks in the backfill could have the same <TaskConfigGroup> of `backfill` or `backfil_github_apps`, for example. This makes it easier for us to know what tasks are related.

giovanni-guidini (approved):
LGTM

celery_config.py
adrian-codecov: Ty for this! That honestly makes a lot more sense 🙏

PR #468, opened 2024-05-24
1459 ba pr message tweak

adrian-codecov:
We're adjusting the BA PR comment to accommodate customers that don’t want a PR message when there are no changes, posting here for visibility. To summarize:
- We’re adding require_bundle_changes, a setting you can add to your codecov yaml
- If there are no changes
  - If require_bundle_changes = false we show “Bundle size has no change” - we default to this
  - If require_bundle_changes = true, we don’t show the comment at all
- When there are changes, we always display them, regardless of this setting ^

Moved around some logic to make it work but generally this is it

### Legal Boilerplate

### DIFF
--- a/services/bundle_analysis.py
+++ b/services/bundle_analysis.py
@@ -366,24 +366,38 @@ class Notifier:
 
         pullid = pull.database_pull.pullid
         bundle_comparison = ComparisonLoader(pull).get_comparison()
-        skip_comment_without_size_changes = bundle_comparison.total_size_delta == 0 and read_yaml_field(self.current_yaml, ("comment", "require_bundle_changes"), False)
+        skip_comment_without_size_changes = (
+            bundle_comparison.total_size_delta == 0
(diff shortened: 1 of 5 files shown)


services/bundle_analysis.py
```diff
@@ -364,36 +365,59 @@ async def notify(self) -> bool:
...
         bundle_changes = comparison.bundle_changes()
 
-        bundle_rows, total_size_delta = self._create_bundle_rows(
+        bundle_rows = self._create_bundle_rows(
             bundle_changes=bundle_changes, comparison=comparison
         )
         return self._write_lines(
-            bundle_rows=bundle_rows, total_size_delta=total_size_delta, pull=pull
+            bundle_rows=bundle_rows,
+            total_size_delta=comparison.total_size_delta,
```
adrian-codecov, line 419: total_size_delta is now a property

giovanni-guidini (approved):

PR #469, opened 2024-05-27
feat: Add functions to get/set apps to commits

giovanni-guidini:
Adds `set_github_app_for_commit` and `get_github_app_for_commit` functions. These let us set and later get a GithubAppInstallation.id that's related to a Commit.id

ticket: https://github.com/codecov/engineering-team/issues/1737

👀 This commit is part 1/4 of a bigger change that is actually done, but I decided to break it up into multiple PRs so it's easier to review. See how it plays with the rest here: https://github.com/codecov/worker/compare/main...gio/pin-commit-to-ghapp

(I'll try to leave the original branch up to date best as I can)

### DIFF
--- a/services/github.py
+++ b/services/github.py
@@ -1,11 +1,14 @@
 import logging
 from typing import Optional
 
+from redis import RedisError
 from shared.github import InvalidInstallationError
 from shared.github import get_github_integration_token as _get_github_integration_token
 
+from database.models.core import Commit
 from helpers.cache import cache
 from helpers.exceptions import RepositoryWithoutValidBotError
+from services.redis import get_redis_connection
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


services/github.py
```diff
@@ -24,3 +27,35 @@ def get_github_integration_token(
...
+def set_github_app_for_commit(
+    installation_id: str | int | None, commit: Commit
+) -> bool:
+    if installation_id is None:
+        return False
+    redis = get_redis_connection()
+    try:
+        redis.set(
+            COMMIT_GHAPP_KEY_NAME(commit.id), str(installation_id), ex=(60 * 60 * 2)
+        )  # 2h
```
michelletran-codecov, line 44: For readability, let's extract `(60 * 60 * 2)` as a constant as well.

services/github.py
```diff
@@ -24,3 +27,35 @@ def get_github_integration_token(
     except InvalidInstallationError:
         log.warning("Failed to get installation token")
         raise RepositoryWithoutValidBotError()
+
+
+COMMIT_GHAPP_KEY_NAME = lambda commit_id: f"app_to_use_for_commit_{commit_id}"
+
+
+def set_github_app_for_commit(
+    installation_id: str | int | None, commit: Commit
```
michelletran-codecov, line 36: Hmm, `installation_id` has a lot of different  types. In what circumstances should we be expecting a `str` vs` int` vs `None` type? Should this be documented?

michelletran-codecov (commented):
Generally LGTM except for a few comments!

PR #476, opened 2024-05-29
feat: create django version of the repository and bots services

joseph-sentry:
This PR creates copies of the repository and bots services that use the Django models instead of the SQLAlchemy models. This is so that the test results finisher can use the django models instead of the SQLAlchemy models.

Note: there is a difference in the behaviour we're verifying between the regular repository service tests and the django repository service tests. One example is `test_fetch_and_update_pull_request_information_no_compared_to`

### DIFF
--- a/helpers/token_refresh_django.py
+++ b/helpers/token_refresh_django.py
@@ -1,0 +1,42 @@
+import logging
+from typing import Callable, Dict
+
+from asgiref.sync import sync_to_async
+from shared.django_apps.codecov_auth.models import Owner
+from shared.encryption.token import encode_token
+
+from services.encryption import encryptor
+
+log = logging.getLogger(__name__)
+
+
+def get_token_refresh_callback(owner: Owner) -> Callable[[Dict], None]:
+    """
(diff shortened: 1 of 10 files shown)


giovanni-guidini:
Can we instead change the bots services to be database agnostic?

It doesn't seem great to have 2 different copies of such important services - that would have to be maintained both, and is being actively worked on - in the same project. Especially without a clear deprecation date for the SQLAlchemy models.

I believe there's a way to make this service database agnostic so you can pass either model and have it still work (with helper functions for the models methods and fields, mostly) that would be better for us long-term.

giovanni-guidini (changes requested):
Can we instead change the bots services to be database agnostic?

It doesn't seem great to have 2 different copies of such important services - that would have to be maintained both, and is being actively worked on - in the same project. Especially without a clear deprecation date for the SQLAlchemy models.

I believe there's a way to make this service database agnostic so you can pass either model and have it still work (with helper functions for the models methods and fields, mostly) that would be better for us long-term.

PR #479, opened 2024-05-31
feat: get repo service for specific commit

giovanni-guidini:
Change the repository service so we have a function that
checks whether a commit has been pinned to a specific app
and returns the TorngitAdapter with said app.

Note on `services.repository._is_repo_using_integration`: this function has become effectively obsolete since the introduction of `get_adapter_auth_information` because we can tell if we are using the installation by the presence of "installation_info" or not.

👀 This is PR 3/4 in the pinning commits to apps endeavour

### Legal Boilerplate

### DIFF
--- a/services/repository.py
+++ b/services/repository.py
@@ -6,6 +6,7 @@ from typing import Any, Mapping, Optional, Tuple
 
 import shared.torngit as torngit
 from shared.config import get_config, get_verify_ssl
+from shared.django_apps.codecov_auth.models import Service
 from shared.torngit.exceptions import (
     TorngitClientError,
     TorngitError,
@@ -27,6 +28,11 @@ from helpers.token_refresh import get_token_refresh_callback
 from services.bots import (
     get_adapter_auth_information,
 )
+from services.bots.github_apps import (
+    get_github_app_token,
+    get_specific_github_app_details,
+)
+from services.github import get_github_app_for_commit
 from services.yaml import read_yaml_field
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


services/repository.py
```diff
@@ -65,7 +97,9 @@ def get_repo_provider_service(
     data = TorngitInstanceData(
         repo=RepoInfo(
             name=repository.name,
-            using_integration=_is_repo_using_integration(repository),
+            using_integration=(
+                adapter_auth_info["selected_installation_info"] is not None
```
michelletran-codecov, line 101: This will raise a `KeyError` if `selected_installation_info` is not found. If we don't want to error, then `adapter_auth_info.get("selected_installation_info")` is safer.

services/repository.py
```diff
@@ -27,36 +28,67 @@
...
+
+    This is done specifically after emitting checks for a PR using GitHub apps, because only the app
+    that posted the check can edit it later on. The "app for a commit" info is saved in Redis by the NotifyTask.
+    """
+    repository = commit.repository
+    installation_for_commit = get_github_app_for_commit(commit)
+    if installation_for_commit is None:
+        return get_repo_provider_service(repository, fallback_installation_name)
+
+    ghapp_details = get_specific_github_app_details(
```
adrian-codecov, line 58: Looking at the logic inside here, https://github.com/codecov/worker/blob/b8dcb6f4cb010c5c370ef2e9f40dd74e8e009183/services/bots/github_apps.py#L163, can you still use your app ifit is not configured?

michelletran-codecov, line 58: We do a check in `get_repo_provider_service` to ensure that the repo is a GitHub repo. Should we do a check here as well? This will save us a database read to the GitHub App Installation table I think. 

In fact, many parts of this function is GitHub specific... can we skip some more steps below if it's not a GitHub commit?

adrian-codecov (approved):

michelletran-codecov (commented):
Generally LGTM. I'm a bit confused/concerned about all the GitHub specific stuff in the `repository`, but it appears that other functions in that file is referring to GitHub specific stuff, so I'll let it go. Just a few comments/questions.

services/repository.py
giovanni-guidini: This check is in [get_github_app_for_commit](https://github.com/codecov/worker/blob/b8dcb6f4cb010c5c370ef2e9f40dd74e8e009183/services/github.py#L67).

From the docstring: "If the commit doesn't have a particular app assigned to it, return regular `get_repo_provider_service` choice".
Because apps only exist for GitHub it follows that only if it's a GitHub repo the code would get to this point.

PR #486, opened 2024-06-03
Bundle Analysis: associate past assets to current parsed bundle

JerrySentry:
Followup to https://github.com/codecov/shared/pull/231

Retrieve the previous bundle analysis report during processing of current bundle analysis report, then send it to `BundleAnalysisReport.associate_previous_assets` to process asset association.

closes: https://github.com/codecov/engineering-team/issues/1768

### Legal Boilerplate


services/bundle_analysis.py
```diff
@@ -123,6 +123,41 @@ async def initialize_and_save_report(
             db_session.flush()
         return commit_report
```
JerrySentry, line 125: This file is the only real change to this PR, everything else is ruff generated fixes.

giovanni-guidini (approved):
LGTM

PR #487, opened 2024-06-03
 fix: get the logs working again

joseph-sentry:


helpers/logging_config.py
```diff
@@ -1,19 +1,41 @@
...
 
-from celery._state import get_current_task
 from pythonjsonlogger.jsonlogger import JsonFormatter
 
 from helpers.environment import Environment, get_current_env
 
+task_name = None
+task_id = None
+
+# this is a temporary hack to get the task information working in the logs again
```
giovanni-guidini, line 11: Unless there's a GitHub ticket to fix this for good there are good chances that this "temporary hack" will become permanent

giovanni-guidini (commented):
This hack assumes that only 1 task runs at any given time in a single process. I think it's a fair assumption, but have you verified it? Otherwise we can have logs with wrong task_ids

joseph-sentry:
> This hack assumes that only 1 task runs at any given time in a single process. I think it's a fair assumption, but have you verified it? Otherwise we can have logs with wrong task_ids

I agree if there are multiple tasks running at once in a given process. this will be an issue. I added a log line to check if there is concurrent access and i'm going to run this on staging for a while to verify this.

joseph-sentry:
https://l.codecov.dev/deei0v

logs here look good to me

giovanni-guidini (approved):

PR #493, opened 2024-06-06
Bundle analysis add to sizes to timeseries

JerrySentry:
Insert bundle analysis related data points to the timeseries measurements table. Inserts six rows for each bundle,
- javascript_type -> the sum of all javascript typed assets for this bundle
- image_type -> the sum of all javascript typed assets for this bundle
- font_type -> the sum of all javascript typed assets for this bundle
- stylesheet_type -> the sum of all javascript typed assets for this bundle
- asset_type -> specific js assets using its asset UUID as the identifier (ie measurable_id) link with past commits
- bundle_type -> the sum of all assets for this bundle

This change adds a new task (bundle_analysis_save_measurements) that is called during the last step of the bundle_analysis_processor task. The new task will retrieve the bundle analysis report and insert the appropriate measurements to timeseries.

Creating timeseries dataset entries will be done in the API during handling of uploading the bundle file.

closes https://github.com/codecov/engineering-team/issues/1769
closes https://github.com/codecov/engineering-team/issues/1770

### Legal Boilerplate


services/bundle_analysis.py
```diff
@@ -253,6 +265,129 @@ def process_upload(self, commit: Commit, upload: Upload) -> ProcessingResult:
...
+                        commit,
+                        MeasurementName.bundle_analysis_report_size.value,
+                        bundle_report.name,
+                        bundle_report.total_size(),
+                    )
+
+                # For individual javascript associated assets using UUID
+                if MeasurementName.bundle_analysis_asset_size.value in dataset_names:
+                    for asset in bundle_report.asset_reports():
+                        if asset.asset_type == AssetType.JAVASCRIPT:
```
adrian-codecov, line 336: is this going to be `== to AssetType.JAVASCRIPT` or `== to AssetType.JAVASCRIPT.value`? Same for the other enums

tasks/bundle_analysis_save_measurements.py
```diff
@@ -0,0 +1,81 @@
...
+    def run_impl(
+        self,
+        db_session: Session,
+        repoid: int,
+        commitid: str,
+        uploadid: int,
+        commit_yaml: dict,
+        previous_result: Any,
+    ):
+        repoid = int(repoid)
```
adrian-codecov, line 31: Perhaps redundant if you're expecting an int?

adrian-codecov (approved):
Overall looks good to me, small comment left cause I'm not sure what enum value these end up with based on the tests (maybe worth adding?)

Did you also test this locally to ensure this works?

services/bundle_analysis.py
JerrySentry: no `.value` here because the `asset_type` is a `SQLAlchemyEnum` type which does the comparison without getting the actual values

JerrySentry:
> Did you also test this locally to ensure this works?

Yeah, the local timescale DB I showed you with the data was inserted with this PR.

PR #494, opened 2024-06-06
chore: re-work public tokens (part I)

giovanni-guidini:
ticket: https://l.codecov.dev/QLcMzU

1st part of the move to having more dedicated-tokens for certain functionality in the app.
It also now allows dedicated-tokens to come from apps (for GitHub)

### Legal Boilerplate

### DIFF
--- a/services/bots/helpers.py
+++ b/services/bots/helpers.py
@@ -22,7 +22,7 @@ def get_dedicated_app_token_from_config(
             service,
             app_id=app_id,
             installation_id=installation_id,
-            pem_path=f"yaml+file://{service}.dedicated_apps.{token_type.value}",
+            pem_path=f"yaml+file://{service}.dedicated_apps.{token_type.value}.pem",
         )
         return Token(key=actual_token, username=f"{token_type.value}_dedicated_app")
     return None
--- a/services/bots/helpers.py
+++ b/services/bots/helpers.py
@@ -1,3 +1,5 @@
+from typing import Any, Dict
+
 from shared.config import get_config
 from shared.torngit.base import TokenType
 from shared.typings.oauth_token_types import Token
@@ -9,7 +11,9 @@ def get_dedicated_app_token_from_config(
     service: str, token_type: TokenType
 ) -> Token | None:
     # GitHub can have 'dedicated_apps', and those are preferred
-    dedicated_app = get_config(service, "dedicated_apps", token_type.value, default={})
+    dedicated_app: Dict[str, Any] = get_config(
+        service, "dedicated_apps", token_type.value, default={}
+    )
     app_id = dedicated_app.get("id")
     installation_id = dedicated_app.get("installation_id")
     pem_exists = "pem" in dedicated_app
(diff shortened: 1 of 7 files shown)


services/bots/helpers.py
```diff
@@ -0,0 +1,32 @@
+from shared.config import get_config
```
michelletran-codecov, line 1: It took me a while to realize that `get_config` was actually loading from YAML configurations. `from shared.config import get_config` just looked a lot like it's loading application configs, not YAML (I had to dig into the code to verify that it's actually YAML). Anyways, not sure if it's a _me_ thing, or if other people will also be confused.

services/bots/helpers.py
```diff
@@ -0,0 +1,32 @@
...
+    pem_exists = "pem" in dedicated_app
+    if app_id and installation_id and pem_exists:
+        actual_token = get_github_integration_token(
+            service,
+            app_id=app_id,
+            installation_id=installation_id,
+            pem_path=f"yaml+file://{service}.dedicated_apps.{token_type.value}",
+        )
+        return Token(key=actual_token, username=f"{token_type.value}_dedicated_app")
+
```
michelletran-codecov, line 24: I know it's implicit (by not having a return statement) that we return None if we don't go through the if statement, but I think it'll be more readable if we have an explicit return None here.

services/bots/tests/test_helpers.py
```diff
@@ -0,0 +1,116 @@
...
+                        "pem": "another_path",
+                    },
+                    "tokenless": {"id": 1111},  # not configured (missing pem)
+                },
+            }
+        }
+    )
+    mock_get_integration_token.return_value = "installation_access_token"
+    # TokenType.read has a dedicated app AND bot.
+    # Dedicated app is preferred.
```
michelletran-codecov, line 83: This comment will be even more useful as a docstring on the `get_token_type_from_config` function. 

Also, for purposes of test clarity and documentation maintenance, how do you feel about turning each of these `asserts` into their own parametrize tests (with id annotation). If we fail this test here, we will only be notified for the first test that failed (and you'll have to rerun to get other failures/passes). It'll be more convenient to have the test status for all these checks run regardless of which one failed, especially since each of these asserts seem to be testing different usescases.

services/bots/tests/test_helpers.py
```diff
@@ -0,0 +1,116 @@
...
+)
+
+
+@patch("services.bots.helpers.get_github_integration_token")
+def test_get_dedicated_app_token_from_config(
+    mock_get_integration_token, mock_configuration
+):
+    mock_configuration.set_params(
+        {
+            "github": {
```
michelletran-codecov, line 18: This looks almost duplicated from the params set in the next test. Can we turn this dict into a fixture? ([docs for fixtures](https://docs.pytest.org/en/7.1.x/explanation/fixtures.html#about-fixtures))

michelletran-codecov (commented):
Generally LGTM. Just a few comments.

services/bots/public_bots.py
```diff
@@ -58,11 +59,20 @@ def get_token_type_mapping(
             "No admin_bog_token provided, but still continuing operations in case it is not doing an admin call anyway",
             extra=dict(repoid=repo.repoid),
         )
-    return {
-        TokenType.read: admin_bot_token or get_config(repo.service, "bots", "read"),
+
+    mapping = {
         TokenType.admin: admin_bot_token,
+        # [GitHub] Only legacy PATs can post statuses and comment to all public repos, so there can't be a dedicated_app for this
```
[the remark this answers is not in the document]
thomasrockhu-codecov: @michelletran-codecov I think [personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens)

services/bots/helpers.py
giovanni-guidini: > from shared.config import get_config just looked a lot like it's loading application configs

That is the case. The fact that the config lives in a YAML is just an implementation detail. For all intents and purposes it's fine to look at `get_config` as "getting application configs" and understand that the config logically is a map

michelletran-codecov (approved):
LGTM!

PR #496, opened 2024-06-07
dev: Add Most F Rules to Worker

ajay-sentry:
Part of the lint enhancement epic: https://github.com/codecov/engineering-team/issues/1716

This is the "First" Ruff PR for worker, creating the ruff.toml file allowing us to easily configure our ruff rules moving forward, similar to API and Shared. The goal for this PR was to keep the existing "I" and "F401" rules which were already running as well as add a majority of the remaining F rules, and fix as well.

The main rules that were fixed were related to duplicate test names and imports:
- https://docs.astral.sh/ruff/rules/redefined-while-unused/

Closes https://github.com/codecov/engineering-team/issues/1882

### Legal Boilerplate


services/notification/notifiers/mixins/message/helpers.py
```diff
@@ -293,7 +293,10 @@ def _row(title, c1, c2, plus="+", minus="-", neutral=" "):
     spacer = ["=" * row_w]
 
     title = "@@%s@@" % "{text:{fill}{align}{width}}".format(
-        text="Coverage Diff", fill=" ", align="^", width=row_w - 4, strip=True
```
ajay-sentry, line 296: Strip is an unused Param here

services/tests/test_flake_detector.py
```diff
@@ -161,32 +161,6 @@ def test_flake_consecutive_differing_outcomes_no_main_branch_specified(dbsession
     assert flaky_tests[test_id] == {FlakeSymptomType.CONSECUTIVE_DIFF_OUTCOMES}
 
 
-def test_flake_consecutive_differing_outcomes_no_main_branch_specified(dbsession):
```
ajay-sentry, line 164: @joseph-sentry This test looked to be a copy of the one above it, let me know if we actually wanted to keep

tasks/tests/unit/test_backfill_commit_data_to_storage_task.py
```diff
@@ -68,7 +68,7 @@ def test_all_report_rows(self, mock_handle_single_row, dbsession):
         def mock_handle_single_row_return_side_effect(db_session, commit, report_row):
             if report_row.code is None:
                 return {"success": True, "errors": []}
-            if report_row.code is "local":
```
ajay-sentry, line 71: converted "is" to == which is the proper way to do this comparison

It's in a script though so idk if it really changes anything

services/notification/notifiers/mixins/message/helpers.py
ajay-sentry: This was already fixed after rebasing 🤷

tasks/tests/unit/test_backfill_commit_data_to_storage_task.py
ajay-sentry: This was already fixed after rebasing 🤷

services/tests/test_flake_detector.py
adrian-codecov: if it's an exact copy, it's probably a slip up and can be deleted, but if they differ a bit, then it's probably a name mixup and we should rename and keep

adrian-codecov (approved):

services/tests/test_flake_detector.py
ajay-sentry: Yeah it was an exact copy :o

PR #497, opened 2024-06-08
attempt to fix cff duplicates

adrian-codecov:
Change isn't too bad, length is mostly from tests, but it's important to understand the problem. More described here as well: https://github.com/codecov/internal-issues/issues/478. These are the changes in summary but the proper explanation is below:
- Added _possibly_delete_existing_cffs and _determine_cffs_and_depths_in_db fns. These find if there are suitable carryforwards sessions that should be deleted
- Exposed the parent_depth in `session_extras` for uploads
- Added a script to delete uploads + associated DB records
- Added a couple of tests for the above ^

---

We've been seeing some very inconsistent and strange behaviour where customers using carryforward flags (cffs) see a disproportionate growth of cffs over time. It doesn't happen every commit, nor there is a specific pattern to it, but it is confirmed. A little on carryforward uploads.

What are cffs? It's a feature we offer where customers can persist upload coverage across commits without having to explicitly upload on every commit - customers specify this in their codecov yaml. Behind the scenes, we look for a commit's **parent commit** and "copy paste" the relevant upload information into the current commit. 

The problem we're seeing is certain commits carry forward uploads from **two different parent commits**; this should never happen. (you'll see me here say upload and session interchangeably). How does this happen? See this example:

### Example 
Imagine you have a chain of commits, A -> B -> C, where C is the current commit, B is C's parent, and A is B's parent/C's grand-parent. In normal circumstances, if a customer has cffs on for say "unit" flag and they don't provide an upload with the "unit" flag in B and C, B would carry forward the upload from A, and C would carry forward the upload from B. 

While we have a `parent_commit` field for the `Commit` model, we have a [dedicated function](https://github.com/codecov/worker/blob/08d078553953aa8bb5f39b925692110a36885d7d/services/report/__init__.py#L648-L649) to know which parent a commit should carry forward to. This function will find the closest suitable parent commit (normally it's immediate parent) and carryforward reports from it. This function is needed because sometimes a commit's direct parent might not have a valid report or one at all, the commit might be in an error state, there might not be a direct parent at all, etc. I won't go too much in detail in the different states but you can read the function if you're curious, but I want to focus on the scenario where a parent_commit is processing, which is when it is in "pending" state. A parent commit that is processing isn't a suitable candidate to carryforward a report from, because for all we know it doesn't have a valid report or it can error or other causes, so we advance to its parent and use that as a candidate. In the example above, C would look at B, which is processing, so C would carryforward from A.  

This is so far expected and normal. The funky part occurs if you upload two or more times for commit C, one upload when B is processing, and one after B is successfully completed. In the the eyes of C, it would carryforward from A the first time and from B the second time, and if you consider A to have "N" uploads to carryforward and B to have "M" uploads to carry forward, then C would end up with "M + N" carryforward uploads - this is wrong, and C should only have "N" uploads. To take the example even further, if D eventually happens after C, and it carries the uploads from C, it would end up with "M + N" uploads, hence you see this "duplicate" effect every now and then.  -- end of example

Now, if you're thinking, why doesn't this happen all the time as many people upload more than once per commit, you're thinking in the right vein. That's because, for this weird bug to happen, you need to upload multiple times `very fast`, and those uploads need to be `very big`. A commit doesn't stay processing for very long, so the window two different uploads to occur when a commit is processing and when it's not is slim. It is not impossible, but it's very unlikely. On top of that, two different uploads won't run this logic because it needs to go through [this](https://github.com/codecov/worker/blob/08d078553953aa8bb5f39b925692110a36885d7d/services/report/__init__.py#L293-L294) first, which is true when there isn't a report object and false after that is populated (so technically, 2 different uploads should never get here cause the second run will evaluate false and keep going). And thirdly, upload tasks have a lock for a task + repo + commit combo, so two uploads for the same commit couldn't be racing for the same function.

So, how does this actually happen? It's because there are **two** tasks that could run into this functionality: the `upload` and `preprocessing` tasks. These two can theoretically call the logic that chooses the commit's parent for cffs, one that chooses A as the "parent" and the other one that chooses "B" as the parent. 

So if we piece everything together: if you run the preprocess + upload tasks `in quick succession` for a commit with `a lot` of uploads, there is a chance that they choose `different parents`. 

So what am I doing? (finally), I'm writing some logic at "upload insertion time" to determine if there are already cffs from an older parent, and delete those if they exist. In the example above parting from C, I would be deleting the cffs from A when I see cffs coming from B.

Hopefully you enjoyed the read! 

### Legal Boilerplate

### DIFF
--- a/services/report/__init__.py
+++ b/services/report/__init__.py
@@ -4,10 +4,11 @@ import logging
 import sys
 import typing
 import uuid
+from collections import defaultdict
 from dataclasses import dataclass
 from json import loads
 from time import time
-from typing import Any, Dict, Mapping, Optional, Sequence
+from typing import Any, Dict, List, Mapping, Optional, Sequence
 
 import sentry_sdk
 from celery.exceptions import SoftTimeLimitExceeded
@@ -49,6 +50,7 @@ from rollouts import (
     PARALLEL_UPLOAD_PROCESSING_BY_REPO,
 )
 from services.archive import ArchiveService
+from services.delete.delete import delete_upload_by_ids
 from services.redis import (
     PARALLEL_UPLOAD_PROCESSING_SESSION_COUNTER_TTL,
     get_parallel_upload_processing_session_counter_redis_key,
@@ -59,6 +61,8 @@ from services.report.parser.types import ParsedRawReport
 from services.report.raw_upload_processor import process_raw_upload
 from services.repository import get_repo_provider_service
 from services.yaml.reader import get_paths_from_flags, read_yaml_field
+
+MIN_PARENT_COMMIT_DEPTH = 1
 
 
 @dataclass
@@ -646,7 +650,8 @@ class ReportService(BaseReportService):
     ) -> Optional[Commit]:
         parent_commit = commit.get_parent_commit()
         parent_commit_tracking = []
-        count = 1  # `parent_commit` is already the first parent
+        count = MIN_PARENT_COMMIT_DEPTH  # `parent_commit` is already the first parent
+
         while (
             parent_commit is not None
             and parent_commit.state not in ("complete", "skipped")
@@ -688,7 +693,7 @@ class ReportService(BaseReportService):
                     parent_tracing=parent_commit_tracking,
                 ),
             )
-            return None
+            return (None, None)
         if parent_commit.state not in ("complete", "skipped"):
             log.warning(
                 "None of the parent commits were in a complete state to be used as CFing base",
@@ -700,8 +705,8 @@ class ReportService(BaseReportService):
                     would_be_parent=parent_commit.commitid,
                 ),
             )
-            return None
-        return parent_commit
+            return (None, None)
+        return (parent_commit, count)
 
     async def _possibly_shift_carryforward_report(
         self, carryforward_report: Report, base_commit: Commit, head_commit: Commit
@@ -778,8 +783,10 @@ class ReportService(BaseReportService):
                 )
             )
 
-            parent_commit = self.get_appropriate_commit_to_carryforward_from(
-                commit, max_parenthood_deepness
+            parent_commit, parent_depth = (
+                self.get_appropriate_commit_to_carryforward_from(
+                    commit, max_parenthood_deepness
+                )
             )
             if parent_commit is None:
                 log.warning(
@@ -807,6 +814,14 @@ class ReportService(BaseReportService):
                 if self.current_yaml.flag_has_carryfoward(flag_name):
                     flags_to_carryforward.append(flag_name)
             if not flags_to_carryforward:
+                log.warning(
+                    "There weren't flags to carry forward in the parent commit",
+                    extra=dict(
+                        commit=commit.commitid,
+                        repoid=commit.repoid,
+                        parent_commit=parent_commit.commitid,
+                    ),
+                )
                 return Report()
             paths_to_carryforward = get_paths_from_flags(
                 self.current_yaml, flags_to_carryforward
@@ -826,7 +841,10 @@ class ReportService(BaseReportService):
                 parent_report,
                 flags_to_carryforward,
                 paths_to_carryforward,
-                session_extras=dict(carriedforward_from=parent_commit.commitid),
+                session_extras=dict(
+                    carriedforward_from=parent_commit.commitid,
+                    parent_depth=parent_depth,
+                ),
             )
             # If the parent report has labels we also need to carryforward the label index
             # Considerations:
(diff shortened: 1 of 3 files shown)


giovanni-guidini (commented):
Thanks for the detailed explanation in the PR. Very helpful.

After reading through that and the proposed solution it seems that you are trying to fix the _effects_ of the issue, not the _cause_ (the cause being that 2 different tasks might be doing the same thing at the same time).

Wouldn't it be preferable to fix the cause of the problem?
It saves us from having to do extra work (saving sessions and then deleting them).

giovanni-guidini:
@adrian-codecov the main objective of the `create-report` command is to create a report (as the name implies). That process happens [in the api](https://github.com/codecov/codecov-api/blob/bccac2466501259424b8a16044495d28532ddd59/upload/views/reports.py#L40), not in the task.

The pre-process upload task simply initializes the report that was created (carryforward is part of the initialization), but as we've seen that is also done in the upload task. So for people using the `create-report` command there's no real change. The difference is that the initialization will happen in the Upload task, not PreProcessUpload one.

To which you might be wondering "what's the point of a redundant task?"
Because the report only needs to be initialized once the idea of PreProcessUpload task is to speed up report processing by doing the initialization. As such when the upload comes the system is primed to process it.

Anyway, if we remove it the upload task would just take care of the process initialization. It might increase the time to notify by a little bit, but that would be the extend of the impact, I think.

Using the same lock would indeed solve it, but having more tasks fighting for the same lock would increase coordination overhead (and we have a clear way to have less tasks fighting for locks).

I don't see how increasing the timeout would help with the issue

giovanni-guidini:
About this bit
> While I think this is 99% the cause, I'd like to see if it's indeed the case in prod so that's why I thought of first trying out this.

That makes sense, but the code as is seems quite "final" and "permanent". Assuming this solution works to fix the carryforward issues we see (which it seems that it would work, the change itself seems fine*), what would be the next steps to solve the cause?
Or would we just leave this fix there and potentially drown other sources of this issue (with more work by the system), instead of preventing the system from creating this error condition?

(Of course I might be missing a very obvious "this is urgent, we need to patch this asap and worry about not making it happening again later", which would be very valid)



* - The sessions that belong to a report are also saved in the `report_json` object that goes to GCS. I do believe that the report json would list sessions for just 1 of the parents - "as if" there were no duplicates in the report itself - but have you verified that?

services/report/__init__.py
```diff
@@ -1173,6 +1191,74 @@ def save_report(self, commit: Commit, report: Report, report_code=None):
...
+        carryforward_report: Report,
+    ):
+        cffs_and_depths_in_db = self._determine_cffs_and_depths_in_db(
+            db_session=db_session, commit=commit
+        )
+
+        if not cffs_and_depths_in_db:
+            return
+
+        # Gets first 'parent_depth' it finds, as all cffs should be carried forward
```
giovanni-guidini, line 1238: Shouldn't it get the minimum 'parent_depth' instead of the first?

services/report/__init__.py
```diff
@@ -60,6 +62,8 @@
 from services.repository import get_repo_provider_service
 from services.yaml.reader import get_paths_from_flags, read_yaml_field
 
+MIN_PARENT_COMMIT_DEPTH = 1
```
giovanni-guidini, line 65: Will this ever not be 1?

services/report/__init__.py
```diff
@@ -1173,6 +1191,74 @@ def save_report(self, commit: Commit, report: Report, report_code=None):
...
+        for parent_depth, upload_ids in cffs_and_depths_in_db.items():
+            if parent_depth > carryforward_report_depth:
+                log.info(
+                    "Deleting upload from DB",
+                    extra=dict(
+                        repoid=commit.repoid,
+                        commit=commit.commitid,
+                    ),
+                )
+                # Potentially make task here - this would make the preprocess task likely timeout
```
giovanni-guidini, line 1257: The upload task would be likely to timeout here too? Because that's a way bigger problem

adrian-codecov: It could depending on the amount of things to delete, hence the ask here to see if it would make sense to create a separate deletion task

services/report/__init__.py
adrian-codecov: This value itself will always be 1, but the parent depth a commit chooses is determined by this https://github.com/codecov/worker/blob/f12fe8f4701da9201ff5c64b55540b891d9cb8ea/services/report/__init__.py#L648-L649. I just made this into a constant cause I was referencing it in 2 different places

services/report/__init__.py
adrian-codecov: The next four lines do that, it goes through the Report object which contains the uploads that will be carryforwarded and will determine the depth they have. I'm initializing `carryforward_report_depth` with the arbitrary depth of 1 and in case line 1241 doesn't run, so the variable has a non-null value. I update it shortly after in line 1243 if a depth for the carryforward reports are found, does that make sense?

services/report/__init__.py
giovanni-guidini: I think that would be a good idea, yes.
Especially if it only changes the database and not the `Report` object it would make a lot of sense to move this to a different task

services/report/__init__.py
giovanni-guidini: well... no. The question is if the code should get the _smallest_ 'parent_depth' from all the sessions in the report. Both the comment and the code are getting the 1st parent_depth that appears. Those values can be different depending on the order of sessions that were carried forward, and you made no claims about the order of the sessions

asked differently: Is the 1st session that appears always the smallest one?

PR #501, opened 2024-06-12
feat: add backfill test instances task

joseph-sentry:
Depends on: https://github.com/codecov/shared/pull/247
Fixes: https://github.com/codecov/engineering-team/issues/1648 and https://github.com/codecov/engineering-team/issues/1647

Requires risky migration from https://github.com/codecov/shared/commit/05641f5e2ff8890e74c87c9b6d705145b5c9c943 to be run

### DIFF
--- a/tasks/backfill_test_instances.py
+++ b/tasks/backfill_test_instances.py
@@ -9,26 +9,34 @@ from tasks.base import BaseCodecovTask
 
 log = logging.getLogger(__name__)
 
+from django.db import transaction
+
 
 class BackfillTestInstancesTask(
     BaseCodecovTask, name=backfill_test_instances_task_name
 ):
-    def run_impl(self, *args, dry_run=True, **kwargs):
+    def run_impl(
+        self,
+        *args,
+        count=None,
+        **kwargs,
+    ):
         log.info(
             "Updating test instances",
         )
 
-        test_instance_list = (
-            TestInstance.objects.select_related("upload__report__commit")
-            .filter(
-                Q(branch__isnull=True)
-                | Q(commitid__isnull=True)
-                | Q(repoid__isnull=True)
-            )
-            .all()
-        )
+        test_instances_missing_info_count = TestInstance.objects.filter(
+            Q(branch__isnull=True) | Q(commitid__isnull=True) | Q(repoid__isnull=True)
+        ).count()
 
-        for i in range(0, test_instance_list.count(), 1000):
+        if count is None:
+            count = test_instances_missing_info_count
+        else:
+            count = min(count, test_instances_missing_info_count)
+
+        log.info("Updating {count} test instances")
+
+        for _ in range(0, test_instances_missing_info_count, 1000):
             updates = []
             test_instances_missing_info = (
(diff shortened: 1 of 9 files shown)


tasks/backfill_test_instances.py
```diff
@@ -17,27 +18,34 @@ def run_impl(self, *args, dry_run=True, **kwargs):
             "Updating test instances",
         )
 
-        test_instance_filter = TestInstance.objects.select_related(
-            "upload__report__commit"
-        ).filter(
-            branch=None,
-            commitid=None,
+        test_instance_list = (
+            TestInstance.objects.select_related("upload__report__commit")
```
adrian-codecov, line 22: Good to see we're using the django models! 🥇

tasks/backfill_test_instances.py
```diff
@@ -17,27 +18,34 @@ def run_impl(self, *args, dry_run=True, **kwargs):
...
-        chunks = [
-            all_test_instances[i : i + chunk_size]
-            for i in range(0, num_test_instances, chunk_size)
-        ]
-
-        for chunk in chunks:
-            for test_instance in chunk:
+        for i in range(0, test_instance_list.count(), 1000):
+            updates = []
+            thing = (
```
adrian-codecov, line 33: Could we rename this to something more meaningful 😅, seems to me this is something like "test_instances_without_identifiers" or something like that?

tasks/sync_test_results.py
```diff
@@ -0,0 +1,87 @@
...
+from django.db import connection
+from shared.celery_config import sync_test_results_task_name
+from shared.django_apps.reports.models import Test
+
+from app import celery_app
+from tasks.base import BaseCodecovTask
+
+log = logging.getLogger(__name__)
+
+# TODO: turn these into Django ORM calls
```
adrian-codecov, line 12: I think it should be this or close to this

```
from django.db.models import Avg, Case, When, Value
from django.db.models.functions import Now
from django.utils import timezone
from datetime import timedelta

... some code

# 30 days ago
thirty_days_ago = timezone.now() - timedelta(days=30)

failure_rate_queryset = TestInstance.objects.filter(
    repoid=<repoid>,
    created_at__gt=thirty_days_ago,
    outcome__in=['pass', 'failure', 'error']
).values('test_id').annotate(
    failure_rate=Avg(
        Case(
            When(outcome='pass', then=Value(0.0)),
            When(outcome__in=['failure', 'error'], then=Value(1.0)),
            output_field=FloatField()
        )
    )
)

commit_agg_queryset = TestInstance.objects.filter(
    repoid=<repoid>,
    created_at__gt=thirty_days_ago,
    outcome__in=['failure', 'error']
).values('test_id').annotate(
    commits=ArrayAgg('commitid', distinct=True)
)
```

adrian-codecov (commented):
Have we tested this in staging? Idk if we have test instances there though

tasks/backfill_test_instances.py
joseph-sentry: oops 😅, `test_instances_without_identifiers` sounds good

PR #506, opened 2024-06-19
small cleanup

giovanni-guidini:


tasks/notify.py
```diff
@@ -341,9 +342,6 @@ def run_impl_within_lock(
                 )
             else:
                 base_report = None
-            head_report = report_service.get_existing_report_for_commit(
```
giovanni-guidini, line 344: This happens on line 300 as well. I'm just removing the repeated operation.

michelletran-codecov (approved):

PR #516, opened 2024-06-24
Add message for users to install Codecov app if they're using public commenter

michelletran-codecov:
This is to drive more adoption for the Codecov App.

Issue link: https://github.com/codecov/engineering-team/issues/1833

### Legal Boilerplate


tasks/notify.py
```diff
@@ -400,6 +404,21 @@ def run_impl_within_lock(
...
+        """Returns a boolean indicating if the message will be sent by codecov-commenter.
+        If the user doesn't have an installation, and if the token type for the repo is codecov-commenter,
+        then it's likely that they're using the commenter bot.
+        """
+        commenter_bot_token = get_config(repository_service.service, "bots", "comment")
+        return (
+            repository_service.service == "github"
+            and repository_service.data.get("installation") is None
+            and repository_service.get_token_by_type(TokenType.comment)
+            == commenter_bot_token
```
michelletran-codecov, line 419: I'm not 100% sure about this part. Am I missing some checks to make sure that the user doesn't have any installations and is also using the commenter bot?

services/notification/notifiers/tests/unit/test_comment.py
```diff
@@ -800,6 +800,63 @@ async def test_create_message_files_section_with_critical_files(
...
+    async def test_create_message_with_github_app_comment(
+        self,
+        dbsession,
+        mock_configuration,
+        mock_repo_provider,
+        sample_comparison,
+        mocker,
+    ):
+        comparison = sample_comparison
+        mocker.patch.object(
```
giovanni-guidini, line 813: This might be unecessary in this test if you only care for the 1st line of the comment

tasks/notify.py
giovanni-guidini: This looks OK

Maybe just for peace of mind you should add 1 extra check `commenter_bot_token is not None` to make sure something is configured for the commenter bot...

giovanni-guidini (approved):

PR #521, opened 2024-06-25
add needs_webhook_secret_backfill to upload task

nora-codecov:
set up the webhook_secret in passing if it is empty - starts the fix for https://github.com/codecov/internal-issues/issues/366

### DIFF
--- a/tasks/upload.py
+++ b/tasks/upload.py
@@ -27,8 +27,12 @@ from app import celery_app
 from database.enums import CommitErrorTypes, ReportType
 from database.models import Commit, CommitReport
 from database.models.core import GITHUB_APP_INSTALLATION_DEFAULT_NAME
-from helpers.checkpoint_logger import _kwargs_key
-from helpers.checkpoint_logger import from_kwargs as checkpoints_from_kwargs
+from helpers.checkpoint_logger import (
+    _kwargs_key,
+)
+from helpers.checkpoint_logger import (
+    from_kwargs as checkpoints_from_kwargs,
+)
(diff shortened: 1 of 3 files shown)


tasks/upload.py
```diff
@@ -959,25 +974,51 @@ def possibly_setup_webhooks(self, commit, repository_service):
...
+                            action="EDIT",
+                        ),
+                    )
+
                 if webhook_secret is not None:
                     repository.webhook_secret = webhook_secret
-                repo_data["repo"]["hookid"] = hookid
-                return True
+
+                return True if should_post_webhook else False
```
adrian-codecov, line 1013: With the other suggestion above, I think it'd be a bit more readable to put return True in the `should_post_webhook` conditional, and false in the `else` in 993, wdyt?

adrian-codecov (approved):
Lefts some comments/suggestions but overall lgtm!

tasks/upload.py
```diff
@@ -959,25 +974,51 @@ def possibly_setup_webhooks(self, commit, repository_service):
...
+                        extra=dict(
+                            repository_service=repository_service.service,
+                            repoid=repository.repoid,
+                            commit=commit.commitid,
+                            hookid=repository.hookid,
+                            action="EDIT",
+                        ),
+                    )
+
                 if webhook_secret is not None:
```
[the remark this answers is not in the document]
nora-codecov: I like it! updating

tasks/upload.py
nora-codecov: agree! updating

PR #524, opened 2024-06-26
feat: pre risky migration flaky test changes

joseph-sentry:
This commit makes the changes possible before the risky migrations in the supporting shared PR are done.

- update shared version
- update test results parser version
- add time-machine as a dependency
- add Flake and ReducedError sqlalchemy models
- modify TestResultsNotificationPayload to contain a set of flaky test ids instead of a dict[str, TestResultsNotificationFlake]
- change flaky test results comment format
- change flake detection in test results finisher to gather the Flake objects for a given repo and compare their test ids to the test ids of the failures relevant to the test results comment

### DIFF
--- a/database/models/reports.py
+++ b/database/models/reports.py
@@ -322,3 +322,28 @@ class TestResultReportTotals(CodecovBaseModel, MixinBaseClass):
     skipped = Column(types.Integer)
     failed = Column(types.Integer)
     error = Column(types.String(100), nullable=True)
+
+
+class ReducedError(CodecovBaseModel, MixinBaseClass):
+    __tablename__ = "reports_reducederror"
+    message = Column(types.Text)
+
+
+class Flake(CodecovBaseModel, MixinBaseClass):
+    __tablename__ = "reports_flake"
+    repoid = Column(types.Integer, ForeignKey("repos.repoid"))
+    repository = relationship("Repository", backref=backref("flakes"))
+
+    testid = Column(types.Text, ForeignKey("reports_test.id"))
+    test = relationship(Test, backref=backref("flakes"))
+
+    reduced_error_id = Column(
+        types.Integer, ForeignKey("reports_reducederror.id"), nullable=True
+    )
+    reduced_error = relationship(ReducedError, backref=backref("flakes"))
+
+    recent_passes_count = Column(types.Integer)
+    count = Column(types.Integer)
+    fail_count = Column(types.Integer)
+    start_date = Column(types.DateTime)
+    end_date = Column(types.DateTime, nullable=True)
--- a/tasks/test_results_finisher.py
+++ b/tasks/test_results_finisher.py
@@ -9,23 +9,14 @@ from test_results_parser import Outcome
 
 from app import celery_app
 from database.enums import FlakeSymptomType, ReportType, TestResultsProcessingError
-from database.models import Commit, TestResultReportTotals
+from database.models import Commit, Flake, TestResultReportTotals
 from helpers.checkpoint_logger import from_kwargs as checkpoints_from_kwargs
 from helpers.checkpoint_logger.flows import TestResultsFlow
 from helpers.string import EscapeEnum, Replacement, StringEscaper, shorten_file_paths
 from rollouts import FLAKY_TEST_DETECTION
-from services.failure_normalizer import FailureNormalizer
-from services.flake_detection import (
-    DefaultBranchFailureDetector,
-    DiffOutcomeDetector,
-    FlakeDetectionEngine,
-    FlakeDetectionResult,
-    UnrelatedMatchesDetector,
-)
 from services.lock_manager import LockManager, LockRetry, LockType
 from services.test_results import (
     TestResultsNotificationFailure,
-    TestResultsNotificationFlake,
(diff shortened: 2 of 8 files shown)


database/models/reports.py
```diff
@@ -322,3 +322,28 @@ class TestResultReportTotals(CodecovBaseModel, MixinBaseClass):
...
     failed = Column(types.Integer)
     error = Column(types.String(100), nullable=True)
+
+
+class ReducedError(CodecovBaseModel, MixinBaseClass):
+    __tablename__ = "reports_reducederror"
+    message = Column(types.Text)
+
+
+class Flake(CodecovBaseModel, MixinBaseClass):
```
adrian-codecov, line 332: Is the idea to then replace these w/ their django counterparts after you run the risky migration?

joseph-sentry: no, these will remain, these changes just don't depend on the risky migrations to run

michelletran-codecov (commented):
Generally LGTM modulo comment below and adding SQLAlchemy models. The current query to get flakes _feels_ pretty isolated. I'm wondering if there's a way for us to use the Django models instead. If there are no writes in this task, then there will unlikely be transaction contention. Also, the current query is pretty isolated, but I'm guessing that we'll need to tie it with `TestInstance` at some point? If that's the only place where we're interacting with the existing SQLAlchemy models, then would it be doable with referencing the ids?

Of course, keeping two different database models and connections for this task is also adding unnecessary complexity, so I'm not opposed to also just adding SQLAlchemy models.

database/models/reports.py
adrian-codecov: Mmm is there a reason we don't want to use the django models instead?

joseph-sentry: it's possible the `Flake` model is not needed, but the `ReducedError` model is needed because we will be writing the reduced_error_id in the test results processor in future changes

joseph-sentry:
I'm having a lot of trouble getting the tests to work with both the Django ORM and sqlalchemy, i'm trying to get them to both connect to the same db but there's complications because django wants to create it's own test database and sqlalchemy tries creating one as well. I don't think the complexity is worth the benefits.

database/models/reports.py
adrian-codecov: After talking, we'll defer the use of django models as it poses some difficulties with testing atm

tasks/test_results_finisher.py
```diff
@@ -296,121 +283,43 @@ def process_impl_within_lock(
...
+            identifier=repoid, default=False
+        ) and read_yaml_field(
+            commit_yaml, ("test_analytics", "flake_detection"), False
+        ):
+            flaky_tests = set()
+            repo_flakes = (
+                db_session.query(Flake)
+                .filter(  # type:ignore
+                    Flake.repoid == repoid,
+                    Flake.end_date.is_(None),
```
michelletran-codecov, line 310: I see that the index is added on a compound of ["repository", "test", "reduced_error", and "date"]. This means that for this query, it's _only_ going to use the index for "repository" (because [Postgres processes index from left to right](https://www.postgresql.org/docs/current/indexes-multicolumn.html)). 

I _believe_ we can also add an index for null fields specifically to make it more efficient, but I don't have as much experience with this. :p If we want to make this query efficient, we might want to explore doing a compound index with (repository, enddate = null) or something.

michelletran-codecov (commented):
A few more comments about query performance.

tasks/test_results_finisher.py
joseph-sentry: would adding on a `Flake.testid.in_(list_of_test_ids)` make this more efficient then? I plan to to eventually add the reduced_error_id to the query as well, so in the end we will be making full use of this index.

tasks/test_results_finisher.py
```diff
@@ -296,121 +283,43 @@ def process_impl_within_lock(
...
+                )
+                .all()
```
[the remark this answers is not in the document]
joseph-sentry: I agree that we could limit and offset this query to reduce memory usage but maybe it'd be better to filter on test ids in the query like i mentioned in my comment below, and also only select the test id from the query which is all we're looking for here. Is there another reason other than memory usage that we would limit this query?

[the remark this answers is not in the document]
michelletran-codecov: >  only select the test id from the query which is all we're looking for here.

I'm good with providing a id list. We will probably also want to ensure that the list isn't too long (i.e. we can cap this on the application side). It's _probably_ fine to query all the tests for now. However, we will want to keep an eye on the performance of this query (correlated with number of failed tests it's trying to retrieve).

> Is there another reason other than memory usage that we would limit this query?

Yes. Having a bound on the number of items returned will also make the processing of the results more predictable. For example, we don't have to worry (as much) about long running tasks because it's going to only ever process i.e. 30 items.

tasks/test_results_finisher.py
michelletran-codecov: > I plan to to eventually add the reduced_error_id to the query as well, so in the end we will be making full use of this index.

Ah OK! This is fine then. 👍

michelletran-codecov (approved):
LGTM!

PR #525, opened 2024-06-26
feat: flake detection post risky migration changes 

joseph-sentry:
These changes should be committed after the risky migrations in the
flake detection shared change are run.

- Add reduced_error foreign key to test instance sqlalchemy models
- add the process flakes task which takes in a repoid and list of commit
  ids as input and upserts Flake objects in the DB based on the test
  instances on those commits. This task is what implements the
  flakiness heuristic.
- call the process flakes task in sync pulls on merged commits
- call the process flakes task in the test results finisher when we
  receive test results from the main branch

### DIFF
--- a/tasks/process_flakes.py
+++ b/tasks/process_flakes.py
@@ -13,6 +13,8 @@ log = logging.getLogger(__name__)
 
 FlakeDict = dict[Any, Flake]
 
+FLAKE_EXPIRY_COUNT = 30
+
 
 class ProcessFlakesTask(BaseCodecovTask, name=process_flakes_task_name):
     """
@@ -55,14 +57,16 @@ class ProcessFlakesTask(BaseCodecovTask, name=process_flakes_task_name):
         return {"successful": True}
 
 
-def get_test_instances(commit_id, repo_id, branch):
-    test_instances = TestInstance.objects.filter(
-        commitid=commit_id, repoid=repo_id, branch=branch
-    ).all()
+def get_test_instances(commit_id: str, repo_id: int, branch: str) -> list[TestInstance]:
+    test_instances = list(
+        TestInstance.objects.filter(
+            commitid=commit_id, repoid=repo_id, branch=branch
+        ).all()
+    )
     return test_instances
 
 
-def generate_flake_dict(repo_id) -> FlakeDict:
+def generate_flake_dict(repo_id: int) -> FlakeDict:
     flakes = Flake.objects.filter(repository_id=repo_id, end_date__isnull=True).all()
     flake_dict = dict()
     for flake in flakes:
@@ -70,15 +74,15 @@ def generate_flake_dict(repo_id) -> FlakeDict:
     return flake_dict
 
 
-def update_passed_flakes(flake: Flake):
+def update_passed_flakes(flake: Flake) -> None:
     flake.count += 1
     flake.recent_passes_count += 1
-    if flake.recent_passes_count == 30:
+    if flake.recent_passes_count == FLAKE_EXPIRY_COUNT:
         flake.end_date = dt.datetime.now(tz=dt.UTC)
     flake.save()
 
 
-def upsert_failed_flake(test_instance: TestInstance, repo_id, flake: Flake | None):
+def upsert_failed_flake(test_instance: TestInstance, repo_id: int, flake: Flake | None):
     if flake is None:
         flake = Flake(
             repository_id=repo_id,
--- a/tasks/process_flakes.py
+++ b/tasks/process_flakes.py
@@ -1,7 +1,7 @@
 import datetime as dt
 import logging
-from typing import Any
 
+from sentry_sdk import metrics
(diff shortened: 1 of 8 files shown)


tasks/process_flakes.py
```diff
@@ -0,0 +1,103 @@
...
+    flake_dict = dict()
+    for flake in flakes:
+        flake_dict[flake.test_id] = flake
+    return flake_dict
+
+
+def update_passed_flakes(flake: Flake):
+    flake.count += 1
+    flake.recent_passes_count += 1
+    if flake.recent_passes_count == 30:
```
giovanni-guidini, line 76: Is this number (30) arbitrary? I might be a good candidate for a constant value with a name

tasks/process_flakes.py
```diff
@@ -0,0 +1,103 @@
...
+
+def generate_flake_dict(repo_id) -> FlakeDict:
+    flakes = Flake.objects.filter(repository_id=repo_id, end_date__isnull=True).all()
+    flake_dict = dict()
+    for flake in flakes:
+        flake_dict[flake.test_id] = flake
+    return flake_dict
+
+
+def update_passed_flakes(flake: Flake):
```
giovanni-guidini, line 73: [nit] can you make explicit that the function returns nothing, for consistency
```python
def update_passed_flakes(flake: Flake) -> None:
```

tasks/process_flakes.py
```diff
@@ -0,0 +1,103 @@
...
+    flake.count += 1
+    flake.recent_passes_count += 1
+    if flake.recent_passes_count == 30:
+        flake.end_date = dt.datetime.now(tz=dt.UTC)
+    flake.save()
+
+
+def upsert_failed_flake(test_instance: TestInstance, repo_id, flake: Flake | None):
```
giovanni-guidini, line 81: [nit] missing some typehints

tasks/process_flakes.py
```diff
@@ -0,0 +1,103 @@
...
+        commitid=commit_id, repoid=repo_id, branch=branch
+    ).all()
+    return test_instances
+
+
+def generate_flake_dict(repo_id) -> FlakeDict:
+    flakes = Flake.objects.filter(repository_id=repo_id, end_date__isnull=True).all()
+    flake_dict = dict()
+    for flake in flakes:
+        flake_dict[flake.test_id] = flake
```
giovanni-guidini, line 69: Is it possible for a test_id to have more than 1 flake active at the same time?
(in which case one of them would be overwritten)

tasks/tests/unit/test_process_flakes.py
```diff
@@ -0,0 +1,346 @@
...
+    upsert_failed_flake,
+)
+
+
+class RepoSimulator:
+    def __init__(self):
+        self.repo = RepositoryFactory()
+        self.repo.save()
+        self.test_count = 0
+        self.branch_name = 0
```
giovanni-guidini, line 27: [nit] I didn't expect `branch_name` to be an integer

tasks/process_flakes.py
joseph-sentry: in this version of the code the flakes are unique for every repo and test and there's only ever one that is "active" (end_date is None)

tasks/process_flakes.py
```diff
@@ -0,0 +1,107 @@
...
+                        update_passed_flakes(flake)
+                elif (
+                    test_instance.outcome == TestInstance.Outcome.FAILURE.value
+                    or test_instance.outcome == TestInstance.Outcome.ERROR.value
+                ):
+                    flake = flake_dict.get(test_instance.test_id)
+                    upserted_flake = upsert_failed_flake(test_instance, repo_id, flake)
+                    if flake is None:
+                        flake_dict[upserted_flake.test_id] = upserted_flake
+
```
michelletran-codecov, line 56: Can we also log success of this task? It would be useful for debugging.

michelletran-codecov (commented):
Generally LGTM. I have a slight concern about querying and loading i.e. 50,000 tests in memory (might be fine, but we'll see). We might want to consider doing paginated updates for the tests instances and doing smaller batch queries for the flakes table. But we can see how this performs before we add that.

tasks/process_flakes.py
```diff
@@ -0,0 +1,107 @@
...
+    test_instances = list(
+        TestInstance.objects.filter(
+            commitid=commit_id, repoid=repo_id, branch=branch
+        ).all()
+    )
+    return test_instances
+
+
+def generate_flake_dict(repo_id: int) -> FlakeDict:
+    flakes = Flake.objects.filter(repository_id=repo_id, end_date__isnull=True).all()
```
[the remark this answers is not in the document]
joseph-sentry: this query may need a separate index, because while we may be using the `reduced_error_id` field in the future we will definitely not be using the `test_id`

michelletran-codecov (approved):

PR #528, opened 2024-06-27
feat: add extra sentry metrics tagged with repoid

joseph-sentry:


tasks/test_results_finisher.py
```diff
@@ -140,6 +140,17 @@ def process_impl_within_lock(
         checkpoints = checkpoints_from_kwargs(TestResultsFlow, kwargs)
 
         checkpoints.log(TestResultsFlow.TEST_RESULTS_FINISHER_BEGIN)
+        
+        # TODO: remove this later, we can do this now because there aren't many users using this
```
michelletran-codecov, line 144: Do you have more details about when we should remove this? Should we have a ticket to remind ourselves?

michelletran-codecov (approved):
Just a question on removal criteria for the metrics and how long we're expecting to be collecting them.

tasks/test_results_finisher.py
joseph-sentry: I think it depends on the number of unique repos that end up triggering the metrics, the reason we don't want to keep emitting these metrics with the repo tag for too long is because as we get more and more unique repos using this feature the higher the [cardinality](https://docs.sentry.io/product/explore/metrics/#cardinality) of the metric will be. At some point it seems the tags will be automatically dropped if their cardinality is too high so this metric will be obsoleted anyways, so we can drop it then.

PR #529, opened 2024-06-27
fix: coverage_drop condition

giovanni-guidini:
The `coverage_drop` comment condition looks for the project status threshold,
but the config might be a simple bool. This causes the notifier to fail with
and exception.

These changes fix that and add some more testing to cover this case (that should
have been added earlier, I know. My bad)

related to: https://github.com/codecov/engineering-team/issues/1966
closes: https://github.com/codecov/engineering-team/issues/2003

### Legal Boilerplate

### DIFF
--- a/services/notification/notifiers/comment/conditions.py
+++ b/services/notification/notifiers/comment/conditions.py
@@ -154,7 +154,10 @@ class HasEnoughRequiredChanges(AsyncNotifyCondition):
         project_status_config = read_yaml_field(
             comparison.comparison.current_yaml, ("coverage", "status", "project"), {}
         )
-        threshold = Decimal(project_status_config.get("threshold", 0))
+        threshold = 0
+        if isinstance(project_status_config, dict):
+            # Project status can also be a bool value, so check is needed
+            threshold = Decimal(project_status_config.get("threshold", 0))
         target_coverage = Decimal(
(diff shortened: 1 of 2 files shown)


services/notification/notifiers/comment/conditions.py
```diff
@@ -154,7 +154,10 @@ async def _check_coverage_drop(comparison: ComparisonProxy) -> bool:
         project_status_config = read_yaml_field(
             comparison.comparison.current_yaml, ("coverage", "status", "project"), {}
         )
-        threshold = Decimal(project_status_config.get("threshold", 0))
+        threshold = 0
+        if isinstance(project_status_config, dict):
+            # Project status can also be a bool value, so check is needed
+            threshold = Decimal(project_status_config.get("threshold", 0))
```
adrian-codecov, line 160: What does it mean for threshold to be a bool? If set to True, what value does it use as a threshold?

giovanni-guidini: If the project status is a bool it means "I want project status with the default configuration". The default threshold is 0.

adrian-codecov: sounds good, ty for clarifying

adrian-codecov (approved):

PR #532, opened 2024-07-01
feat: update bundle_analysis comment requirement

giovanni-guidini:
Updates the comment requirements for bundle_analysis.
Respects the `comment.bundle_change_threshold` defined in the YAML
and only notifies if changes are bigger than the threshold.
(threshold is expected to be an int. So technically [this PR](https://github.com/codecov/shared/pull/271) is a dependency of the present changes, because it does the
coercion into an integer, as expected)

Also adds the `"bundle_increase"` as a possible requirements. In this case
on top of the threshold the change also needs to cause the bundle to _increase_
in size for a comment to be sent.

closes: https://github.com/codecov/engineering-team/issues/2021

### Legal Boilerplate

### DIFF
--- a/services/bundle_analysis.py
+++ b/services/bundle_analysis.py
@@ -3,7 +3,7 @@ import os
 import tempfile
 from dataclasses import dataclass
 from functools import cached_property
-from typing import Any, Dict, Iterator, List, Optional, Tuple
+from typing import Any, Dict, Iterator, List, Literal, Optional, Tuple
 
 import sentry_sdk
 from sentry_sdk import metrics as sentry_metrics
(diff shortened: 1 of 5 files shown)


services/bundle_analysis.py
```diff
@@ -507,6 +507,45 @@ def bundle_analysis_loader(self):
...
+                return True
+            case True:
+                return abs(comparison.total_size_delta) > changes_threshold
+            case "bundle_increase":
+                return (
+                    comparison.total_size_delta > 0
+                    and comparison.total_size_delta > changes_threshold
+                )
+            case _:
+                log.warning(
```
giovanni-guidini, line 541: This is technically dead code because the YAML validation should let it get this far.
I added this just so it wouldn't break the interface by returning None... maybe I should remove it? Even if it does return `None` it won't break the app.

joseph-sentry: I think it's fine as long as you leave a comment explaining that this code should _never_ be reached, (maybe even throw an exception? emit a metric? not sure, but we would need some way to observe that this does not happen)

joseph-sentry (approved):
lgtm, left a comment regarding the unreachable code

PR #534, opened 2024-07-02
feat: don't select suspended apps

giovanni-guidini:
Currently we ignore if an app is suspended or not.
An app is suspended by the user.
If we select a suspended app we will fail to get an access_token for it
and eventually fail with `InstallationError`

These changes filter out suspended apps from being selected.
In case all apps for the user are suspended we skip notifications.
Other tasks would fail with `NoConfiguredAppsAvailable`.

closes https://github.com/codecov/internal-issues/issues/519

### Legal Boilerplate

### DIFF
--- a/services/bots/github_apps.py
+++ b/services/bots/github_apps.py
@@ -169,6 +169,26 @@ def get_specific_github_app_details(
     )
 
 
+def _filter_rate_limited_apps(
+    apps_to_consider: List[GithubAppInstallation],
+) -> List[GithubAppInstallation]:
+    redis_connection = get_redis_connection()
+    return list(
+        filter(
+            lambda obj: not is_installation_rate_limited(
+                redis_connection, obj.installation_id, app_id=obj.app_id
+            ),
+            apps_to_consider,
+        )
+    )
+
+
+def _filter_suspended_apps(
+    apps_to_consider: List[GithubAppInstallation],
+) -> List[GithubAppInstallation]:
+    return list(filter(lambda obj: not obj.is_suspended, apps_to_consider))
+
+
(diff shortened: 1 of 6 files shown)


services/bots/github_apps.py
```diff
@@ -225,18 +245,15 @@ def get_github_app_info_for_owner(
         owner, installation_name, repository
     )
     apps_matching_criteria_count = len(apps_to_consider)
```
adrian-codecov, line 247: Does `apps_to_consider` have a hierarchy of which app to try first over another?

services/bots/github_apps.py
```diff
@@ -169,6 +169,26 @@ def get_specific_github_app_details(
...
+        filter(
+            lambda obj: not is_installation_rate_limited(
+                redis_connection, obj.installation_id, app_id=obj.app_id
+            ),
+            apps_to_consider,
+        )
+    )
+
+
+def _filter_suspended_apps(
```
adrian-codecov, line 186: Optional: making tests for this that isn't necessarily tethered to the notify test

services/bots/github_apps.py
giovanni-guidini: yes, as documented in the function that produces this list: https://github.com/codecov/worker/blob/8eb5cc4dd98e754c111552eb619b70464df9ea5a/services/bots/github_apps.py#L58

The ordering of apps in the list is the hierarchy.

adrian-codecov: I see I see

adrian-codecov (approved):
Approved, extra tests optional

PR #540, opened 2024-07-04
Fix process flakes

joseph-sentry:
- Fix bug where we weren't queuing up the process flakes task when we were supposed to
- Refine test instances query in process flakes

### DIFF
--- a/tasks/process_flakes.py
+++ b/tasks/process_flakes.py
@@ -1,9 +1,13 @@
 import datetime as dt
 import logging
 
+from django.db.models import Q
 from sentry_sdk import metrics
 from shared.celery_config import process_flakes_task_name
-from shared.django_apps.reports.models import Flake, TestInstance
+from shared.django_apps.reports.models import (
+    Flake,
+    TestInstance,
+)
 
 from app import celery_app
 from tasks.base import BaseCodecovTask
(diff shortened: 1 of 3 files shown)


tasks/process_flakes.py
```diff
@@ -61,13 +69,32 @@ def run_impl(
...
+    branch: str,
+    flaky_tests: list[str],
+) -> list[TestInstance]:
+    # get test instances on this repo commit branch combination that either:
+    # - failed
+    # - belong to an already flaky test
     test_instances = list(
         TestInstance.objects.filter(
-            commitid=commit_id, repoid=repo_id, branch=branch
+            Q(commitid=commit_id)
```
giovanni-guidini, line 87: [nit]

I think this can be refactored to be easier to read if you break up the conditions and give them a name

```python
basic_filters = Q(commitid=commit_id) & Q(repoid=repo_id) & Q(branch=branch)
test_failed_filters = Q(outcome=TestInstance.Outcome.ERROR.value) | Q(outcome=TestInstance.Outcome.FAILURE.value)
test_already_flaky_filters= Q(outcome=TestInstance.Outcome.PASS.value) & Q(test_id__in=flaky_tests)
test_instances = list( 
  TestInstance.objects.filter( 
    basic_filters & (test_failed_filters | test_already_flaky_filters)
  ).all()
)
```

giovanni-guidini (approved):
LGTM

PR #541, opened 2024-07-04
feat: create simple reduce error function

joseph-sentry:
This function will be improved in the future but for now we want a VERY simple way of distinguishing errors from each other

Partially fixes: https://github.com/codecov/engineering-team/issues/2032

### DIFF
--- a/services/failure_normalizer.py
+++ b/services/failure_normalizer.py
@@ -111,3 +111,7 @@ class FailureNormalizer:
                         # With future matches
                         failure_message = failure_message.replace(actual_match, key, 1)
         return failure_message
+
+
+def reduce_error(err_message: str) -> str:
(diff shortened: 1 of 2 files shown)


services/failure_normalizer.py
```diff
@@ -111,3 +111,7 @@ def normalize_failure_message(self, failure_message: str):
                         # With future matches
                         failure_message = failure_message.replace(actual_match, key, 1)
         return failure_message
+
+
+def reduce_error(err_message: str) -> str:
+    return "".join(err_message.split()[0:3]) + "".join(err_message.split()[-3:])
```
adrian-codecov, line 117: Will an error always take the shape of [0:3]?

joseph-sentry:
i will come back to this later

PR #543, opened 2024-07-04
chore: don't delete comments.

giovanni-guidini:
Currently the `require_changes` will deelte an existing comment if the requirements are not met but a commit exists. The commit vanishes without trace, which is not great UX.

To improve UX we are changing this behavior to bypass required changes if a comment already exists for that PR. We will update that comment.

We did verify that this does not trigger a new email being sent to users, so it should be fine.

closes: https://github.com/codecov/engineering-team/issues/2023

### Legal Boilerplate

### DIFF
--- a/services/notification/notifiers/comment/conditions.py
+++ b/services/notification/notifiers/comment/conditions.py
@@ -2,9 +2,6 @@ import logging
 from abc import ABC, abstractmethod
 from decimal import Decimal
 
-from shared.torngit.exceptions import (
-    TorngitClientError,
-)
 from shared.validation.types import (
     CoverageCommentRequiredChanges,
     CoverageCommentRequiredChangesORGroup,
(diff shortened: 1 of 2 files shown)


services/notification/notifiers/comment/conditions.py
```diff
@@ -207,6 +204,12 @@ async def check_condition_OR_group(
     async def check_condition(
         notifier: AbstractBaseNotifier, comparison: ComparisonProxy
     ) -> bool:
+        if comparison.pull and comparison.pull.commentid:
+            log.info(
```
michelletran-codecov, line 208: Should this be a warning?

michelletran-codecov (approved):
Code LGTM! 

`on_failure_side_effect` isn't really used anymore. Should we also remove this from the abstract classes? (maybe open up a new ticket to clean up `on_failure_side_effect`?)

services/notification/notifiers/comment/conditions.py
giovanni-guidini: I don't think so... It's the expected behavior after all.
I'm adding this just to help us with debugging in the future

giovanni-guidini:
> Code LGTM!
> 
> `on_failure_side_effect` isn't really used anymore. Should we also remove this from the abstract classes? (maybe open up a new ticket to clean up `on_failure_side_effect`?)

I think it should be on a separate ticket if we decide to remove that functionality, yes.
It's a good point though

PR #545, opened 2024-07-04
fix: only process flakes for head of pull

joseph-sentry:
instead of processing all the commits on a pull that was merged, it's more correct to process only the testinstances that occurred on the head of the pull, since we can only guarantee that the test instances on the head  of the pull are reflective of the state of the code on main.


adrian-codecov (approved):
Changes make sense to me

PR #547, opened 2024-07-08
feat: use merge commit sha for was_squash

joseph-sentry:
Depends on: https://github.com/codecov/shared/pull/280
We want to use the merge commit sha provided by the repo providers APIs to deduce whether the merge was done via regular merge or squash merge.

We can do this by getting the merge commit sha, then making an extra API call to get it's parents. If it has multiple parents then it's a regular merge, else it's a squash.

This commit introduces this as an experiment that is meant to run in the background. Metrics and logs were introduced in order to observe the correctness of the experiment.


tasks/sync_pull.py
```diff
@@ -379,9 +393,30 @@ def update_pull_commits(
...
+        # otherwise it's a squash
+
+        merge_commit_sha = pull_dict.get("merge_commit_sha")
+        log.info(
+            "Sync Pull using merge commit sha experiment running",
+            extra=dict(repoid=repoid, pullid=pullid, merge_commit_sha=merge_commit_sha),
+        )
+        if merge_commit_sha is not None:
+            merge_commit = repository_service.get_commit(
+                pull_dict.get("merge_commit_sha")
```
Swatinem, line 410:
```suggestion
                merge_commit_sha
```

tasks/sync_pull.py
```diff
@@ -379,9 +393,30 @@ def update_pull_commits(
...
+            )
+            if len(merge_commit["parents"]) > 1:
+                return False
+            else:
+                return True
```
Swatinem, line 415:
```suggestion
            return len(merge_commit["parents"]) <= 1
```

tasks/sync_pull.py
```diff
@@ -379,9 +393,30 @@ def update_pull_commits(
...
+        # if the merge commit exists for this PR, and that commit
+        # has multiple parents, then it's a regular merge commit
+        # otherwise it's a squash
+
+        merge_commit_sha = pull_dict.get("merge_commit_sha")
+        log.info(
+            "Sync Pull using merge commit sha experiment running",
+            extra=dict(repoid=repoid, pullid=pullid, merge_commit_sha=merge_commit_sha),
+        )
+        if merge_commit_sha is not None:
```
Swatinem, line 408:
```suggestion
        if merge_commit_sha is None:
            return None
```

Just a personal preference: early-return, and then un-indent the normal code path

Swatinem (approved):
lgtm, just some code style nits

PR #549, opened 2024-07-08
WIP

adrian-codecov:
This PR is to import token/bot logic from shared and replace its previous implementation in worker. This is WIP as of July 8th

### Legal Boilerplate

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -88,6 +88,13 @@ def db(engine, sqlalchemy_connect_url):
     Base.metadata.create_all(engine)
 
 
+@pytest.fixture(scope="session")
+def django_db_setup():
+    from django.conf import settings
+
+    settings.DATABASES["default"]["NAME"] = "background_test"
+
+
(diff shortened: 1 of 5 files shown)


conftest.py
```diff
@@ -88,6 +88,13 @@ def db(engine, sqlalchemy_connect_url):
     Base.metadata.create_all(engine)
 
 
+@pytest.fixture(scope="session")
```
adrian-codecov, line 91: Fixture to avoid django trying to recreate a DB and instead pointing to the preexisting DB

database/models/core.py
```diff
@@ -105,6 +111,31 @@ class Owner(CodecovBaseModel):
         passive_deletes=True,
     )
 
+    business_email = Column(types.Text, server_default=FetchedValue())
```
adrian-codecov, line 114: Missing fields from SQLAlchemy model compared to it's Django counterpart

PR #553, opened 2024-07-10
feat: add flaky shadow mode

joseph-sentry:
we want to run the write part of the flake detection process without doing the read part of the flake detection process

### DIFF
--- a/tasks/sync_pull.py
+++ b/tasks/sync_pull.py
@@ -20,7 +20,7 @@ from database.models import Commit, Pull, Repository
 from helpers.exceptions import RepositoryWithoutValidBotError
 from helpers.github_installation import get_installation_name_for_owner_for_task
 from helpers.metrics import metrics
-from rollouts import FLAKY_TEST_DETECTION
+from rollouts import FLAKY_SHADOW_MODE, FLAKY_TEST_DETECTION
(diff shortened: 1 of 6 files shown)


tasks/sync_pull.py
```diff
@@ -432,11 +432,12 @@ def trigger_process_flakes(
...
-            identifier=repoid, default=False
-        ) and read_yaml_field(
-            current_yaml, ("test_analytics", "flake_detection"), False
-        ):
+        if (
+            FLAKY_TEST_DETECTION.check_value(identifier=repoid, default=False)
+            and read_yaml_field(
+                current_yaml, ("test_analytics", "flake_detection"), False
+            )
+        ) or (FLAKY_SHADOW_MODE.check_value(identifier=repoid, default=False)):
```
michelletran-codecov, line 440: This `if` statement is kind of hard to read. I would suggest pulling up the conditional into a variable or function. If you pull it up into a function, we can even make this reusable in `test_results_finisher.py`.

michelletran-codecov (approved):
Generally LGTM. Just a nit suggestion.

tasks/sync_pull.py
joseph-sentry: addressed

PR #556, opened 2024-07-11
Add activate user task

michelletran-codecov:
This adds a task that links a user to an account. In particular, we want this to be called in the NotifyTask if the organization has `should_attempt_author_auto_activation` enabled.

Why is this a different task? This is mainly to isolate the models to avoid using SQLAlchemy. In the _ideal_ scenario, we would add this association directly in [the `activate_user` method](https://github.com/codecov/worker/blob/a9332e29cdd53636a6293439b751ad42f92f7238/services/activation.py#L88). The second best place would be in the [New User Activation Task](https://github.com/codecov/worker/blob/main/tasks/new_user_activated.py). However, this would mean porting the code to Django, and it is a big lift for now. So, adding this new task to _just_ do the account-user association. In the future, this task _could_ potentially be the one we migrate the New User Activation Task to.

ticket ref: https://github.com/codecov/engineering-team/issues/1993
depends on: https://github.com/codecov/shared/pull/284

### Legal Boilerplate

### DIFF
--- a/tasks/activate_account_user.py
+++ b/tasks/activate_account_user.py
@@ -1,0 +1,61 @@
+import logging
+
+from shared.celery_config import activate_account_user_task_name
+from shared.django_apps.codecov_auth.models import Account, Owner
+
+from app import celery_app
+from tasks.base import BaseCodecovTask
+
+log = logging.getLogger(__name__)
+
+
+class ActivateAccountUserTask(BaseCodecovTask, name=activate_account_user_task_name):
+    def run_impl(
+        self,
+        *,
+        user_ownerid: int,
+        org_ownerid: int,
+        **kwargs,
+    ):
+        """
+        Runs the task to activate a user onto an account.
+        :param user_ownerid: the user's owner id
+        :param org_ownerid: the organization owner id
+        """
+        log_context = {"user_ownerid": user_ownerid, "org_ownerid": org_ownerid}
+        log.info(
+            "Syncing account for user",
+            extra=log_context,
+        )
+
+        owner_user: Owner = Owner.objects.get(pk=user_ownerid)
+
+        # NOTE: We're currently ignoring organizations that don't have an account.
+        org_owner: Owner = Owner.objects.get(pk=org_ownerid)
+        account: Account | None = org_owner.account
+        if not account:
+            log.info(
+                "Organization does not have an account. Skipping account user activation."
+            )
+            return {"successful": True}
+
+        if account.can_activate_user(owner_user.user):
+            account.activate_owner_user_onto_account(owner_user)
+            account.save()
+        else:
+            log.info(
+                "User was not able to activate on account. It could be that the user is already activated, "
+                "or the account is in an inconsistent state.",
+                extra=log_context,
+            )
+
+        log.info(
+            "Successfully synced account for user",
+            extra=log_context,
+        )
+
+        return {"successful": True}
+
+
+RegisteredActivateAccountUserTask = celery_app.register_task(ActivateAccountUserTask())
+activate_account_user_task = celery_app.tasks[ActivateAccountUserTask.name]
--- a/tasks/notify.py
+++ b/tasks/notify.py
@@ -4,6 +4,7 @@ from typing import Optional
 from asgiref.sync import async_to_sync
 from celery.exceptions import MaxRetriesExceededError, SoftTimeLimitExceeded
 from shared.celery_config import (
+    activate_account_user_task_name,
     new_user_activated_task_name,
     notify_task_name,
     status_set_error_task_name,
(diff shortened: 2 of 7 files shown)


tasks/notify.py
```diff
@@ -678,6 +679,13 @@ def schedule_new_user_activated_task(self, org_ownerid, user_ownerid):
             args=None,
             kwargs=dict(org_ownerid=org_ownerid, user_ownerid=user_ownerid),
         )
+        # Activate the account user if it exists.
+        self.app.tasks[activate_account_user_task_name].apply_async(
+            kwargs=dict(
+                user_ownerid=user_ownerid,
+                org_ownerid=org_ownerid,
+            ),
+        )
```
nora-codecov, line 688: what's the difference between `celery_app.send_task()` and `self.app.tasks[activate_account_user_task_name].apply_async()`?

tasks/activate_account_user.py
```diff
@@ -0,0 +1,55 @@
...
+        account: Account | None = org_owner.account
+        if not account:
+            log.info(
+                "Organization does not have an account. Skipping account user activation."
+            )
+            return {"successful": True}
+
+        if account.can_activate_user(owner_user.user):
+            account.activate_owner_user_onto_account(owner_user)
+            account.save()
```
nora-codecov, line 44: I think we should add a log here to catch the condition when the org _does_ have an account but can_activate_user comes back **False** - what do you think?
and if this situation occurs - is the user_owner left in an in-between state or one that we would need to correct later?

nora-codecov (approved):
left some questions but I don't think they are blocking ✅

tasks/tests/unit/test_notify_task.py
```diff
@@ -235,15 +238,26 @@ def test_determine_decoration_type_from_pull_attempt_activation(
             enriched_pull.database_pull.repository.owner.ownerid,
             pr_author.ownerid,
         )
-        assert mocked_send_task.call_count == 1
-        mocked_send_task.assert_called_with(
+        assert mocked_send_task.call_count == 2
+
+        new_user_activation_call = mocked_send_task.call_args_list[0]
+        account_user_activation_call = mocked_send_task.call_args_list[1]
+        assert new_user_activation_call == call(
```
Swatinem, line 245: similar to the difference between `send_task` and `apply_async`, what is the difference between these two asserts? would it make sense to use the same patterns for both?

tasks/tests/unit/test_activate_account_user.py
```diff
@@ -0,0 +1,86 @@
...
+    assert AccountsUsers.objects.first().account == account
+    assert AccountsUsers.objects.first().user == user.user
+
+    ActivateAccountUserTask().run_impl(
+        user_ownerid=user.ownerid, org_ownerid=org.ownerid
+    )
+
+    # Nothing happens... user already exists.
+    assert AccountsUsers.objects.first().account == account
+    assert AccountsUsers.objects.first().user == user.user
```
Swatinem, line 86: I would additionally assert that there is only one row in this table, so there is no duplicate

Swatinem (commented):
I primarily asking questions here, in order to learn more about the code :-)

tasks/activate_account_user.py
michelletran-codecov: > is the user_owner left in an in-between state or one that we would need to correct later?

Yes, potentially. Unfortunately, the auto-activation code are raw SQL or a Postgres function. If we did our checks [in the activation code](https://github.com/codecov/worker/blob/main/services/activation.py), then it would be easier to keep them in sync. Anyways, I'll log when this is `False` it for now and we'll see how often this gets out of sync. To fix this properly, I'll probably want to do a SQLAlchemy/raw SQL -> Django migration for the activation stuff (as well as the new user activation task possibly). Maybe this can be a followup of this change...

michelletran-codecov: Added an info log. I think that the most likely scenario for this to be False is because the user has already been added. So, going to just keep the message level as INFO.

tasks/notify.py
michelletran-codecov: I think the main difference between `send_task` and `apply_async` is the way to reference the task. `apply_async` requires the task to be registered on the existing process, whereas `send_task` requires only then name. Otherwise, they do the same thing (starts running a new task). I'm not entirely sure why `send_task` was used here. I didn't want to change the existing call, so I left it with `send_task`, but I _think_ that it can also be called with `apply_async`.

tasks/tests/unit/test_notify_task.py
michelletran-codecov: The asserts deal with the different ways that `send_task` and `apply_async` the mocked `send_task`. Both asserts just check the call args (but the async call adds more stuff, which is why it's more complex).

PR #557, opened 2024-07-11
repository secret was being saved to repo when the update failed

nora-codecov:
fixes an issue from https://github.com/codecov/worker/pull/521


tasks/upload.py
```diff
@@ -991,13 +991,15 @@ def possibly_setup_webhooks(self, commit, repository_service):
                     )
                     repository.hookid = hookid
                     repo_data["repo"]["hookid"] = hookid
+                    repository.webhook_secret = webhook_secret
```
adrian-codecov, line 994: So we save the webhook if `should_post_webhook` or before?

adrian-codecov (approved):

tasks/upload.py
nora-codecov: yes, only add the `webhook_secret` to the `repository` if the `create_webhook` or `webhook_update` update goes through without error

PR #563, opened 2024-07-17
Add db_session to ActivateAccountUserTask

michelletran-codecov:
This field is needed to `run_impl` properly. See Sentry error: https://codecov.sentry.io/share/issue/62a996975a424922b6ceb9d0d1b91e47/

### Legal Boilerplate

### DIFF
--- a/tasks/activate_account_user.py
+++ b/tasks/activate_account_user.py
@@ -2,6 +2,7 @@ import logging
 
 from shared.celery_config import activate_account_user_task_name
 from shared.django_apps.codecov_auth.models import Account, Owner
+from sqlalchemy.orm.session import Session
 
 from app import celery_app
 from tasks.base import BaseCodecovTask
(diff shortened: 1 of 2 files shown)


nora-codecov (approved):

tasks/activate_account_user.py
```diff
@@ -12,6 +13,7 @@
 class ActivateAccountUserTask(BaseCodecovTask, name=activate_account_user_task_name):
     def run_impl(
         self,
+        _db_session: Session,
```
adrian-codecov, line 16: Potentially dumb question, but does adding a session in any way affect any transaction related things? Like will the BaseCodecovTask use the session in any way that could silently interfere with anything within the task?

michelletran-codecov: Not a dumb question! I don't think it will affect the transaction because the session is created but isn't used. So there's nothing in an uncommitted transaction to cause issues. AFAICT, the ProcessFlakesTask _also_ creates a session without using it, and there hasn't been any trouble with Django DB access.

adrian-codecov: Cool cool, yeah that makes sense, even if it were to commit something in the parent class, there's nothing in this task itself that's using the session itself

adrian-codecov (approved):

PR #570, opened 2024-07-22
Notify user of errors with uploading in comment

joseph-sentry:
fixes: https://github.com/codecov/internal-issues/issues/517

### DIFF
--- a/services/test_results.py
+++ b/services/test_results.py
@@ -2,18 +2,17 @@ import logging
 from collections import defaultdict
 from dataclasses import dataclass
 from hashlib import sha256
-from typing import List, Mapping, Sequence, Tuple
+from typing import List, Mapping, Sequence
 
-from shared.torngit.exceptions import TorngitClientError
 from shared.yaml import UserYaml
 from sqlalchemy import desc
 
 from database.enums import ReportType
 from database.models import Commit, CommitReport, RepositoryFlag, TestInstance, Upload
+from helpers.notifier import BaseNotifier
 from rollouts import FLAKY_SHADOW_MODE, FLAKY_TEST_DETECTION
 from services.report import BaseReportService
 from services.repository import (
-    fetch_and_update_pull_request_information_from_commit,
     get_repo_provider_service,
 )
 from services.yaml import read_yaml_field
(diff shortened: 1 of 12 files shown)


services/test_results.py
```diff
@@ -246,38 +221,6 @@ def build_message(self, payload: TestResultsNotificationPayload) -> str:
...
-
-        message = self.build_message(payload)
-
-        sent_to_provider = await self.send_to_provider(message)
-        if sent_to_provider == False:
-            return (False, "torngit_error")
-
-        return (True, "comment_posted")
-
-    def insert_breaks(self, table_value):
```
nora-codecov, line 270: was this in the code but not being used? 💀

tasks/notify_error.py
```diff
@@ -0,0 +1,119 @@
...
+
+@dataclass
+class ErrorNotifier(BaseNotifier):
+    failed_upload: int
+    total_upload: int
+
+    def build_message(
+        self,
+    ) -> str:
+        error_message = f"❗️ We couldn't process [{self.failed_upload}] out of [{self.total_upload}] uploads. Codecov cannot generate a coverage report with partially processed data. Please review the upload errors on the commit page."
```
nora-codecov, line 27: does this emoji reference work? When I changed notifiers with emojis before I had to use something like `:tada:` (see `_create_welcome_message`)

tasks/notify_error.py
```diff
@@ -0,0 +1,119 @@
...
+        current_yaml=None,
+        **kwargs,
+    ):
+        log.info(
+            "Starting notify error task",
+            extra=dict(commit=commitid, repoid=repoid, commit_yaml=current_yaml),
+        )
+
+        # get all upload errors for this commit
+        #
```
nora-codecov, line 47: nit: remove unused line

tasks/notify_error.py
```diff
@@ -0,0 +1,119 @@
...
+
+        checkpoints = checkpoints_from_kwargs(UploadFlow, kwargs)
+
+        commits_query = db_session.query(Commit).filter(  # type:ignore
+            Commit.repoid == repoid,
+            Commit.commitid == commitid,  # type:ignore
+        )
+
+        commit: Commit = commits_query.first()
+        assert commit, "Commit not found in database."
```
nora-codecov, line 58: can you explain this line? are you making sure that the `commit` does not exist? why does the `.first()` function return as that message?

tasks/notify_error.py
```diff
@@ -0,0 +1,119 @@
...
+
+        log.info(
+            "Notifying upload errors",
+            extra=dict(
+                repoid=repoid,
+                commitid=commitid,
+                num_failed_upload=num_failed_upload,
+                num_total_upload=num_total_upload,
+            ),
+        )
```
nora-codecov, line 82: does this throw any errors if any of the `extra`s are null or not found in the transformations above?

nora-codecov (commented):
nice job with `BaseNotifier` - helpful to have cleanups like this. Overall looks good but should be thoroughly tested in `Staging` since you're updating a handful of pieces - wanted to check that you did that testing 👍

tasks/notify_error.py
joseph-sentry: if they're null it'll display null, but if they are undefined it will throw an error, but i don't think it's possible for them to be undefined (or else my ide would probably complain)

tasks/notify_error.py
joseph-sentry: it's the opposite, we're making sure that commit exists, and if it doesn't we throw the assertion error with that message

services/test_results.py
joseph-sentry: yes 😅

tasks/notify_error.py
nora-codecov: aaahhhh, i was reading it as though the `assert` was asserting that `element 1` and `element 2` were equal, rather than that `element 1` is truthy else throw `element 2`

nora-codecov (approved):
nice

PR #573, opened 2024-07-23
feat: make newfiles part of files behaviour

joseph-sentry:
If a user specifies the files formatting in their comment layout config, we want it to show the newfiles formatting as well.

### DIFF
--- a/services/notification/notifiers/mixins/message/__init__.py
+++ b/services/notification/notifiers/mixins/message/__init__.py
@@ -275,13 +275,9 @@ class MessageMixin(object):
         if all(x not in sections for x in headers):
             sections.insert(0, "condensed_header")
 
-        print(sections)
-        if (
-            "files" in sections
-            and "newfiles" not in sections
-            and "condensed_files" not in sections
-        ):
-            sections.insert(0, "newfiles")
+        for i, section in enumerate(sections):
+            if section == "files" or section == "tree":
+                sections.append("newfiles")
 
         return [
             section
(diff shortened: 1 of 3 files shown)


services/notification/notifiers/mixins/message/__init__.py
```diff
@@ -275,6 +275,14 @@ def get_upper_section_names(self, settings):
         if all(x not in sections for x in headers):
             sections.insert(0, "condensed_header")
 
+        print(sections)
```
giovanni-guidini, line 278: print

services/notification/notifiers/mixins/message/__init__.py
```diff
@@ -275,6 +275,14 @@ def get_upper_section_names(self, settings):
...
+        if (
+            "files" in sections
+            and "newfiles" not in sections
+            and "condensed_files" not in sections
+        ):
+            sections.insert(0, "newfiles")
```
giovanni-guidini, line 284: so we're gonna show "files" AND "newfiles" ?...

The question is: do we pop the "files" entry anywhere?
Like instead of this `if` can we do `sections.replace("files", "condensed_files")` ?

Also let's prefer `condensed_files` instead of `new_files` cause that's not new anymore

joseph-sentry: We indeed want to show both if `files` has been selected.

adrian-codecov (approved):
Left a suggestion, otherwise lgtm

PR #574, opened 2024-07-26
Add a metric for the raw report size

Swatinem:
It might be nice to have a histogram with the raw size in bytes for various report formats

### DIFF
--- a/services/report/report_processor.py
+++ b/services/report/report_processor.py
@@ -3,7 +3,7 @@
 import logging
 import numbers
 from json import load
-from typing import Any, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 from lxml import etree
 from sentry_sdk import metrics as sentry_metrics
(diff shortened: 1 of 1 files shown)


services/report/report_processor.py
```diff
@@ -55,6 +55,26 @@
...
+        10 * KiB,
+        100 * KiB,
+        200 * KiB,
+        500 * KiB,
+        1 * MiB,
+        2 * MiB,
+        5 * MiB,
+        10 * MiB,
+        20 * MiB,
+        50 * MiB,
```
matt-codecov, line 74: can you add some higher buckets? maybe 200, 500, 1000, 2000 or something? measurements higher than the largest bucket are effectively clamped. if your 75th percentile is larger than your max bucket, a p75 graph will lie to you and say the max bucket size is your p75

we don't know the distribution of these sizes so it maybe won't be an issue, but i don't think we currently deal with so much prometheus data that the extra buckets will kill us. we could also thin out some of the lower bucket granularities if we are concerned about that, since for our purposes there is not much difference between 10kb and 100kb or 1mb and 2mb

(i like the `KiB` and `MiB` constants + thanks for a good metric description!)

services/report/report_processor.py
```diff
@@ -159,13 +179,14 @@ def process_report(
...
         return None
-    processors = get_possible_processors_list(report_type)
+    processors = get_possible_processors_list(report_type) if report_type else []
     for processor in processors:
         if processor.matches_content(parsed_report, first_line, name):
             sentry_metrics.incr(
                 "services.report.report_processor.parser",
                 tags={"type": type(processor).__name__},
             )
+            RAW_REPORT_SIZE.labels(processor=processor.name).observe(report.size)
```
matt-codecov, line 189: checked the callsite for this function to double-check that `report` here is indeed a single raw coverage report embedded inside of a larger Codecov upload https://github.com/codecov/worker/blob/a07ddef3a3ba329c9cda901d92fe345a7abb0375/services/report/raw_upload_processor.py#L159-L161

matt-codecov (approved):
lgtm! would like to see buckets for larger report sizes, but also once this is deployed i imagine we'll immediately query it in grafana and see if there's even an issue. it's easy to change the bucket sizes later

PR #576, opened 2024-07-30
Major refactor of BundleAnalysis notification service

giovanni-guidini:


michelletran-codecov (commented):
Just some thoughts

giovanni-guidini:
I'm closing this now because the changes are in. This PR was only to show the "big picture".

It's unfortunate that I missed the review comments here because there were very good insights here, and we should probably make those happen.

I guess they will be left for the next iteration:
* changing the context interfaces to be more declarative;
* rely on composition over inheritance;
* get rid of the not-quite-right builder pattern

PR #579, opened 2024-07-31
refactor: contexts and builders

giovanni-guidini:
introduces the NotificationContext and ContextBuilder classes.
the only specialization currently is for the PR comment.

the context builders generate the context details necessary to create and
send the message later on. Each step is called in order, and if one step fails
we bail with an exception. This means the notification can't be sent.


services/bundle_analysis/new_notify/contexts/__init__.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @cached_property
+    def repository_service(self) -> TorngitBaseAdapter:
+        return get_repo_provider_service(
+            self.repository,
+            installation_name_to_use=self.gh_app_installation_name,
+        )
+
+    @property
+    def commit_report(self) -> CommitReport:
```
adrian-codecov, line 52: For this and other properties that have a getter/setter, wouldn't accessing the property without setting a value for it first lead to an AttributeError? I think we should add some validation to each, something like 

``` python
@property
def commit_report(self) -> CommitReport:
    if self._commit_report is None:
        raise ValueError("...")
    return self._commit_report
```

Or an alternative would be to add None placeholders for properties that aren't initialized in the constructor, like 

```
    def __init__(
        self, commit: Commit, current_yaml: UserYaml, gh_app_installation_name: str
    ) -> None:
        self.commit = commit
        self.current_yaml = current_yaml
        self.gh_app_installation_name = gh_app_installation_name
        self._commit_report = None
```

Or, there's the approach of not using getters/setters if we choose the None route above. It would be a bit unclearer to know where that property gets populated, but reading this doc https://realpython.com/python-getter-setter/ gave me the impression that the main point of the setters here would be to add additional requirements/behaviors to the value being set, while we're just setting the property to a value here. 

I'm okay with whichever style we choose to get/set these properties outside of the constructor, but I would make sure we have a way to deal with accessing a non-initialized variable, like commit_report here

services/bundle_analysis/new_notify/contexts/__init__.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class WrongContextBuilderError(Exception):
+    pass
+
+
+class NotificationContextBuilder:
+    """Creates the BaseBundleAnalysisNotificationContext one step at a time, in the correct order."""
+
+    # TODO: Find a way to re-use the base-context (by cloning it instead of recreating it for every notification)
```
adrian-codecov, line 82: Could we try a singleton pattern approach perhaps? Make it once, save it, and if it exists already, use it?

services/bundle_analysis/new_notify/contexts/comment.py
```diff
@@ -0,0 +1,156 @@
...
+from services.bundle_analysis.new_notify.types import NotificationType
+from services.repository import (
+    EnrichedPull,
+    fetch_and_update_pull_request_information_from_commit,
+)
+
+log = logging.getLogger(__name__)
+
+
+class BundleAnalysisCommentNotificationContext(BaseBundleAnalysisNotificationContext):
```
adrian-codecov, line 32: Optional: since we call the notification type `PR_COMMENT`, would it make sense to call this `BundleAnalysisPRCommentNotificationContext` comment instead?

services/bundle_analysis/new_notify/contexts/comment.py
```diff
@@ -0,0 +1,156 @@
...
+    """Context for the Bundle Analysis PR Comment. Extends BaseBundleAnalysisNotificationContext."""
+
+    notification_type = NotificationType.PR_COMMENT
+
+    @property
+    def pull(self) -> EnrichedPull:
```
adrian-codecov, line 38: Same comment about empty property handling if applicable. I'm not sure if the builder pattern "handles' this question by virtue handling it in the builder class itself

services/bundle_analysis/new_notify/contexts/comment.py
```diff
@@ -0,0 +1,156 @@
...
+        Configuration is done via UserYaml.
+        If a comment was previously made for this PR the required changes are bypassed so that we
+        update the existing comment with the latest information.
+        Raises:
+            NotificationContextBuildError: required changes are not met.
+        """
+        current_yaml = self._notification_context.current_yaml
+        required_changes: bool | Literal["bundle_increase"] = (
+            current_yaml.read_yaml_field(
+                "comment", "require_bundle_changes", _else=False
```
adrian-codecov, line 119: Optional: using keyword arguments for what comment and require_bundle_changes here are. While it's a bit more verbose, I find it helps me read the code better, but it's also a stylistic preference so I'm okay either way

services/bundle_analysis/new_notify/contexts/comment.py
```diff
@@ -0,0 +1,156 @@
...
+            log.info(
+                "Skipping required_changes verification because comment already exists",
+                extra=dict(
+                    pullid=pull.database_pull.id,
+                    commitid=self._notification_context.commit.commitid,
+                ),
+            )
+            return self
+        comparison = self._notification_context.bundle_analysis_comparison
+        should_continue = {
```
adrian-codecov, line 136: I found this a bit hard to read at first - the `False: True` broke my brain 😂, but after going through it see why you did it this way. Perhaps a case statement could help with readability, I trust your judgement 👍

giovanni-guidini: It's funny you say that because I was originally going to go with a `match` statement. But you can't directly assign to a variable using that.

Looking online some stack overflow question suggested this pattern and I was like "ok cool"

services/bundle_analysis/new_notify/contexts/comment.py
giovanni-guidini: What do you mean by keyword arguments?
for `"comment" and "require_bundle_changes"` ? You can't do that in this particular function I don't think...

services/bundle_analysis/new_notify/contexts/__init__.py
giovanni-guidini: > While this is currently used by the Builder class where it deals with this somewhat, nothing prevents this class being used elsewhere in the future so just doing some damage control for then 👌

I think the whole idea of having a [builder pattern](https://refactoring.guru/design-patterns/builder) is exactly to exclusively use the builder to create the object, and use it in a more `read_only` manner.
Which means that the creating of the Context shouldn't happen without the builder (except possibly for tests, but we are careful with those)

But yeah, you are correct that it's not clear if you happen to use the context without initializing it why this happened.
I will follow the 1st suggestion of having a custom error that indicates "you should build the Context using the builder before using it"

services/bundle_analysis/new_notify/contexts/__init__.py
giovanni-guidini: Singleton I don't think was the answer.
I did something tho, in which you can initialize the builder from an existing context by looking at what fields the existing context has and porting them over to the new one.

Because notification is a read-only process it's OK to have the same reference in multiple places.

services/bundle_analysis/new_notify/contexts/comment.py
adrian-codecov: Yeaah I suppose you'd have to write the return statement in each case instead of assigning a variable. We can leave it as is

services/bundle_analysis/new_notify/contexts/comment.py
adrian-codecov: Oh I see, read_yaml_field accepts a positional argument in `def read_yaml_field(self, *keys, _else=None)`, so you can't explicitly assign `keys=(...)`, you'd have to modify the fn's internals. Okay it makes sense to leave it as is since the fn expects positional arguments

adrian-codecov (approved):

services/bundle_analysis/new_notify/contexts/__init__.py
adrian-codecov: Huh, does this have a pattern?? Not really familiar w/ this

matt-codecov (approved):
didn't review the tests, but this is clean and i can see how this would be used for our other notification paths. great work!

services/bundle_analysis/new_notify/contexts/comment.py
```diff
@@ -0,0 +1,171 @@
...
+        if not should_continue:
+            raise NotificationContextBuildError("evaluate_has_enough_changes")
+        return self
+
+    def build_context(self) -> "BundleAnalysisCommentContextBuilder":
+        super().build_context()
+        async_to_sync(self.load_enriched_pull)()
+        self.load_bundle_comparison()
+        self.evaluate_has_enough_changes()
+        return self
```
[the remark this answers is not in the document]
giovanni-guidini: > so you don't really need the builder pattern exactly

You're probably right... I do want to avoid the load function being declared in the Context though... I would prefer to have the context be more similar to a dataclass.

I think using composition for the load steps (they are declared in different classes that are composed into the context) and the context itself declared the build steps might reduce the amount of code we have and overall be more flexible and easier to reuse steps

services/bundle_analysis/new_notify/contexts/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+    """Base NotificationContext for bundle analysis notifications.
+    It includes basic information that all bundle analysis notifications need.
+    Use NotificationContextBuilder to populate the context.
+
+    Example:
+      builder = NotificationContextBuilder(commit, current_yaml, GITHUB_APP_INSTALLATION_DEFAULT_NAME)
+      notification_context = builder.build_context().get_result()
+    """
+
+    notification_type: NotificationType
```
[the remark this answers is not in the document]
giovanni-guidini: yes... I think to notify when the context fails to build. Might be in a later PR

PR #580, opened 2024-07-31
refactor: template-based messages

giovanni-guidini:
Introduces the MessageStrategy and the PR_comment specialization.

Templates for Markdown are more annoying than expected because of the `\n` after the tags.
But it's nice to have them in any case.

The process of building and sending the message depends on the NotificationContext that
is created previously.

### DIFF
--- a/services/bundle_analysis/new_notify/messages/comment.py
+++ b/services/bundle_analysis/new_notify/messages/comment.py
@@ -1,0 +1,115 @@
+import logging
+from typing import TypedDict
+
+from django.template import loader
+from shared.bundle_analysis import (
+    BundleAnalysisComparison,
+    BundleChange,
+)
+from shared.torngit.exceptions import TorngitClientError
+
+from services.bundle_analysis.new_notify.contexts.comment import (
+    BundleAnalysisPRCommentNotificationContext,
+)
+from services.bundle_analysis.new_notify.helpers import (
+    bytes_readable,
+    get_github_app_used,
+)
+from services.bundle_analysis.new_notify.messages import MessageStrategyInterface
+from services.notification.notifiers.base import NotificationResult
+from services.urls import get_bundle_analysis_pull_url
+
+log = logging.getLogger(__name__)
+
+
+class BundleRow(TypedDict):
+    bundle_name: str
+    bundle_size: str
+    change_size_readable: str
+    change_icon: str
+
+
+class BundleCommentTemplateContext(TypedDict):
+    pull_url: str
+    total_size_delta: int
+    total_size_readable: str
+    bundle_rows: list[BundleRow]
+
+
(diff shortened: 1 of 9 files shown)


services/bundle_analysis/new_notify/messages/comment.py
```diff
@@ -0,0 +1,115 @@
...
+            comment_id = pull.bundle_analysis_commentid
+            if comment_id:
+                await repository_service.edit_comment(pull.pullid, comment_id, message)
+            else:
+                res = await repository_service.post_comment(pull.pullid, message)
+                pull.bundle_analysis_commentid = res["id"]
+            return NotificationResult(
+                notification_attempted=True,
+                notification_successful=True,
+                github_app_used=get_github_app_used(repository_service),
```
adrian-codecov, line 66: Thinking about the entity work we've been doing, is there a reason we favour knowing the app that's used vs the entity that's used? I suppose the whole entity initiative isn't mature enough to be the "norm", and it's handier to know the gh app vs knowing nothing (although it can be none). I'm wondering if we think it's valuable to recognize the entity moving forward, so having the NotificationResult accept an entity_used = get_github_app_used(repository_service) (for now) and eventually open it up to be anything. Wdyt?

services/bundle_analysis/new_notify/messages/templates/bundle_analysis_notify/bundle_comment.md
```diff
@@ -0,0 +1,9 @@
+## [Bundle]({{pull_url}}) Report
```
adrian-codecov, line 1: Sweet!

adrian-codecov (commented):
Skipped over the overlapping logic bw this and https://github.com/codecov/worker/pull/579/files, generally looks good. Just left a q but otherwise lgtm

services/bundle_analysis/new_notify/messages/comment.py
giovanni-guidini: I do think it's a good idea to generalize that field. It seems like it would be a useful thing in debugging. (But that's not related to the changes in this PR)

One thing to note is that the filed now is simply an ID. That get's saved in [CommitNotification](https://github.com/codecov/shared/blob/58c6f4f0a0fa1a5d931466e0a63f5f15349546d5/shared/django_apps/core/models.py#L475) (so we can update it later for users with multiple apps). A simple ID doesn't seem enough to differentiate between different entities.

adrian-codecov: Hmm okok good to know, should we change this in the future we can make it more explicit if we're tracking an id or perhaps follow the entity string convention. It's good for now

adrian-codecov (approved):

PR #581, opened 2024-07-31
refactor: service

giovanni-guidini:
Ties the previous PRs together in a new service for bundle analysis notifications.

### DIFF
--- a/services/bundle_analysis/new_notify/__init__.py
+++ b/services/bundle_analysis/new_notify/__init__.py
@@ -83,7 +83,7 @@ class BundleAnalysisNotifyService:
         Each NotificationType is paired with a ContextBuilder and MessageStrategyInterface.
         The MessageStrategy is later used to build and send the message based on the NotificationContext
         """
-        builders_lookup: dict[
+        notifier_lookup: dict[
             NotificationType,
             tuple[NotificationContextBuilder, MessageStrategyInterface],
         ] = {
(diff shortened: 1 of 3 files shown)


services/bundle_analysis/new_notify/__init__.py
```diff
@@ -0,0 +1,164 @@
...
+            NotificationType.PR_COMMENT: (
+                BundleAnalysisPRCommentContextBuilder,
+                BundleAnalysisCommentMarkdownStrategy,
+            )
+        }
+        builder_class, message_strategy_class = builders_lookup.get(
+            notification_type, (None, None)
+        )
+
+        if builder_class is None:
```
Swatinem, line 99: Is this supposed to be extended in the future?

If so: can you move that dict to the global scope, as it is fully static.
If not: this could be a simple `if notification_type is not NotificationType.PR_COMMENT:`, along with inlining the `message_strategy_class`.

Also, I’m not quite sure about this, but I don’t think `mypy` can reason about `message_strategy_class` based on `builder_class`, so it will likely complain that it can be `None` down below when calling it.
So it might be better to check for a non-None `.get()` result first before unpacking the tuple.

services/bundle_analysis/new_notify/__init__.py
```diff
@@ -0,0 +1,164 @@
...
+                notifications_successful=tuple(),
+            )
+
+        notification_full_contexts = filter(
+            None,
+            map(
+                partial(self.create_context_for_notification, base_context),
+                notification_types,
+            ),
+        )
```
Swatinem, line 145: Similar to the comment above, if this will never be extended beyond `NotificationType.PR_COMMENT` in the future, you can probably just filter for that type early on and simplify this code quite a bit.

services/bundle_analysis/new_notify/__init__.py
giovanni-guidini: yes, it's supposed to be extended in the future. I could use a `match` statement on `NotificationType` but I guess I prefer the dictionary approach

Swatinem (approved):

PR #583, opened 2024-08-02
add metric for overall report size

matt-codecov:
arpad added one for the individual coverage files inside of an upload in https://github.com/codecov/worker/pull/574. this PR does the same for the overall upload

an upload contains one or more coverage files, but also path fixes and the output of `git ls-files`. so it can be much larger than just the sum of its coverage files. also those coverage files can be b64-encoded which affects the size too

(the .gitignore changes help me use Sapling with this repo and aren't important)

### DIFF
--- a/services/report/__init__.py
+++ b/services/report/__init__.py
@@ -55,6 +55,11 @@ from services.redis import (
 )
 from services.report.parser import get_proper_parser
 from services.report.parser.types import ParsedRawReport
+from services.report.parser.version_one import VersionOneReportParser
+from services.report.prometheus_metrics import (
+    RAW_UPLOAD_RAW_REPORT_COUNT,
+    RAW_UPLOAD_SIZE,
+)
 from services.report.raw_upload_processor import process_raw_upload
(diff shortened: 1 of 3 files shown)


services/report/__init__.py
```diff
@@ -55,10 +55,39 @@
...
 
+print("services/report/__init__.py is being imported")
+
+KiB = 1024
+MiB = KiB * KiB
+RAW_UPLOAD_SIZE = Histogram(
+    "worker_services_report_raw_upload_size",
+    "Size (in bytes) of a raw upload (which may contain several raw reports)",
+    ["version"],
+    buckets=[
```
Swatinem, line 71: we could move this `SIZE_BUCKETS` constant to `shared.metrics` as a followup.

Swatinem (changes requested):
lgtm, though you should remove all the `print` statements before merging. I believe those are just temporary to debug some import issue?

michelletran-codecov (approved):

PR #584, opened 2024-08-02
feat: swap the notification services

giovanni-guidini:
Activates the new notification service for bundle analysis

### DIFF
--- a/tasks/bundle_analysis_notify.py
+++ b/tasks/bundle_analysis_notify.py
@@ -63,15 +63,6 @@ class BundleAnalysisNotifyTask(BaseCodecovTask, name=bundle_analysis_notify_task
         except LockRetry as retry:
             self.retry(max_retries=5, countdown=retry.countdown)
 
-    def get_success_value(
-        self, configured_notifications_count: int, successful_notifications_count: int
-    ) -> str:
-        if configured_notifications_count == 0:
-            return "nothing_to_notify"
-        if configured_notifications_count == successful_notifications_count:
-            return "full_success"
-        return "partial_success"
-
     def process_impl_within_lock(
         self,
         *,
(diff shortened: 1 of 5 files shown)


tasks/bundle_analysis_notify.py
```diff
@@ -99,15 +109,16 @@ def process_impl_within_lock(
...
                 db_session, self.name, commit.repository.owner
             )
-            notifier = BundleNotifier(
+            notifier = BundleAnalysisNotifyService(
                 commit, commit_yaml, gh_app_installation_name=installation_name_to_use
             )
-            success = async_to_sync(notifier.notify)()
+            result = notifier.notify()
+            configured_notifications_count = len(result.notifications_configured)
+            successful_notifications_count = len(result.notifications_successful)
```
Swatinem, line 121: How about instead of getting these two lengths first and then then calling `get_success_value` on them,
you have either a `NotifyResult.to_success_value`, or alternatively a `success_value_from_result(Result)`.
That would encapsulate these things better.

tasks/bundle_analysis_notify.py
```diff
@@ -64,6 +63,15 @@ def run_impl(
...
             self.retry(max_retries=5, countdown=retry.countdown)
 
+    def get_success_value(
+        self, configured_notifications_count: int, successful_notifications_count: int
+    ) -> str:
+        if configured_notifications_count == 0:
+            return "nothing_to_notify"
+        if configured_notifications_count == successful_notifications_count:
+            return "full_success"
+        return "partial_success"
```
Swatinem, line 73: would it make sense to create a strictly typed, instead of "stringly typed", enum for this?

giovanni-guidini: It would look cleaner, but these are mostly for debug so it doesn't matter as much.
But ok I can create a proper enum 👍

Swatinem (approved):

PR #585, opened 2024-08-02
Gracefully handle nr=0 in JaCoCo report

michelletran-codecov:
JaCoCo currently sometimes output line 0 for coverage report, which is a confusing behaviour. This occurs primarily when the Kotlin compiler cannot properly resolve synthetic functions. Although this should ultimately be fixed in the Kotlin-JaCoCo report generation, we're just going to skip the line to allow for the rest of the report processing to work.

ref: https://github.com/codecov/internal-issues/issues/504

### Legal Boilerplate

### DIFF
--- a/services/report/languages/tests/unit/test_jacoco.py
+++ b/services/report/languages/tests/unit/test_jacoco.py
@@ -1,8 +1,10 @@
 import datetime
+import logging
 import xml.etree.cElementTree as etree
 from time import time
 
 import pytest
+from pytest import LogCaptureFixture
 
 from helpers.exceptions import ReportExpiredException
 from services.report.languages import jacoco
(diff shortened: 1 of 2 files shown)


services/report/languages/tests/unit/test_jacoco.py
```diff
@@ -64,9 +71,16 @@ def fixes(path):
         report_builder_session = report_builder.create_report_builder_session(
             "file_name"
         )
-        report = jacoco.from_xml(
-            etree.fromstring(xml % int(time())), report_builder_session
-        )
+        with self.caplog.at_level(logging.WARNING, logger=jacoco.__name__):
```
giovanni-guidini, line 74: I didn't know you could do this, that's nice.

giovanni-guidini (approved):

Swatinem (approved):

PR #588, opened 2024-08-02
Add exception handling for rate limits

adrian-codecov:
This is a partial PR that's still in progress as I'd like a second opinion on the approach.

The idea here is, after https://github.com/codecov/shared/pull/303 and https://github.com/codecov/shared/pull/310 are merged, we'll be able to know when an app/owner is rate limited. We'll do that specifically stated in pr-310 by throwing a `EntityRateLimitedException` in the `get_adapter_auth_information` function. This is because this function is called every time we instantiate a git class, for example [here](https://github.com/codecov/worker/blob/3ff487e62f4f727b538307512c3f8dec91351d65/services/owner.py#L23-L24), so it becomes an opportune to raise this exception. This could also be individually done in the `get_owner_provider_service` and `get_provider_provider_service` functions if we deem appropriate, as that is the function that actually instantiates the git provider.

This PR is currently WIP, as I want to make sure the idea makes sense before investing a lot of time with its execution. The gist is: with 310 in, we get the rate limited exception in the `get_adapter_auth_information` function, which is called in both the `get_<>_provider_service` functions, so we'd want to try/except everywhere the `get_<>_provider_service` is instantiated in worker. I mostly want a second opinion in whether this feels appropriate or if there's an alternative for it. If appropriate, I'd continue doing this in all the instances I can see for the provider_service functions (I'm capturing a subset atm and stopped to do this PR) as well as adjust tests to ensure the exception is properly thrown. As well, I'm currently only logging a warning, while I could be doing something more.

### Legal Boilerplate


helpers/notifier.py
```diff
@@ -56,7 +57,12 @@ def build_message(self) -> str:
     async def notify(
         self,
     ) -> NotifierResult:
-        self.repo_service = get_repo_provider_service(self.commit.repository)
+        try:
+            self.repo_service = get_repo_provider_service(self.commit.repository)
+        except EntityRateLimitedException as e:
+            log.warning(
+                f"Entity {e.entity_name} rate limited trying to notify. Please try again later"
```
giovanni-guidini, line 64: What do you mean "please try again later"? There's are no users looking at this.

And the exception is caught so it won't stop the task either... What do we want to do? retry the task at a later time or actually fail it?

adrian-codecov: Yeah as I explained in the description the warning rn is a placeholder, just an example of how this would look like if I continued with the implementation.

The approach I'd like your opinion on is:

- throwing the exception in shared [here](https://github.com/codecov/shared/pull/310/files#diff-0b2eaad1f90324bc24507241c930116cca92412ec6a38973e29ce5d82134313cR69). An alternative is to throw the exception at the `get_repo_provider_service` and `get_owner_provider_service` level, wdyt?
- catching that exception everywhere the report_service/owner_service is instantiated in worker. It would be a try/catch that captures the `EntityRateLimitedException` and likely just end the task there and log something. I can add some properties when I throw the `EntityRateLimitedException` so I can log them - entity_name, ttl, etc. Maybe there's some cases where we retry if the ttl is under some amount of time, but that can likely be a later thing. 

Just wanted to hear your opinion before doing all the try/catch + tests if that didn't feel like the approach to do

PR #591, opened 2024-08-06
prometheus: add some extra buckets for upload metrics

matt-codecov:
(click on VPN) https://l.codecov.dev/S0nAAE is a dash with some new upload size / report size metrics. it has some quirks:
- `histogram_quantile` interpolates values <1 for the number of reports in an upload
- report json size, raw upload size, and raw reports per upload hit the top bucket at p99 often
- chunks file size doesn't hit the top bucket but it's pretty close

i added some buckets to help with those quirks


services/report/prometheus_metrics.py
```diff
@@ -16,12 +16,15 @@
...
 )
 
 RAW_UPLOAD_RAW_REPORT_COUNT = Histogram(
     "worker_services_report_raw_upload_raw_report_count",
     "Number of raw coverage files contained in a raw upload",
     ["version"],
-    buckets=[1, 5, 10, 20, 50],
+    # The 0.98 bucket is to stop Prometheus from interpolating values much
+    # lower than 1 in its histogram_quantile function.
+    buckets=[0.98, 1, 5, 10, 20, 50, 100],
```
Swatinem, line 29: This is really weird tbh. Will this ever emit metrics <1? In which case, we would probably want to early-return much earlier? Also, this metric would probably benefit from a bunch more smaller buckets, maybe even all the way from 1..=10?

Swatinem (approved):

services/report/prometheus_metrics.py
matt-codecov: agree this is very weird. i don't think we _expect_ any raw uploads to contain 0 raw reports, but i don't know whether it can happen anyway. that's part of why i'm introducing this weird bucket

the reason for it is: the true median may be 1.0, but all prometheus can say for sure is that it's in the bucket between 0.0 and 1.0. based on the relative size of the bucket, it will interpolate a value in that range to call the median

by adding a bucket at 0.98, the range for the estimated median is limited to 0.98-1.0. so it'll look nicer on a graph, but also, we can ask prometheus how many values lower than 0.98 were observed to know whether any 0-report uploads are coming in

matt-codecov: more granular buckets in the single digits is a good idea, i'll rebalance the buckets a bit

PR #592, opened 2024-08-06
Clean up `UploadXXX` tasks

Swatinem:
Various improvements like fixing typos, adding types and using more modern code patterns

For the `Upload` task in particular:
- Use a `LoggerAdapter` to simplify adding more logging context
- Use a custom `chunks` generator together with list comprehension instead of manual loops
  - ☝🏻 this also fixes a problem where previously the last chunk/task would not be properly flagged as `is_final`

---

As this re-indents a bunch of code, I highly recommend to use "ignore whitespace" when reviewing.

### DIFF
--- a/tasks/tests/unit/test_upload_finisher_task.py
+++ b/tasks/tests/unit/test_upload_finisher_task.py
@@ -11,7 +11,7 @@ from helpers.checkpoint_logger import CheckpointLogger, _kwargs_key
 from helpers.checkpoint_logger.flows import UploadFlow
 from tasks.upload_finisher import (
     ReportService,
-    ShouldCallNotifResult,
+    ShouldCallNotifyResult,
     UploadFinisherTask,
 )
 
@@ -225,7 +225,7 @@ class TestUploadFinisherTask(object):
             UploadFinisherTask().should_call_notifications(
                 commit, commit_yaml, processing_results, None
             )
-            == ShouldCallNotifResult.NOTIFY
+            == ShouldCallNotifyResult.NOTIFY
         )
 
     def test_should_call_notifications_local_upload(self, dbsession):
@@ -244,7 +244,7 @@ class TestUploadFinisherTask(object):
             UploadFinisherTask().should_call_notifications(
                 commit, commit_yaml, processing_results, "local_report1"
             )
-            == ShouldCallNotifResult.DO_NOT_NOTIFY
+            == ShouldCallNotifyResult.DO_NOT_NOTIFY
         )
 
     def test_should_call_notifications_manual_trigger(self, dbsession):
@@ -263,7 +263,7 @@ class TestUploadFinisherTask(object):
             UploadFinisherTask().should_call_notifications(
                 commit, commit_yaml, processing_results, None
             )
-            == ShouldCallNotifResult.DO_NOT_NOTIFY
+            == ShouldCallNotifyResult.DO_NOT_NOTIFY
         )
 
     def test_should_call_notifications_manual_trigger_off(self, dbsession):
@@ -286,14 +286,14 @@ class TestUploadFinisherTask(object):
             UploadFinisherTask().should_call_notifications(
                 commit, commit_yaml, processing_results, None
             )
-            == ShouldCallNotifResult.NOTIFY
+            == ShouldCallNotifyResult.NOTIFY
         )
 
     @pytest.mark.parametrize(
         "notify_error,result",
         [
-            (True, ShouldCallNotifResult.NOTIFY_ERROR),
-            (False, ShouldCallNotifResult.DO_NOT_NOTIFY),
+            (True, ShouldCallNotifyResult.NOTIFY_ERROR),
+            (False, ShouldCallNotifyResult.DO_NOT_NOTIFY),
         ],
     )
     def test_should_call_notifications_no_successful_reports(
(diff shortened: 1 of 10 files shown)


tasks/tests/unit/test_upload_task.py
```diff
@@ -609,8 +595,7 @@ def test_upload_task_call_multiple_processors(
                 ],
                 report_code=None,
                 in_parallel=False,
-                is_final=False,
-            ),
+                is_final=True,
```
Swatinem, line 598: This is an interesting driveby fix. The previous code would get this wrong for the last chunk happens to be on a chunk border. I initially replicated the same buggy behavior, but then found a better fix for it that correctly flags the final chunk/task.

tasks/tests/unit/test_upload_finisher_task.py
```diff
@@ -11,7 +11,7 @@
 from helpers.checkpoint_logger.flows import UploadFlow
 from tasks.upload_finisher import (
     ReportService,
-    ShouldCallNotifResult,
```
adrian-codecov, line 14: 🤣

tasks/upload_processor.py
```diff
@@ -256,7 +240,7 @@ def process_impl_within_lock(
                         repoid=repoid,
                         commit=commitid,
                         arguments=arguments,
-                        commit_yaml=commit_yaml.to_dict(),
+                        commit_yaml=commit_yaml,
```
adrian-codecov, line 243: So the commit_yaml we're logging here is the one we pass in the function definition, but before it was the one you got rid off in here, https://github.com/codecov/worker/pull/592/files#diff-4d7b2da9b8b9cf3e98040654f5510214c375f699ccafe41558b2f37519174110L212. Do you know if it's safe to delete the `commit_yaml = UserYaml(commit_yaml)` line? Could we mistakenly be logging/using the incorrect value?

tasks/upload_processor.py
```diff
@@ -372,22 +352,23 @@ def process_impl_within_lock(
...
 
             for processed_individual_report in processings_so_far:
-                deleted_archive = self._possibly_delete_archive(
-                    processed_individual_report, report_service, commit
-                )
-                if not deleted_archive:
-                    self._rewrite_raw_report_readable(
-                        processed_individual_report, report_service, commit
+                upload = processed_individual_report.pop("upload_obj", None)
+                raw_report = processed_individual_report.pop("raw_report", None)
```
adrian-codecov, line 356: I think this could be moved inside the `if upload` conditional as it's only used in the `_rewrite_raw_report_readable` fn.

tasks/upload_processor.py
```diff
@@ -522,25 +484,23 @@ def _attempt_rewrite_raw_report_readable_error_case(
 
     def _rewrite_raw_report_readable(
         self,
-        processing_result: dict,
+        raw_report: ParsedRawReport,
```
adrian-codecov, line 487: Seeing that the previous implementation had an `if raw_report`, should this be typed with optional?

tasks/upload.py
```diff
@@ -365,48 +374,26 @@ def run_impl_within_lock(
...
-                        extra=dict(
-                            repoid=upload_context.repoid,
-                            commit=upload_context.commitid,
-                            report_type=upload_context.report_type.value,
-                            countdown=retry_countdown,
-                        ),
-                    )
-                    upload_context.prepare_kwargs_for_retry(kwargs)
-                    self.retry(countdown=retry_countdown, kwargs=kwargs)
+        if retry_countdown := _should_debounce_processing(upload_context):
```
adrian-codecov, line 388: Nice to see this syntax in, last time I tried we were in an older python version an it broke production 🙃. Good reminder that we can use this!

tasks/upload.py
```diff
@@ -552,22 +529,16 @@ def run_impl_within_lock(
...
-                extra=dict(
-                    repoid=commit.repoid,
-                    commit=commit.commitid,
-                    argument_list=argument_list,
-                ),
             )
         return {"was_setup": was_setup, "was_updated": was_updated}
 
-    def fetch_commit_yaml_and_possibly_store(self, commit, repository_service):
+    def fetch_commit_yaml_and_possibly_store(self, commit: Commit, repository_service):
```
adrian-codecov, line 537: If you'd like to type `report_service`, that's TorngitBaseAdapter 👍

adrian-codecov (commented):
Thanks for this change Arpad! All for typing + better conventions. 

I left some comments although I didn't have as thorough input with the parallel processing part in uploads.py as I'm not as familiar, I'd encourage you to have someone's eyes on that if possible, also being a big change in the system. 

And have you given this a test run locally/in staging to ensure it works with these changes?

tasks/upload_processor.py
Swatinem: As far as I traced the code, there was only one user of the `UserYaml` form, and I inlined that.
Though this change assumes that `dict_form == UserYaml(dict_form).to_dict()` holds, which I’m not 100% it does?
In either case, the dict form seems to be only used for logging, which I guess is not as critical.

tasks/upload_processor.py
Swatinem: The `pop` should happen regardless though. We want to always pop both fields off the dict, even if `upload` happens to be non-existant.

tasks/upload_processor.py
Swatinem: The check for `if raw_report` was moved to the caller instead. Though one of call sites goes through `ReportService.parse_raw_report_from_storage`, and I trust that its return type annotation is correct in this case.

tasks/upload.py
Swatinem: Ah nice, that is good to know! I found out that `PreProcessUpload.fetch_commit_yaml_and_possibly_store` is a duplicated version of this function, so I moved it and its associated tests over to `services.repository`, as it was also not dependent on `self` at all.

Swatinem:
Converting this to a draft. I will be splitting this up into smaller more digestible pieces.

PR #593, opened 2024-08-06
fix: try to notify comment even when PR is closed

joseph-sentry:
Fixes part of: https://github.com/codecov/internal-issues/issues/670

### DIFF
--- a/services/notification/notifiers/comment/__init__.py
+++ b/services/notification/notifiers/comment/__init__.py
@@ -83,11 +83,11 @@ class CommentNotifier(MessageMixin, AbstractBaseNotifier):
     ) -> NotificationResult:
         # TODO: remove this when we don't need it anymore
         # this line is measuring how often we try to comment on a PR that is closed
-        if comparison.pull.state != "open":
+        if comparison.pull is not None and comparison.pull.state != "open":
             sentry_metrics.incr(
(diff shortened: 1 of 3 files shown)


services/notification/notifiers/comment/__init__.py
```diff
@@ -38,7 +37,6 @@ class CommentNotifier(MessageMixin, AbstractBaseNotifier):
     notify_conditions: List[NotifyCondition] = [
         ComparisonHasPull,
         PullRequestInProvider,
-        PullRequestOpen,
```
adrian-codecov, line 41: Could notifying on closed pull requests lead to an unforeseen increase in rate limit?

adrian-codecov (approved):
Other than my other comment, ✅

services/notification/notifiers/comment/__init__.py
joseph-sentry: yeah, that's one of the drawbacks, but i think it's at most an extra api call per PR that's closed.

adrian-codecov: sounds good, I think we can go for it and if we see a spike we likely know why. We can also, and likely, make it a yaml config option to specify commenting post merge

PR #596, opened 2024-08-07
Update base image to python:3.12.4-slim-bookworm

spalmurray-codecov:
Build was failing locally and it seems the base image we were using an out of date base image (`python:3.12.2-slim-bookworm`). Not sure in exactly what way this was breaking things, but using `python:3.12.4-slim-bookworm` fixes the issue.

### DIFF
--- a/docker/Dockerfile.requirements
+++ b/docker/Dockerfile.requirements
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1.4
-ARG PYTHON_IMAGE=python:3.12-slim-bookworm
+ARG PYTHON_IMAGE=python:3.12.4-slim-bookworm
 # BUILD STAGE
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.requirements
```diff
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1.4
-ARG PYTHON_IMAGE=python:3.12.2-slim-bookworm
+ARG PYTHON_IMAGE=python:3.12-slim-bookworm
```
joseph-sentry, line 2: I think we should pin the image to `3.12.4` because I think if they were to release a `3.12.5` this would automatically move over to that

spalmurray-codecov: Do u think we should change API as well?

matt-codecov (approved):

PR #598, opened 2024-08-07
re-export shared bot exceptions instead of redefining them

matt-codecov:
worker copy/pastes the definitions of exception types that live in shared now. worker code tries to catch the worker redefinitions, but shared is throwing the original shared definitions. even though they're named/defined identically, they're totally unrelated types. so the exceptions aren't caught

i added a unit test for the specific case i saw where an exception was not getting caught. there may be others

### DIFF
--- a/tasks/tests/unit/test_preprocess_upload.py
+++ b/tasks/tests/unit/test_preprocess_upload.py
@@ -8,7 +8,7 @@ from database.tests.factories.core import (
     ReportFactory,
     RepositoryFactory,
 )
-from helpers.exceptions import RepositoryWithoutValidBotError
+from helpers.exceptions import OwnerWithoutValidBotError, RepositoryWithoutValidBotError
 from services.report import ReportService
 from tasks.preprocess_upload import PreProcessUpload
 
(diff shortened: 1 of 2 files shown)


giovanni-guidini (approved):

tasks/tests/unit/test_preprocess_upload.py
```diff
@@ -216,6 +216,29 @@ def test_run_impl_unobtainable_lock(self, dbsession, mocker, mock_redis):
...
 
+    def test_get_repo_service_repo_and_owner_lack_bot(self, dbsession, mocker):
+        def mock_owner(foo):
+            print("in mock owner", foo)
+            raise OwnerWithoutValidBotError()
+
+        mock_owner_bot = mocker.patch(
+            "shared.bots.repo_bots.get_owner_or_appropriate_bot"
+        )
+        mock_owner_bot.side_effect = OwnerWithoutValidBotError()
```
Swatinem, line 227: Technically this mock returns the *locally imported* `helpers.exceptions.OwnerWithoutValidBotError`, which could still exhibit the same problem (as in: being a different type from `shared.bots.exceptions.OwnerWithoutValidBotError`)

Swatinem (approved):

tasks/tests/unit/test_preprocess_upload.py
matt-codecov: this is a "gotcha" for mocking, but from what i was reading python still understands `OwnerWithoutValidBotError` is the same type regardless of the import path

PR #610, opened 2024-08-12
feat: commit status message (for BA)

giovanni-guidini:
Introduces the message strategy to send commit statuses for BA.
Does _NOT_ activate these statuses yet.

There were some intermediary changes, please check the other commits for more details.


services/bundle_analysis/comparison.py
```diff
@@ -20,23 +20,31 @@
...
 
 class ComparisonLoader:
-    def __init__(self, pull: EnrichedPull):
-        self.pull = pull
+    def __init__(self, base_commit: Commit | None, head_commit: Commit | None):
+        self._base_commit: Commit | None = base_commit
+        self._head_commit: Commit | None = head_commit
+
+    @classmethod
+    def from_EnrichedPull(cls, pull: EnrichedPull) -> "ComparisonLoader":
```
Swatinem, line 28:
```suggestion
    def from_EnrichedPull(cls, pull: EnrichedPull) -> ComparisonLoader:
```

I believe this is now valid syntax in Python 3.12

services/bundle_analysis/new_notify/contexts/commit_status.py
```diff
@@ -0,0 +1,185 @@
...
+        return self
+
+    def build_context(self) -> "CommitStatusNotificationContextBuilder":
+        super().build_context()
+        async_to_sync(self.load_optional_enriched_pull)()
+        self.load_bundle_comparison()
+        self.load_commit_status_level()
+        self.load_commit_status_url()
+        self.load_cache_ttl()
+        return self
```
Swatinem, line 182:
```suggestion
        return (self.load_bundle_comparison()
            .load_commit_status_level()
            .load_commit_status_url()
            .load_cache_ttl())
```

Since you made all these methods return `Self` :-)

services/bundle_analysis/new_notify/messages/tests/test_commit_status.py
```diff
@@ -0,0 +1,260 @@
...
+class TestCommitStatusMessage:
+    @pytest.mark.parametrize(
+        "user_config, expected",
+        [
+            pytest.param(
+                {}, " Bundle change: -48.89% (Threshold: 5.0%)", id="default_config"
+            ),
+            pytest.param(
+                {"bundle_analysis": {"warning_threshold": 500000}},
+                " Bundle change: -372.56kB (Threshold: 500.0kB)",
```
Swatinem, line 62: why the leading space? Is it possible to remove that from the template?

services/bundle_analysis/new_notify/messages/commit_status.py
```diff
@@ -0,0 +1,128 @@
...
+            )
+        )
+
+    def _is_message_unchanged(
+        self, context: CommitStatusNotificationContext, message: str | bytes
+    ) -> bool:
+        """Checks if the message we are about to change was already sent."""
+        last_payload = cache.get_backend().get(self._cache_key(context))
+        print("CACHE", cache)
+        print("LAST PAYLOAD", self._cache_key(context), last_payload)
```
Swatinem, line 80:
```suggestion
```

same for the `print` down below.

services/bundle_analysis/new_notify/messages/commit_status.py
```diff
@@ -0,0 +1,128 @@
...
+        if last_payload is NO_VALUE or last_payload != message:
+            return False
+        return True
+
+    async def send_message(
+        self, context: CommitStatusNotificationContext, message: str | bytes
+    ) -> NotificationResult:
+        repository_service = context.repository_service
+        if not self._is_message_unchanged(context, message):
```
Swatinem, line 89: I would actually inline the `_is_message_unchanged` call here, as I believe its the only callsite? That would allow you to only do the (potentially expensive) `make_hash_sha256` call only once, maybe also inlining the `_cache_key` method.

And I’m personally a big fan on early returns, so I would turn this `if` around, and return the `False` result right away, and then unindent the rest of the block where you actually set the cache.

Swatinem (commented):
lgtm, left some minor comments

services/bundle_analysis/comparison.py
giovanni-guidini: My IDE complains that it is not defined. And the container fails with
```
NameError: name 'ComparisonLoader' is not defined
```

So maybe not? I might be missing something

Swatinem: Also found out that `from typing import Self` exists, and that it seems to be even a better fit when inheriting from base classes, as it will just forward the concrete type that methods are being called on. And its less to type (apart from the import)

Swatinem (approved):

PR #613, opened 2024-08-13
Add an e2e-ish test for the Upload flow

Swatinem:
Instead of relying on overly mocked unit tests, this rather adds a mostly end-to-end test. It Pushes uploads to Storage/Redis, and then invokes the `Upload` task, running it to completion. After completion of the whole upload flow, it runs assertions on the final report after all processing is complete.

### DIFF
--- a/tasks/tests/integration/test_upload_e2e.py
+++ b/tasks/tests/integration/test_upload_e2e.py
@@ -1,0 +1,181 @@
+import json
+from functools import partial
+from typing import Iterable
+from uuid import uuid4
+
+import pytest
+from redis import Redis
+from shared.reports.types import ReportLine
+from shared.yaml import UserYaml
+from sqlalchemy.orm import Session
+
+from database.tests.factories import CommitFactory, RepositoryFactory
+from services.archive import ArchiveService
+from services.redis import get_redis_connection
+from services.report import ReportService
+from tasks.tests.utils import hook_repo_provider, hook_session, run_tasks
+from tasks.upload import upload_task
+
+
+def write_raw_upload(
+    redis: Redis,
+    archive_service: ArchiveService,
+    repoid: int,
+    commitid: str,
+    contents: bytes,
+):
+    report_id = uuid4().hex
+    written_path = archive_service.write_raw_upload(commitid, report_id, contents)
+    upload = json.dumps({"reportid": report_id, "url": written_path})
+
+    redis_key = f"uploads/{repoid}/{commitid}"
+    redis.lpush(redis_key, upload)
+
+
+def lines(lines: Iterable[tuple[int, ReportLine]]) -> list[tuple[int, int]]:
+    return list(((lineno, line.coverage) for lineno, line in lines))
+
+
+@pytest.mark.integration
+@pytest.mark.django_db()
+def test_full_upload(
+    dbsession: Session,
+    mocker,
+    mock_repo_provider,
+    mock_storage,
+    mock_configuration,
+):
+    # patch various `get_db_session` imports
+    hook_session(mocker, dbsession)
+    # to not close the session after each task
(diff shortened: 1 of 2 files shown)


tasks/tests/integration/test_upload_e2e.py
```diff
@@ -0,0 +1,194 @@
...
+
+from database.tests.factories import CommitFactory, RepositoryFactory
+from services.archive import ArchiveService
+from services.report import ReportService
+from tasks.tests.unit.test_upload_task import FakeRedis
+from tasks.tests.utils import hook_session, run_tasks
+from tasks.upload import upload_task
+
+
+# TODO: instead of `mock_redis`, can we use the real thing?
```
giovanni-guidini, line 19: I think we already use containers to run the tests in CI so it might not be so hard to spin up a redis instance there...
Apparently there's also (pytest-redis)[https://pypi.org/project/pytest-redis/] that might be useful

tasks/tests/integration/test_upload_e2e.py
```diff
@@ -0,0 +1,194 @@
...
+    mock_redis,
+    mock_repo_provider,
+    mock_storage,
+    mock_configuration,
+):
+    hook_session(dbsession, mocker)
+
+    # to not close the session after each task
+    mocker.patch("tasks.base.BaseCodecovTask.wrap_up_dbsession")
+    # avoid some calls that are failing for unknown reasons:
```
giovanni-guidini, line 62: The first 2 calls actually reach out to the git provider. That might be the reason they are failing?

tasks/tests/integration/test_upload_e2e.py
```diff
@@ -0,0 +1,194 @@
...
+    )
+    mocker.patch(
+        "tasks.upload_processor.get_repo_provider_service",
+        return_value=mock_repo_provider,
+    )
+    mocker.patch(
+        "tasks.notify.get_repo_provider_service",
+        return_value=mock_repo_provider,
+    )
+    # for some reason, the `archive` down below still does not contain the files+sessions file?
```
giovanni-guidini, line 79: I suppose the files+sessions is only created by the `UploadProcessor` task, and maybe that haven't run at this point?

If the raw upload is missing then maybe...
I recently had similar issues in tests where I wrote things to archive and I had to patch this function to get it to work:
```
mocker.patch(
   "services.bundle_analysis.comparison.get_appropriate_storage_service",
   return_value=mock_storage,
)
```

tasks/tests/integration/test_upload_e2e.py
```diff
@@ -0,0 +1,194 @@
...
+                "commitid": commitid,
+            }
+        )
+
+    archive = mock_storage.storage["archive"]
+    repo_hash = ArchiveService.get_archive_hash(repository)
+    raw_chunks_path = f"v4/repos/{repo_hash}/commits/{commitid}/chunks.txt"
+    assert raw_chunks_path in archive
+
+    # TODO: where is the `report_json` ???
```
giovanni-guidini, line 149: It is usually saved in storage... in `commit._report_json_path` maybe it's the pointer to the data?

giovanni-guidini (commented):
Hoping the comments are useful

tasks/tests/integration/test_upload_e2e.py
Swatinem: I just removed all the mocking here, and I can just use the redis instance thats already running in docker which works fine.

giovanni-guidini (approved):
LGTM, I think we should merge this verison now.

I'd like to also add some considerations for "future work"

What happened to the notifications? There was a task scheduled and the requests likely failed... we might want to verify that in the future?

There are some inner details on the reports we have had issues with in the past (particularly around the sessions), it would be something else that we might add more asserts into.

The carryforward logic would benefit from having an e2e test for it as well I believe

PR #622, opened 2024-08-14
feat: add determine_seat_activation function

joseph-sentry:
This function will determine whether the PR author needs to have their seat activation given an EnrichedPull object.

This function is being added because we want to add seat activation checking for bundle analysis and test analytics.

Relevant: 
https://github.com/codecov/engineering-team/issues/2264
https://github.com/codecov/engineering-team/issues/2254

### DIFF
--- a/services/seats.py
+++ b/services/seats.py
@@ -1,0 +1,118 @@
+import logging
+from dataclasses import dataclass
+from enum import Enum
+
+from sqlalchemy import func
+from sqlalchemy.orm import Session
+
+from database.models import Owner
+from services.billing import is_pr_billing_plan
+from services.decoration import _is_bot_account
+from services.repository import EnrichedPull
+
+log = logging.getLogger(__name__)
+
+
+class ShouldActivateSeat(Enum):
+    NO_ACTIVATE = "no_activate"
+    MANUAL_ACTIVATE = "manual_activate"
+    AUTO_ACTIVATE = "auto_activate"
+
+
+@dataclass
+class SeatActivationInfo:
+    should_activate_seat: ShouldActivateSeat = ShouldActivateSeat.NO_ACTIVATE
+    owner_id: int | None = None
+    author_id: int | None = None
+    reason: str | None = None
+
+
+def determine_seat_activation(pull: EnrichedPull) -> SeatActivationInfo:
+    """
+    this function will determine if a user needs to be activated based on information about the user, their org, their repo, and the PR
+    1. if the repo is public they don't need to be activated
+    2. get repostiory owner info
+    2.1 (custom gitlab logic) if they are on gitlab we get their root group as the org instead of the repo owner
+    3. get pr author info
+    4. if org doesn't use seats, or author is included in seats already then no need to be activated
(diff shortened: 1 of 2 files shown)


services/seats.py
```diff
@@ -0,0 +1,118 @@
...
+
+    if db_pull.repository.private is False:
+        return SeatActivationInfo(reason="public_repo")
+
+    org = db_pull.repository.owner
+
+    db_session: Session = db_pull.get_db_session()
+    if org.service == "gitlab" and org.parent_service_id:
+        # need to get root group so we can check plan info
+        (gl_root_group,) = db_session.query(
```
adrian-codecov, line 63: This returns a tuple?

services/seats.py
```diff
@@ -0,0 +1,118 @@
...
+            func.public.get_gitlab_root_group(org.ownerid)
+        ).first()
+
+        org = (
+            db_session.query(Owner)
+            .filter(Owner.ownerid == gl_root_group.get("ownerid"))
```
adrian-codecov, line 69: Could gl_root_group be None? If so this .get could error

services/seats.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+
+    if not is_pr_billing_plan(org.plan):
+        return SeatActivationInfo(reason="no_pr_billing_plan")
+
+    pr_author = (
+        db_session.query(Owner)
+        .filter(
+            Owner.service == org.service,
+            Owner.service_id == provider_pull["author"]["id"],
```
adrian-codecov, line 80: If author is part of the provider_pull, isn't that the Owner object itself? 

If not, I'd move this to `provider_pull.get("author", {}).get("id)` in case author is None

services/seats.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+        return SeatActivationInfo(reason="no_provider_pull")
+
+    if db_pull.repository.private is False:
+        return SeatActivationInfo(reason="public_repo")
+
+    org = db_pull.repository.owner
+
+    db_session: Session = db_pull.get_db_session()
+    if org.service == "gitlab" and org.parent_service_id:
```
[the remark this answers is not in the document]
joseph-sentry: no

services/seats.py
joseph-sentry: i ripped this code straight from the existing `services.decoration` file, not too sure on the context: https://github.com/codecov/worker/blob/main/services/decoration.py#L99

services/seats.py
joseph-sentry: view comment above

services/seats.py
joseph-sentry: provider_pull comes from the torngit provider API calls, I think it's a good idea to use `get()` either way

adrian-codecov: oooh true true, yeah not an owner object, okay that makes sense

services/seats.py
adrian-codecov: mmm, Ill assume it's fine but we can keep an eye on this

adrian-codecov (approved):

PR #629, opened 2024-08-16
Refactor `Upload` task code a bit

Swatinem:
This does a couple of things:
- Moves some logging out to the main task function
- Removes some checks, as there is already a check for `argument_list` in the main task fn
- Also removes checks for `checkpoints`, as those should always exist, and adds assertions for that

---

This is on top of #628


tasks/upload.py
```diff
@@ -570,65 +550,51 @@ def schedule_task(
...
-        if (
-            commit_report.report_type is None
-            or commit_report.report_type == ReportType.COVERAGE.value
-        ):
-            res = self._schedule_coverage_processing_task(
+        if upload_context.report_type == ReportType.COVERAGE:
+            assert (
+                commit_report.report_type is None
+                or commit_report.report_type == ReportType.COVERAGE.value
+            )
```
Swatinem, line 562: I’m still a bit confused what the relationship between `upload_context.report_type`, `commit_report.report_type` and `report_code` is?

tasks/upload.py
```diff
@@ -381,54 +379,40 @@ def run_impl(
...
         repoid = upload_context.repoid
-        commitid = upload_context.commitid
         report_type = upload_context.report_type
-        report_code = upload_context.report_code
 
-        checkpoints = None
         if report_type == ReportType.COVERAGE:
             try:
-                checkpoints = checkpoints_from_kwargs(UploadFlow, kwargs)
+                assert checkpoints
```
giovanni-guidini, line 404: I think this raises `AssertionError` instead of `ValueError`, right?...

tasks/upload.py
giovanni-guidini: I think `upload_context.report_type` is the type the API tells the worker we should be processing.

`commit_report.report_type` is the report type for that `commit_report` (saved in the DB)

`report_code` is something different... it was intended to be used with local uploads, but I don't think anyone currently uses it. The code is most of the time "default" (I think), but using the CLI you can upload a report and give it a different code. This way you can check coverage without being impacted by other uploads (from CI for example)

tasks/upload.py
```diff
@@ -540,95 +514,87 @@ def run_impl_within_lock(
...
     ):
         commit_yaml = commit_yaml.to_dict()
 
-        res = None
-        if (
-            commit_report.report_type is None
-            or commit_report.report_type == ReportType.COVERAGE.value
-        ):
-            res = self._schedule_coverage_processing_task(
+        if upload_context.report_type == ReportType.COVERAGE:
```
giovanni-guidini, line 558: (for my own learning) is the `match` statement a less pythonic way of doing these if / elses ? I think I would personally prefer that

giovanni-guidini (approved):
LGTM, but left 1 comment about the exceptions

tasks/upload.py
Swatinem: to be quite honest, I don’t know what this would raise, because this never should. if it ever does, it is a programming error that should be fixed immediately.

tasks/upload.py
Swatinem: Good point, I’m not quite sure. I have only seen match/case statements in codecov, haven’t seen them anywhere in Sentry.
I don’t think python / mypy has proper exhaustiveness checking, so I don’t think there are any benefits from using a match/case statement.

tasks/upload.py
Swatinem: That is interesting. My confusion primarily comes from here:
- The code was using `upload_context.report_type` for creating the `XYZReportService`
- The `commit_report` is being created using only the `upload_context.report_code`
- The old code was then matching on `commit_report.report_type` for scheduling the tasks, which I changed.

Therefore, I believe that all 3 variables actually mean the same thing. Or phrased the other way around: the code would be broken in surprising ways if they didn’t.

PR #636, opened 2024-08-22
Clean up `PathFixer` related code a bit

Swatinem:
This primarily moves code around, cleans up dead code, inlines methods that only have a single caller, and renames a bunch of things for clarity.


helpers/pathmap.py
```diff
@@ -13,6 +38,21 @@ def __init__(self, *args, **kwargs):
         # Original value indicator
         self._ORIG = "\\*__orig__*//"
 
+    def resolve_path(self, path: str, ancestors=None):
```
michelletran-codecov, line 41: Great move! Quickly looking up the type for `ancestors` and `resolve_path`:

```suggestion
    def resolve_path(self, path: str, ancestors:int | None=None) -> str | None:
```

helpers/pathmap.py
```diff
@@ -13,6 +38,21 @@ def __init__(self, *args, **kwargs):
...
+        path = _clean_path(path)
+
+        new_path = self.lookup(path, ancestors)
+
+        if new_path:
+            if ancestors and not _check_ancestors(path, new_path, ancestors):
+                # path ancestor count is not valud
```
michelletran-codecov, line 48: typo
```suggestion
                # path ancestor count is not valid
```

services/path_fixer/fixpaths.py
```diff
@@ -131,7 +131,7 @@ def first_not_null_index(_list) -> Optional[int]:
 _star_to_glob = re.compile(r"(?<!\.)\*").sub
 
 
-def _fixpaths_regs(fix: str) -> str:
+def fixpaths_regs(fix: str) -> str:
     key = tuple(fix.split("::"))[0]
     # [DEPRECIATING] because handled by validators, but some data is cached in db
```
michelletran-codecov, line 136: typo
```suggestion
    # [DEPRECATED] because handled by validators, but some data is cached in db
```

helpers/tests/pathmap/test_pathmap.py
```diff
@@ -1,86 +1,68 @@
...
+        "a/b/../Path With\\ Space",
+    ]
+
+    after = [
+        None,
+        "src/components/login.js",
+        "package/__init__.py",
+        "path.py",
+        "a/Path With Space",
+    ]
```
michelletran-codecov, line 45: I would suggest using pytest [parametrize](https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids). With parametrize, the test will run all cases even if one fails.

michelletran-codecov (commented):
Refactor looks great! Just some typos

helpers/tests/pathmap/test_pathmap.py
Swatinem: Thanks for the suggestion. In this particular case, `after` serves a dual purpose, initializing the tree, and serving as assertion, so I would rather skip it in this particular case.

PR #637, opened 2024-08-23
Rewrite the `pathmap.Tree`

Swatinem:
This mostly rewrites the `Tree`, making the following changes and optimizations:
- Uses a real `Node` struct with children and terminals, instead of abusing special keys for it.
- Avoids constructing needless non-terminal strings for all intermediate nodes.
- Constructs the tree directly iteratively, instead of creating a parallel tree and merging recursively.
- Switches from recursion to iteration for `_drill`. It should be possible to also avoid recursion in lookup, but with a bit more effort.

This should primarily improve construction performance and improve memory usage, which was the primary pain points with the previous implementation.

### DIFF
--- a/helpers/pathmap.py
+++ b/helpers/pathmap.py
@@ -1,7 +1,7 @@
-import collections
-import operator
+from dataclasses import dataclass, field
 from difflib import SequenceMatcher
 from os.path import relpath
+from typing import Sequence
 
 
 def _clean_path(path):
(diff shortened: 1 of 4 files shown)


helpers/pathmap.py
```diff
@@ -28,19 +28,52 @@ def _check_ancestors(path, match, ancestors):
...
-    def __init__(self, *args, **kwargs):
-        self.instance = {}
+    def __init__(self, paths: Sequence[str]):
+        self.root = Node()
+        for path in paths:
+            self.insert(path)
+
+    def insert(self, path: str):
+        # the path components, in reverse order
+        components = reversed(path.split("/"))
```
joseph-sentry, line 66: I think it'd be a good idea to add a comment explaining why we process the path in reverse as a docstring on the Tree class. It took me a while to remember why and I think it'd be very useful for devs seeing this for the first time. Something like: "this is processed in reverse because we're trying to match paths that have differing parent dirs at the start that at some point converge to matching names for the parent dirs. For ex: `dir1/dir2/file1.txt` and `tmpdir1/tmpdir2/dir1/dir2/file1.txt`.

joseph-sentry (approved):

joseph-sentry (approved):

helpers/pathmap.py
Swatinem: Thanks for the suggestion. I added a docstring explaining the internal structure with an example.

PR #641, opened 2024-08-26
Minor `PreProcessUpload` cleanup, and more tracing

Swatinem:


giovanni-guidini (approved):
I still think we should just remove PreProcessUpload, but change LGTM

PR #645, opened 2024-08-26
copy change

nora-codecov:
small change in wording requested by design team

https://github.com/codecov/engineering-team/issues/2237


giovanni-guidini (approved):
Changes look perfectly fine to me.

But I think it's inaccurate to say that the table displays "Files with reduced coverage". In fact it displays "files with missing lines" (i.e. files changes in the diff that the patch is not 100%)

Looking at [the code that generates the warning message](https://github.com/codecov/worker/blob/438000b1d04883d314059401353f501feaca2fa4/services/notification/notifiers/mixins/message/helpers.py#L163) I don't see it checking if coverage dropped.

I would also argue that it's possible to change some 30 lines to a file, having 29 of those be covered, and 1 not covered, and have the file appear in that table, even as coverage goes up.

I suggest double checking that we really want this message. But as I said, change looks fine.

EDIT: here's the mentioned example: https://github.com/giovanni-guidini/example-python/pull/24
I will agree though that, for the most part, even a single line missing coverage will mean that coverage drops.

PR #647, opened 2024-08-27
Don't auto create bucket when not needed.

trent-codecov:
This is causing issues with GCS rate limits on calling bucket create. We know the bucket exists already.

### Legal Boilerplate


jason-ford-codecov (approved):

main.py
```diff
@@ -72,10 +72,16 @@ def setup_worker():
     storage_client = get_storage_client()
```
Morton22, line 72: https://github.com/codecov/worker/pull/647/files

Morton22 (commented):
https://github.com/codecov/worker/pull/647/files

PR #648, opened 2024-08-27
initial commit

JerrySentry:
### Legal Boilerplate

### DIFF
--- a/tasks/bundle_analysis_processor.py
+++ b/tasks/bundle_analysis_processor.py
@@ -2,6 +2,7 @@ import logging
 from copy import deepcopy
 from typing import Any, Dict
 
+from asgiref.sync import async_to_sync
 from celery.exceptions import CeleryError, SoftTimeLimitExceeded
 from shared.reports.enums import UploadState
 from shared.yaml import UserYaml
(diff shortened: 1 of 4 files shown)


tasks/bundle_analysis_processor.py
```diff
@@ -106,20 +107,29 @@ def process_impl_within_lock(
...
+
         # these are populated in the upload task
+        # unless when this task is called on a non-BA upload then we have to create an empty upload
         upload_pk = params["upload_pk"]
-        assert upload_pk is not None
+        if upload_pk is None:
+            commit_report = async_to_sync(report_service.initialize_and_save_report)(
+                commit
+            )
+            upload_pk = report_service.create_report_upload(
```
giovanni-guidini, line 123: It seems these can have a state of `uploaded` or `carryforward`. I think this is one that was carried forward?

JerrySentry: Updated. The upload_type[_id] fields are now carriedforward in this case, implemented in the `update_upload` function. (state field is for started/processed flags).

tasks/upload.py
```diff
@@ -555,6 +556,27 @@ def schedule_task(
     ):
         commit_yaml = commit_yaml.to_dict()
 
+        # Carryforward the parent BA report for the current commit's BA report when handling uploads
```
giovanni-guidini, line 559: I think this block should be moved to a dedicated function function

giovanni-guidini (approved):

tasks/upload.py
JerrySentry: Done.

PR #650, opened 2024-08-28
fix: don't require BASE report for patch in comparison

giovanni-guidini:
This changes is related to codecov/engineering-team#2387.
Currently we require that the BASE report exists to compute the comparison, which includes patch data.
In fact we only need the HEAD report and a base commit, to have a diff from the git provider.
These changes calculate the patch coverage with those assumptions, thus we no longer need a BASE report to have patch coverage.

I also removed some timers in favor of sentry tracing

### DIFF
--- a/tasks/compute_comparison.py
+++ b/tasks/compute_comparison.py
@@ -1,10 +1,13 @@
 import logging
+from typing import Literal, TypedDict
 
+import sentry_sdk
 from asgiref.sync import async_to_sync
 from shared.celery_config import compute_comparison_task_name
 from shared.components import Component
 from shared.helpers.flag import Flag
 from shared.reports.readonly import ReadOnlyReport
+from shared.reports.types import ReportTotals
 from shared.torngit.exceptions import TorngitRateLimitError
 from shared.yaml import UserYaml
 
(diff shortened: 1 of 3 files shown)


tasks/compute_comparison.py
```diff
@@ -24,51 +26,89 @@
...
+        log.info("Files impact calculated", extra=log_extra)
+        path = self.store_results(comparison, impacted_files)
 
         comparison.report_storage_path = path
-        comparison.patch_totals = impacted_files.get("changes_summary").get(
+        calculated_patch_totals = impacted_files.get("changes_summary").get(
             "patch_totals"
         )
+        if calculated_patch_totals != comparison.patch_totals:
+            log.warning(
```
Swatinem, line 103: under which circumstances would this ever happen?

Swatinem (approved):

tasks/compute_comparison.py
giovanni-guidini: I don't know, to be honest. Hopefully never.

But as I'm switching from that calculation to a copy of the one we use for the notifications I thought it would be prudent to add this just in case.

I also thought it was a good idea to always keep the result that come from the same place

PR #654, opened 2024-08-29
feat: count seats for BA notify

giovanni-guidini:
closes: https://github.com/codecov/engineering-team/issues/2254

Aligning with coverage and test analytics, PR comment for BA will
take up seats now.

### DIFF
--- a/services/bundle_analysis/new_notify/messages/comment.py
+++ b/services/bundle_analysis/new_notify/messages/comment.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Literal, TypedDict
+from typing import TypedDict
 
 import sentry_sdk
 from django.template import loader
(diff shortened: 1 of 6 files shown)


services/bundle_analysis/new_notify/messages/comment.py
```diff
@@ -50,6 +68,21 @@ def build_message(self, context: BundleAnalysisPRCommentNotificationContext) ->
...
+    def build_upgrade_message(
+        self, context: BundleAnalysisPRCommentNotificationContext
+    ) -> str:
+        template = loader.get_template("bundle_analysis_notify/upgrade_comment.md")
+        context = UpgradeCommentTemplateContext(
+            activation_link=get_members_url(context.pull.database_pull),
+            codecov_instance="self_hosted" if requires_license() else "cloud",
+            codecov_name="your instance of Codecov"
+            if requires_license()
+            else "Codecov",
```
Swatinem, line 81: Both of these depend on `requires_license()`. I would rather define a `is_saas: bool` field for the template and match on that. In particular, the template already matches on `codecov_instance` in a bunch of places, but the `codecov_name` is being resolved in python code instead.

PR #655, opened 2024-08-29
Add Sentry Events/Metrics for ParallelVerification

Swatinem:
fixes https://github.com/codecov/engineering-team/issues/2449

### DIFF
--- a/tasks/parallel_verification.py
+++ b/tasks/parallel_verification.py
@@ -1,7 +1,7 @@
-import copy
 import json
 import logging
 
+import sentry_sdk
 from shared.celery_config import parallel_verification_task_name
 
 from app import celery_app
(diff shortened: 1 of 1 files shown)


tasks/parallel_verification.py
```diff
@@ -147,6 +145,22 @@ def run_impl(
...
+            tags={"result": "success" if is_success else "failure"},
+        )
+        if not is_success:
+            with sentry_sdk.new_scope() as scope:
+                if file_level_mismatched_files:
+                    scope.set_extra("mismatched_files", file_level_mismatched_files)
+                if not top_level_totals_match:
+                    scope.set_extra("parallel_totals", parallel_report.totals.astuple())
+                    scope.set_extra("serial_totals", serial_report.totals.astuple())
+                sentry_sdk.capture_message(
```
joseph-sentry, line 160: Where does this sort of thing show up on the Sentry UI? Would this be considered an error?

joseph-sentry (approved):

tasks/parallel_verification.py
Swatinem: It shows up in the issues list, but it is not considered an error.

PR #658, opened 2024-08-29
Implement fully parallel upload processing

Swatinem:
This adds another variant to the `PARALLEL_PROCESSING` feature/rollout flag which prefers the parallel upload processing pipeline in favor of running it as an experiment.

Upload Processing can run in essentially 4 modes:
- Completely serial processing
- Serial processing, but running "experiment" code (`EXPERIMENT_SERIAL`):
  - In this mode, the final (`is_final`) `UploadProcessor` task saves a copy
    of the final report for later verification.
- Parallel processing, but running "experiment" code (`EXPERIMENT_PARALLEL`):
  - In this mode, another parallel set of `UploadProcessor` tasks runs *after*
    the main set up tasks.
  - These tasks are not persisting any of their results in the database,
    instead the final `UploadFinisher` task will launch the `ParallelVerification` task.
- Fully parallel processing (`PARALLEL`):
  - In this mode, the final `UploadFinisher` task is responsible for merging
    the final report and persisting it.

An example Task chain might look like this, in "experiment" mode:
- Upload
  - UploadProcessor
    - UploadProcessor
      - UploadProcessor (`EXPERIMENT_SERIAL` (the final one))
        - UploadFinisher
          - UploadProcessor (`EXPERIMENT_PARALLEL`)
          - UploadProcessor (`EXPERIMENT_PARALLEL`)
          - UploadProcessor (`EXPERIMENT_PARALLEL`)
            - UploadFinisher (`EXPERIMENT_PARALLEL`)
              - ParallelVerification

The `PARALLEL` mode looks like this:
- Upload
  - UploadProcessor (`PARALLEL`)
  - UploadProcessor (`PARALLEL`)
  - UploadProcessor (`PARALLEL`)
    - UploadFinisher (`PARALLEL`)

### DIFF
--- a/helpers/parallel.py
+++ b/helpers/parallel.py
@@ -1,0 +1,96 @@
+from __future__ import annotations
+
+from enum import Enum
+
+from rollouts import PARALLEL_UPLOAD_PROCESSING_BY_REPO
+
+"""
+This encapsulates Parallel Upload Processing logic
+
+Upload Processing can run in essentially 4 modes:
+- Completely serial processing
+- Serial processing, but running "experiment" code (`EXPERIMENT_SERIAL`):
+  - In this mode, the final (`is_final`) `UploadProcessor` task saves a copy
+    of the final report for later verification.
+- Parallel processing, but running "experiment" code (`EXPERIMENT_PARALLEL`):
+  - In this mode, another parallel set of `UploadProcessor` tasks runs *after*
+    the main set up tasks.
+  - These tasks are not persisting any of their results in the database,
+    instead the final `UploadFinisher` task will launch the `ParallelVerification` task.
+- Fully parallel processing (`PARALLEL`):
+  - In this mode, the final `UploadFinisher` task is responsible for merging
+    the final report and persisting it.
+
+An example Task chain might look like this, in "experiment" mode:
+- Upload
+  - UploadProcessor
+    - UploadProcessor
+      - UploadProcessor (`EXPERIMENT_SERIAL` (the final one))
+        - UploadFinisher
+          - UploadProcessor (`EXPERIMENT_PARALLEL`)
+          - UploadProcessor (`EXPERIMENT_PARALLEL`)
+          - UploadProcessor (`EXPERIMENT_PARALLEL`)
+            - UploadFinisher (`EXPERIMENT_PARALLEL`)
+              - ParallelVerification
+
+
+The `PARALLEL` mode looks like this:
+- Upload
+  - UploadProcessor (`PARALLEL`)
+  - UploadProcessor (`PARALLEL`)
+  - UploadProcessor (`PARALLEL`)
+    - UploadFinisher (`PARALLEL`)
+"""
+
+
+class ParallelFeature(Enum):
+    SERIAL = "serial"
+    EXPERIMENT = "experiment"
+    PARALLEL = "parallel"
+
+    @classmethod
+    def load(cls, repoid: int) -> ParallelFeature:
+        feature = PARALLEL_UPLOAD_PROCESSING_BY_REPO.check_value(
+            identifier=repoid, default="serial"
+        )
+
+        if feature == "experiment" or feature is True:
+            return ParallelFeature.EXPERIMENT
+        if feature == "parallel":
+            return ParallelFeature.PARALLEL
+        return ParallelFeature.SERIAL
+
+
+class ParallelProcessing(Enum):
+    SERIAL = "serial"
+    EXPERIMENT_SERIAL = "experiment-serial"
+    EXPERIMENT_PARALLEL = "experiment-parallel"
+    PARALLEL = "parallel"
+
+    @property
+    def is_parallel(self) -> bool:
+        return (
+            self is ParallelProcessing.EXPERIMENT_PARALLEL
+            or self is ParallelProcessing.PARALLEL
+        )
+
+    @classmethod
+    def from_task_args(
+        cls,
+        repoid: int,
+        in_parallel: bool = False,
+        is_final: bool = False,
+        **kwargs,
+    ) -> ParallelProcessing:
+        feature = ParallelFeature.load(repoid)
+
+        if feature is ParallelFeature.SERIAL:
+            return ParallelProcessing.SERIAL
+        if feature is ParallelFeature.PARALLEL:
+            return ParallelProcessing.PARALLEL
+
+        if in_parallel:
+            return ParallelProcessing.EXPERIMENT_PARALLEL
+        if is_final:
+            return ParallelProcessing.EXPERIMENT_SERIAL
+        return ParallelProcessing.SERIAL
--- a/services/report/__init__.py
+++ b/services/report/__init__.py
@@ -39,11 +39,9 @@ from helpers.exceptions import (
     ReportExpiredException,
     RepositoryWithoutValidBotError,
 )
+from helpers.parallel import ParallelFeature
 from helpers.telemetry import MetricContext
-from rollouts import (
-    CARRYFORWARD_BASE_SEARCH_RANGE_BY_OWNER,
-    PARALLEL_UPLOAD_PROCESSING_BY_REPO,
-)
+from rollouts import CARRYFORWARD_BASE_SEARCH_RANGE_BY_OWNER
 from services.archive import ArchiveService
 from services.report.parser import get_proper_parser
 from services.report.parser.types import ParsedRawReport
@@ -255,11 +253,10 @@ class ReportService(BaseReportService):
                 # This means there is a report to carryforward
                 self.save_full_report(commit, report, report_code)
 
+                parallel_processing = ParallelFeature.load(commit.repository.repoid)
                 # Behind parallel processing flag, save the CFF report to GCS so the parallel variant of
                 # finisher can build off of it later.
-                if PARALLEL_UPLOAD_PROCESSING_BY_REPO.check_value(
-                    identifier=commit.repository.repoid
-                ):
+                if parallel_processing is ParallelFeature.EXPERIMENT:
                     self.save_parallel_report_to_archive(commit, report, report_code)
 
         return current_report_row
@@ -790,13 +787,6 @@ class ReportService(BaseReportService):
         session = processing_result.session
 
         if processing_result.error is None:
-            # this should be enabled for the actual rollout of parallel upload processing.
-            # if PARALLEL_UPLOAD_PROCESSING_BY_REPO.check_value(
-            #     "this should be the repo id"
-            # ):
-            #     upload_obj.state_id = UploadState.PARALLEL_PROCESSED.db_id
-            #     upload_obj.state = "parallel_processed"
-            # else:
             upload.state_id = UploadState.PROCESSED.db_id
             upload.state = "processed"
             upload.order_number = session.id
@@ -1032,6 +1022,9 @@ def delete_uploads_by_sessionid(upload: Upload, session_ids: list[int]):
     """
     This deletes all the `Upload` records corresponding to the given `session_ids`.
     """
+    if not session_ids:
+        return
+
     db_session = upload.get_db_session()
     uploads = (
         db_session.query(Upload.id_)
--- a/tasks/upload.py
+++ b/tasks/upload.py
@@ -32,9 +32,9 @@ from helpers.checkpoint_logger import from_kwargs as checkpoints_from_kwargs
 from helpers.checkpoint_logger.flows import TestResultsFlow, UploadFlow
 from helpers.exceptions import RepositoryWithoutValidBotError
 from helpers.github_installation import get_installation_name_for_owner_for_task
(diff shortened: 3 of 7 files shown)


tasks/upload_processor.py
```diff
@@ -206,23 +199,27 @@ def process_impl_within_lock(
         pr = None
         report_service = ReportService(UserYaml(commit_yaml))
 
+        in_parallel = (
+            parallel_processing.is_experiment_parallel
+            or parallel_processing.is_fully_parallel
+        )
```
michelletran-codecov, line 205: This is combination of checks is done in 3 places. Maybe we can add a method/property to the `ParallelProcessing` dataclass.

helpers/parallel.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+@dataclasses.dataclass(frozen=True)
+class ParallelProcessing:
+    """
+    This encapsulates Parallel Upload Processing logic
+
+    Upload Processing can run in essentially 4 modes:
+    - Completely serial processing
+    - Serial processing, but running "experiment" code (`is_experiment_serial`):
```
michelletran-codecov, line 17: I'm not super clear about what the difference is between `is_experiment_serial` and `is_experiment_parallel`.

tasks/upload_processor.py
```diff
@@ -405,11 +405,10 @@ def process_individual_report(
         report: Report,
         upload: Upload,
         raw_report_info: RawReportInfo,
-        parallel_idx=None,
-        in_parallel=False,
+        parallel_processing: ParallelProcessing,
     ) -> ProcessingResult:
         processing_result = report_service.build_report_from_raw_content(
-            report, raw_report_info, upload=upload, parallel_idx=parallel_idx
+            report, raw_report_info, upload, parallel_processing
```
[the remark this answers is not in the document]
Swatinem: the function was modified to take a `ParallelProcessing` class.

tasks/upload.py
```diff
@@ -630,16 +637,26 @@ def _schedule_coverage_processing_task(
...
         )
-
         serial_tasks = chain(processing_tasks)
 
-        do_parallel_processing = PARALLEL_UPLOAD_PROCESSING_BY_REPO.check_value(
-            identifier=commit.repository.repoid
-        )
-
-        if not do_parallel_processing:
+        if not parallel_processing.run_experiment:
```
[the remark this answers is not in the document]
Swatinem: In the `run_fully_parallel` case, we already early-returned *just* the parallel tasks. So down here we either run the experiment, or fully serial.

helpers/parallel.py
Swatinem: When doing the "experiment", the processor and finisher task run twice for the same upload.
There is also an example flow showing those two runs, and which flags they use.

How can I make this clearer?

helpers/parallel.py
```diff
@@ -0,0 +1,100 @@
...
+        in_parallel: bool = False,
+        fully_parallel: bool = False,
+        is_final: bool = False,
+        parallel_idx: bool | None = None,
+        **kwargs,
+    ) -> Self:
+        slf = ParallelProcessing.initial(repoid)
+
+        if fully_parallel:
+            return dataclasses.replace(slf, is_fully_parallel=True)
```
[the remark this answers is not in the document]
Swatinem: I copied this pattern from some Sentry code that needed the dataclass to be hashable, which requires it to be immutable as well.
I removed the `frozen` and rather mutate in-place now.

tasks/upload_processor.py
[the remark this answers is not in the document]
michelletran-codecov: Ah, the function name was collapsed in the diff. I see this now. 👍

helpers/parallel.py
michelletran-codecov: Hmm, I think my main confusion is around their dependency (i.e. you can't have both of them be true at the same time) and about how serial is "experimental" (because it's the default option right now). Looking at the code, I guess the "default" serial path is separate from the experiment, so there is (in some paths) a distinction on "experiment" serial and "default" serial. 

To address the issue of both variables being dependent, I'm wondering if it'll be clearer if we used an enum called "ExperimentMode" or something. This way, they cannot be true at the same time, and we can just use the same variable to distinguish these cases.

matt-codecov (commented):
i think this is somewhat cleaner than the existing harness but to be honest i still feel a little uncomfortable with all the if/else branches peppered around to copy something here and skip writing something there. it feels too easy to accidentally break real processing or leave side-effects that real users will be able to see

the approach i imagine would be simpler would be a separate task that either runs nightly and chooses a batch of N commits, or is scheduled as a followup after X% of finisher tasks. this task would fetch completed report JSONs and use the sessions list from them to reconstruct `UploadTask` arguments but with dummy commits/repos owned by Codecov plugged in. one dummy repo would be overridden into the expt and the other overridden out of it. we run the identical task arguments for each repo and compare the results

with that approach, any and all copying/staging we need to do for verification can happen in one place, and there's little to no risk of our test procedure accidentally breaking things for production users or accidentally leaving side-effects that they can see. there's nothing to clean up when transitioning from validation to running the actual experiment, it's just a `Feature` with a `test` and `control` group. it doesn't faithfully reproduce carryforward inheritance, but CFF is all settled before anything changes for parallel processing anyway. i think the main downside is having to suppress GitHub API errors because our dummy repos probably won't have unique authentic commits/PRs for each batch of tasks we want to test

out of steam for the day but will see your thoughts tomorrow

helpers/parallel.py
```diff
@@ -0,0 +1,103 @@
...
+        run_fully_parallel = FULLY_PARALLEL_UPLOAD_PROCESSING_BY_REPO.check_value(
+            identifier=repoid, default=False
+        )
+        run_experiment = (
+            False
+            if run_fully_parallel
+            else PARALLEL_UPLOAD_PROCESSING_BY_REPO.check_value(
+                identifier=repoid, default=False
+            )
+        )
```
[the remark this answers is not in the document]
Swatinem: that would make a lot of sense, yes. thanks for the suggestion.

services/report/__init__.py
```diff
@@ -287,26 +285,28 @@ def initialize_and_save_report(
                 # This means there is a report to carryforward
                 self.save_full_report(commit, report, report_code)
 
+                parallel_processing = ParallelProcessing.initial(
+                    commit.repository.repoid
+                )
                 # Behind parallel processing flag, save the CFF report to GCS so the parallel variant of
                 # finisher can build off of it later. Makes the assumption that the CFFs occupy the first
                 # j to i session ids where i is the max id of the CFFs and j is some integer less than i.
```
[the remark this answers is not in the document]
Swatinem: I don’t really know TBH, this is quite some obscure code/logic that hasn’t yet revealed its secrets to me.

With the session id allocations being such a big problem, why not just use the unique database ID of the `upload`? that should ideally solve all these problems as the DB guarantees uniqueness here.

Is there any reason not to do that?

I also believe this ID allocation is one of the reasons the "time under lock" is longer than it necessarily needs to be?

Swatinem:
> to be honest i still feel a little uncomfortable with all the if/else branches peppered around to copy something here and skip writing something there.

I totally agree with this. I’m tempted to just create a new task for parallel processing which removes all the code related to handling multiple uploads in one chunk, and have ideas for further simplification ahead of time.

services/report/__init__.py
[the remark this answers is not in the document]
matt-codecov: hmm. i'm not aware of a reason we couldn't use the database PK. that might be a great idea. in every place i can think of, these session IDs are just dict keys so they don't really need to be clustered/sequential, and i don't think we have to keep them stable across reports

a little code dive:
- [`initialize_and_save_report()`](https://github.com/codecov/worker/blob/7ebdf9f9140a619b71347c4597d6d68079cb2850/services/report/__init__.py#L209) is called in `tasks/upload.py` and `tasks/preprocess_upload.py`. carryforward inheritance happens in here (more specifically, in [here](https://github.com/codecov/worker/blob/7ebdf9f9140a619b71347c4597d6d68079cb2850/services/report/__init__.py#L284-L288))
  - [`create_new_report_for_commit()`](https://github.com/codecov/worker/blob/7ebdf9f9140a619b71347c4597d6d68079cb2850/services/report/__init__.py#L816) generates the carryforward report. carried-forward sessions reuse their IDs from the parent report, this code does not know anything about the DB objects
  - [`save_full_report()`](https://github.com/codecov/worker/blob/7ebdf9f9140a619b71347c4597d6d68079cb2850/services/report/__init__.py#L1181) creates DB records for each of the sessions in the new report. we create new DB records for carried-forward sessions, we don't try to point back at the existing one
- `report.add_session()` is called in [`process_raw_upload()`](https://github.com/codecov/worker/blob/7ebdf9f9140a619b71347c4597d6d68079cb2850/services/report/raw_upload_processor.py#L191). we may not have the actual DB object here, not sure

in `save_full_report()` you could probably override the session ID for each session after you do the DB insert but before you serialize as chunks/report json, and then the `process_raw_upload()` bits are just plumbing + updating the shared API. this approach would make it harder to turn the `Upload` inserts in `save_full_report()` into a batch insert, but it should be possible. if nothing else you can get a stable order for the inserted IDs with raw SQL and then sort/update in python :P
```
with inserted as (
    insert into uploads (...)
    values (...) (...) (...)
    returning *
)
select id
from inserted
order by timestamp desc;
```

Swatinem:
I updated this PR yet again, with the following changes:
- Switched the `PARALLEL_UPLOAD_PROCESSING_BY_REPO` option to a tri-state flag
- Introduced two enums, one wrapping that feature flag, the other managing the 4 states that the various tasks can be in
- Otherwise I kept the logic mostly as is, which also means that there are still tons of `if`s scattered all around.

To be quite honest, I think just keeping the various `if`s littered around is preferable to duplicating all this logic.
Once parallel is fully enabled, there should be a lot of stuff ready to be cleaned up.

One thing that I would still have to take care of is the migration path. Rolling out the feature flag currently has a direct effect on already scheduled tasks, which should be avoided.

tasks/upload.py
```diff
@@ -638,12 +653,24 @@ def _schedule_coverage_processing_task(
...
         )
-
         serial_tasks = chain(processing_tasks)
 
-        if not do_parallel_processing:
-            return serial_tasks.apply_async()
+        if parallel_feature is ParallelFeature.EXPERIMENT:
+            parallel_shadow_experiment = serial_tasks | parallel_tasks
+            return parallel_shadow_experiment.apply_async()
+        # else:
```
michelletran-codecov, line 661: Remove comment

michelletran-codecov (approved):
Generally LGTM. I think I would also wait for @matt-codecov 's review/approval as he has more context into this code.

tasks/upload.py
```diff
@@ -606,9 +606,24 @@ def _schedule_coverage_processing_task(
...
         checkpoints.log(UploadFlow.INITIAL_PROCESSING_COMPLETE)
 
-        do_parallel_processing = PARALLEL_UPLOAD_PROCESSING_BY_REPO.check_value(
-            identifier=commit.repository.repoid
-        ) and not delete_archive_setting(commit_yaml)
+        parallel_feature = ParallelFeature.load(upload_context.repoid)
+        if parallel_feature is ParallelFeature.EXPERIMENT and delete_archive_setting(
+            commit_yaml
+        ):
+            parallel_feature = ParallelFeature.SERIAL
```
matt-codecov, line 613: do you happen to know why this setting should disable the experiment? is it a problem for the fully parallel mode?

tasks/upload_finisher.py
```diff
@@ -568,15 +602,16 @@ def merge_report(cumulative_report: Report, obj):
...
-            # be set to: parallel_processed (instead of processed)
 
             cumulative_report.merge(incremental_report)
 
-            # once the experiment is finished, we should be modifying the Upload here
-            # moving it's state from: parallel_processed -> processed
-
+            if parallel_processing is ParallelProcessing.PARALLEL:
+                # When we are fully parallel, we need to update the `Upload` in the database
+                # with the final session_id (aka `order_number`).
```
matt-codecov, line 610: is this taking the place of the `PARALLEL_PROCESSED` upload state daniel had?

matt-codecov (approved):

tasks/upload.py
Swatinem: I changed the relevant code to avoid creating a *copy* of the upload, in favor of just using the upload as it exists, provided it *does* exist and is not being deleted :-)

Parallel processing does not have that problem, as it only has a single task processing (and deleting) a raw upload.

tasks/upload_finisher.py
Swatinem: not quite. I will discuss these various states a bit more and figure out a good way to go there.

Swatinem: But good that you called this out, I found another bug related to the new code from #745 not being ported to this PR yet, which I now did.

PR #660, opened 2024-08-29
Make `ReportService` sync

Swatinem:
It looks like historically there is a mix of sync/async code.
The consensus(?) seems to favor sync code as celery is fundamentally sync.
Async code also does not play as well with profiling, which will not show the async code directly in a flamegraph, instead just inserting a `wait` for the async code running in a different thread.

### DIFF
--- a/services/bundle_analysis/report.py
+++ b/services/bundle_analysis/report.py
@@ -93,7 +93,7 @@ class ProcessingResult:
 
 
 class BundleAnalysisReportService(BaseReportService):
-    async def initialize_and_save_report(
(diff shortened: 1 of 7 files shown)


matt-codecov (approved):
nice, thank you for the cleanup! you are right that sync is preferred. i got rid of a lot of the `async`s in https://github.com/codecov/worker/pull/304 in a semi-automated fashion but didn't get em all

everything used to be labeled `async`, but the django ORM hates that and as you observed all our work is synchronous so we never had anything to yield to

PR #664, opened 2024-08-29
170 2 adjust event only to fire when due

adrian-codecov:
We use events to trigger calls to gcp. I'm modifying the events function only to fire when certain fields of interest change. Otherwise, we're hammering the google's api a bunch, which we don't need to 

### Legal Boilerplate


database/models/core.py
```diff
@@ -140,6 +140,7 @@ class Repository(CodecovBaseModel):
             random.choice(string.ascii_uppercase + string.digits) for _ in range(10)
         ),
     )
+    upload_token = Column(postgresql.UUID)
```
adrian-codecov, line 143: Nothing crazy here, we've had this field in postgres for a while, just never tracked it in worker

nora-codecov (approved):

PR #665, opened 2024-08-30
tests: extend upload e2e test

giovanni-guidini:
Extends the upload e2e test to also take into account the Comparison
that is generated when the commit is part of a PR.


Swatinem (approved):
nice 🎉

PR #666, opened 2024-08-30
Improve typing of `LanguageProcessor`s

Swatinem:
This adds a lot of missing type annotations.

It also adds tracing to all the processors, and avoids a duplicated call to `remove_non_ascii`.


services/report/languages/clover.py
```diff
@@ -14,11 +15,12 @@
 
 
 class CloverProcessor(BaseLanguageProcessor):
-    def matches_content(self, content, first_line, name):
-        return bool(content.tag == "coverage" and content.attrib.get("generated"))
+    def matches_content(self, content: Element, first_line: str, name: str) -> bool:
+        return content.tag == "coverage" and "generated" in content.attrib
```
giovanni-guidini, line 19: I think there's a slight difference here if `attrib == {"generated": None}`, if that is possible

services/report/languages/elm.py
```diff
@@ -11,21 +10,21 @@
 
 
 class ElmProcessor(BaseLanguageProcessor):
-    def matches_content(self, content, first_line, name):
-        return isinstance(content, dict) and bool(content.get("coverageData"))
+    def matches_content(self, content: dict, first_line: str, name: str) -> bool:
+        return isinstance(content, dict) and "coverageData" in content
```
giovanni-guidini, line 14: Same scenario as the other comment

services/report/report_processor.py
```diff
@@ -166,56 +188,46 @@ def get_possible_processors_list(report_type: str) -> List[Any]:
...
-                    raise
+        if not processor.matches_content(parsed_report, first_line, name):
+            continue
+        processor_name = type(processor).__name__
+
+        sentry_sdk.metrics.incr(
+            "services.report.report_processor.parser",
+            tags={"type": processor_name},
+        )
+        RAW_REPORT_SIZE.labels(processor=processor.name).observe(report.size)
```
giovanni-guidini, line 202: should this user `processor_name` instead of `processor.name`?

services/report/report_processor.py
```diff
@@ -111,29 +114,43 @@ def report_type_matching(report: ParsedUploadedReportFile) -> Tuple[Any, Optiona
...
 
 
-def get_possible_processors_list(report_type: str) -> List[Any]:
-    processor_dict: Dict[str, List[Any]] = {
-        "plist": [XCodePlistProcessor()],
-        "xml": [
+def process_report(
+    report: ParsedUploadedReportFile, report_builder: ReportBuilder
+) -> Report | None:
+    name = report.filename or ""
```
giovanni-guidini, line 136: can we call this `upload_name` so as to not get confused with `processor_name`? (I had to go back to this line to find what it was)

giovanni-guidini (approved):
Left some nit comments. Overall LGTM
(I'd wait next week to deploy this one though)

services/report/languages/clover.py
Swatinem: very good point, As these are XML tags/attributes, I’m not sure if their implementation of `in` is equivalent to dicts anyway, I will revert these cases.

PR #667, opened 2024-08-30
chore: cleanup rollout of bundle analysis notify changes

giovanni-guidini:
This replaces the `notify` for `new_notify` officially (we are already using it anyway,
but just to cleanup old code).
Changes are mostly renaming things, but I found some tests for the old notify and ported then
to the new one. And I've found a "bug", details below.

So the comment table currently displays all bundles, even if there's no change.
The old comment didn't display the table if ALL bundles had no change.

The way I changed it now it will only display bundles that have had a change in size (or removed).
I think the goal of the table is a quick glance at what changed, so it makes sense. But I should double check with design, possibly.

closes: https://github.com/codecov/engineering-team/issues/2087

### Legal Boilerplate


Swatinem (approved):
nice! always love to see obsolete code being removed

PR #670, opened 2024-09-02
Refactor Upload processing and file rewriting a bit

Swatinem:
This moved some code around and improves type annotations. I tried to make the code related to error handling and "rewrite readable" a bit more readable, but wasn’t fully successful.

This should also slightly improve perf, as it should avoid re-SELECT-ing the `upload` after it has been modified. All in all, I still don’t understand why the upload model is fetched twice, but I guess its just some magic accessor that does this implicitly.

### DIFF
--- a/tasks/upload_processor.py
+++ b/tasks/upload_processor.py
@@ -16,6 +16,7 @@ from sqlalchemy.exc import SQLAlchemyError
 from app import celery_app
 from database.enums import CommitErrorTypes
 from database.models import Commit, Upload
+from database.models.core import Pull, Repository
 from helpers.exceptions import RepositoryWithoutValidBotError
 from helpers.github_installation import get_installation_name_for_owner_for_task
 from helpers.metrics import metrics
@@ -25,9 +26,9 @@ from helpers.parallel_upload_processing import (
 )
 from helpers.save_commit_error import save_commit_error
 from rollouts import PARALLEL_UPLOAD_PROCESSING_BY_REPO
+from services.archive import ArchiveService
 from services.redis import get_redis_connection
-from services.report import Report, ReportService
-from services.report.parser.types import ParsedRawReport
+from services.report import ProcessingResult, RawReportInfo, Report, ReportService
 from services.repository import get_repo_provider_service
 from services.yaml import read_yaml_field
 from tasks.base import BaseCodecovTask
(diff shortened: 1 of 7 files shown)


tasks/upload_processor.py
```diff
@@ -322,6 +339,7 @@ def process_impl_within_lock(
                     )
             else:
                 with metrics.timer(f"{self.metrics_prefix}.save_report_results"):
+                    print(report)
```
giovanni-guidini, line 342: 👀

tasks/upload_processor.py
```diff
@@ -322,6 +339,7 @@ def process_impl_within_lock(
                     )
             else:
                 with metrics.timer(f"{self.metrics_prefix}.save_report_results"):
```
giovanni-guidini, line 341: I wonder if changing this for the sentry trace would be a good idea

tasks/upload_processor.py
```diff
@@ -436,77 +450,65 @@ def should_delete_archive(self, commit_yaml: UserYaml) -> bool:
...
         self,
-        raw_report: ParsedRawReport,
-        upload: Upload,
-        report_service: ReportService,
+        archive_service: ArchiveService,
         commit: Commit,
+        raw_report_info: RawReportInfo,
     ) -> None:
-        archive_url = upload.storage_path
+        assert raw_report_info.raw_report
```
giovanni-guidini, line 488: Hum... this could throw an exception, preventing any reports that would come after this one in the loop from being re-written.

Are we fine with that?

giovanni-guidini (commented):
Looking good, thanks for the hard work here.
Just a few comments

tasks/upload_processor.py
Swatinem: I already added a tracing annotation. will remove this metric then.

tasks/upload_processor.py
Swatinem: This is a bit unfortunate. I use `RawReportInfo` as an "out-parameter", and because of typing, `raw_report` can be `None`.

However `_rewrite_raw_report_readable` is only ever called in code paths that check that `raw_report` is in fact not `None`, and to satisfy types within that function, I assert that fact inside the function, but mostly only for the type checker.

I understand this is quite unfortunate, and ideally I would like to get rid of all the error handling that made me turn this into an out-parameter. I will take another look to figure out if I can solve things a bit differently.

giovanni-guidini (approved):

PR #671, opened 2024-09-02
Stop writing `UploadLevelTotals`

Swatinem:
I believe these are completely unused. They were still being created/updated though.


Swatinem:
I would like to still do this, yes. Though I believe I have hit a roadblock back then when working on the API changes.

PR #675, opened 2024-09-03
Tune `mypy` strictness settings

Swatinem:
In particular, this does not require files to be fully typed anymore.

I also added a couple of TODOs for checks we would want to eventually enable.

---

I now end up with the following, which on the surface looks quite bad:

```
Found 614 errors in 113 files (checked 240 source files)
```

There is quite some errors related to untyped imports. `shared` can be resolved by adding a `py.typed` file I guess.
Other dependencies would probably require typing stubs to be installed?

Then there is a ton of errors which look legit and should be fixed.

Otherwise this hits the limits of "I am not a python developer" :-D


giovanni-guidini (approved):
nice 👍

PR #677, opened 2024-09-03
Update flaky shadow mode

joseph-sentry:
This PR does 2 things:
- makes it so the flaky shadow mode feature flag also enables the read path for flake detection, meaning when that feature flag resolves to true users will also get notified of any flaky tests that they fail on in their PRs
- change the behaviour of sync pulls to avoid calling the process flakes task when a repo isn't uploading test results

### DIFF
--- a/services/test_results.py
+++ b/services/test_results.py
@@ -402,14 +402,11 @@ def latest_test_instances_for_a_given_commit(db_session, commit_id):
 def should_write_flaky_detection(repoid: int, commit_yaml: UserYaml) -> bool:
     return (
         FLAKY_TEST_DETECTION.check_value(identifier=repoid, default=False)
-        and read_yaml_field(commit_yaml, ("test_analytics", "flake_detection"), False)
         or FLAKY_SHADOW_MODE.check_value(identifier=repoid, default=False)
-    )
+    ) and read_yaml_field(commit_yaml, ("test_analytics", "flake_detection"), True)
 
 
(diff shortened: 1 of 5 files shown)


services/test_results.py
```diff
@@ -408,6 +408,8 @@ def should_write_flaky_detection(repoid: int, commit_yaml: UserYaml) -> bool:
 
 
 def should_read_flaky_detection(repoid: int, commit_yaml: UserYaml) -> bool:
-    return FLAKY_TEST_DETECTION.check_value(
-        identifier=repoid, default=False
-    ) and read_yaml_field(commit_yaml, ("test_analytics", "flake_detection"), False)
+    return (
+        FLAKY_TEST_DETECTION.check_value(identifier=repoid, default=False)
+        and read_yaml_field(commit_yaml, ("test_analytics", "flake_detection"), False)
+        or FLAKY_SHADOW_MODE.check_value(identifier=repoid, default=False)
```
michelletran-codecov, line 414: If there's some problem that we want to turn off displaying the results to users, then which flag would we want to flip?

michelletran-codecov: Turning off `FLAKY_SHADOW_MODE` would also disable writing. So, I think we might want another flag to just disable the reads.

joseph-sentry: i updated the changes such that FLAKY_SHADOW_MODE retains it's current functionality on the write path and we're going to use FLAKY_DETECTION to toggle the beta, so we can separately disable the read path without affecting the write path

michelletran-codecov (approved):

PR #679, opened 2024-09-03
fix: UI and status report diff patch numbers

giovanni-guidini:
related: https://github.com/codecov/internal-issues/issues/721
related: https://github.com/codecov/engineering-team/issues/2387

We have a problem that the UI and the statuses are reporting different numbers for patch.
Part of this issue is that the `NotifyTask` (calculates patch for status) has access to the `Pull` while the `ComputeComparison` (calculates for the UI) doesn't.

This is a problem because the base for patch coverage might be different, so numbers would be different (see [here](https://github.com/codecov/worker/blob/9078de24b4dbed1bb64250b2b3864336954f310c/tasks/compute_comparison.py#L334) that we assume no action is needed cause there's nothing we can do).

So the 1st change here is to have the `NotifyTask` calculate and save patch coverage.
The second change is to cache the patch calculation so we don't compute it more than once.

### Legal Boilerplate

### DIFF
--- a/tasks/notify.py
+++ b/tasks/notify.py
@@ -31,7 +31,11 @@ from helpers.github_installation import get_installation_name_for_owner_for_task
 from helpers.save_commit_error import save_commit_error
 from services.activation import activate_user
 from services.commit_status import RepositoryCIFilter
-from services.comparison import ComparisonContext, ComparisonProxy
+from services.comparison import (
+    ComparisonContext,
+    ComparisonProxy,
+    get_or_create_comparison,
+)
 from services.comparison.types import Comparison, FullCommit
 from services.decoration import determine_decoration_details
 from services.github import get_github_app_for_commit, set_github_app_for_commit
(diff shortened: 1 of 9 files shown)


tasks/notify.py
```diff
@@ -511,6 +514,33 @@ def has_upcoming_notifies_according_to_redis(
...
+    @sentry_sdk.trace
+    def save_patch_totals(self, comparison: ComparisonProxy) -> None:
+        """Saves patch coverage to the CompareCommit, if it exists.
+        This is done to make sure the patch coverage reported by notifications and UI is the same
+        (because they come from the same source)
+        """
+        if comparison.project_coverage_base.commit is None:
+            # This is the base that will be saved in the CommitComparison
+            # Even if the patch coverage could come from a different commit
+            return
```
matt-codecov, line 526: in other words, we identify a `CommitComparison` by its project coverage base irrespective of whether its patch coverage base is different?

matt-codecov (approved):
i think this is good? i feel weird about `save_patch_totals()` using the project coverage base, but i don't think there is a patch coverage base field on the model. and i don't think we expect there to be multiple `CommitComparison`s with the same head and project base but different patch base. so i think the logic is right?

Swatinem (approved):

tasks/notify.py
giovanni-guidini: Yes, that's right... but this was a good callout.
The Comparison is created by the `UploadFinisher` task [here](https://github.com/codecov/worker/blob/main/tasks/upload_finisher.py#L322). I thought that `Pull.compared_to` was the project base. But maybe that's not the case... I'll review before merging

giovanni-guidini: Tracking the code to [this](https://github.com/codecov/worker/blob/68d5a8930db22d978081aa6e388351083fe1ccd1/tasks/notify.py#L334) (where we decide what is the base_commit for the comparison) it seems that it's always the project base, unless there is no pull, then we use the commit's parent

So I guess the logic is fine. If we have a project base we would have created a comparison. Otherwise there is no Pull object and so there's no Comparison.

But this also means we might emit patch coverage in the UI 🤔
Should we then `get_or_create` a comparison instead of only updating if it exists... it might be a better idea

PR #683, opened 2024-09-04
Remove unused `update_archive`

Swatinem:


services/archive.py
```diff
@@ -39,27 +38,16 @@ def get_path(self, **kwaargs) -> str:
...
-    region = None
-
-    """
-    A hash key of the repo for internal storage
-    """
     storage_hash = None
-
     """
-    Boolean. True if enterprise, False if not.
+    A hash key of the repo for internal storage
```
matt-codecov, line 49: is this comment in the wrong place?

matt-codecov (approved):

services/archive.py
Swatinem: I guess its the weird way that python doc comments work, as they appear (inside, or) *after* the item they are documenting

PR #693, opened 2024-09-06
fix: deduplicate component measurements

joseph-sentry:
we were getting measurements with duplicate "index_elements" values in the measurements list which was causing errors because that's not allowed, so we have to deduplicate them. I added logs for when we get duplicates.

Fixes: https://github.com/codecov/internal-issues/issues/717

### DIFF
--- a/services/tests/test_timeseries.py
+++ b/services/tests/test_timeseries.py
@@ -422,6 +422,12 @@ class TestTimeseriesService(object):
                         "flag_regexes": [],
                         "paths": [r"folder/*"],
                     },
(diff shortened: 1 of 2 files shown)


giovanni-guidini (approved):
nice, but this assumes all duplicated measurements originate in the same task. Is it possible to have 2 different tasks producing the same measurements?

joseph-sentry:
the issue is only when in the sql statement you insert a row, then a different row in the insert statement has a conflict with the previous one, it's fine if 2 different tasks have 2 duplicate measurements (i would expect that to be the case given the on conflict do update)

PR #695, opened 2024-09-06
Fix commit parent selection when there are multiple parents

michelletran-codecov:
We were not actually performing the branch filter on the git parents. This fixes it and also refactors the code a bit to avoid duplication.

### Legal Boilerplate

### DIFF
--- a/services/repository.py
+++ b/services/repository.py
@@ -168,8 +168,7 @@ async def fetch_appropriate_parent_for_commit(
         else:
             possible_commit = possible_commit_query.filter(
                 Commit.branch == commit.branch,
-            )
-            possible_commit = possible_commit_query.first()
+            ).first()
         if possible_commit:
             return possible_commit.commitid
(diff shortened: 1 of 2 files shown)


services/repository.py
```diff
@@ -213,6 +200,16 @@ async def fetch_appropriate_parent_for_commit(
     return closest_parent_without_message
 
 
+async def possibly_filter_out_branch(commit: Commit, query: Query) -> Commit | None:
```
matt-codecov, line 203: is this actually async?

matt-codecov (commented):
the three cases you replaced with your new helper function all appear to be logically equivalent to the helper function. am i missing the fix?

services/repository.py
michelletran-codecov: Ah, it doesn't have to be! Removed it.

matt-codecov (approved):

PR #698, opened 2024-09-10
feat: collect extra shas for gitlab notif

giovanni-guidini:
depends on https://github.com/codecov/shared/pull/354

These changes collect extra SHAs that we need to (potentially) notify for GitLab users
that use "merge request result" pipelines and the "merge train" feature.

These changes only _collect_ the extra SHAs. The actual notification will happen later.

### DIFF
--- a/tasks/notify.py
+++ b/tasks/notify.py
@@ -585,7 +585,7 @@ class NotifyTask(BaseCodecovTask, name=notify_task_name):
     @sentry_sdk.trace
     async def get_gitlab_extra_shas_to_notify(
         self, commit: Commit, repository_service: TorngitBaseAdapter
-    ) -> set[str] | None:
+    ) -> set[str]:
         """ "
         Fetches extra commit SHAs we should send statuses too for GitLab.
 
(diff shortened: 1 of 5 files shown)


tasks/notify.py
```diff
@@ -529,6 +539,9 @@ def save_patch_totals(self, comparison: ComparisonProxy) -> None:
             return
         head_commit = comparison.head.commit
         db_session = head_commit.get_db_session()
+        if db_session is None:
```
Swatinem, line 542: under which circumstances does this ever happen?

Swatinem (approved):

tasks/notify.py
giovanni-guidini: I'm not sure. I don't think it should happen, but my type hinter tells me this the result of `Commit.get_db_session()` is `Any | None`. 

Also we have issues [WORKER-N5V](https://codecov.sentry.io/issues/5448418739/?project=4165036&query=%27NoneType%27&referrer=issue-stream&statsPeriod=7d&stream_index=3&utc=true) and [WORKER-NH1](https://codecov.sentry.io/issues/5620843269/?project=4165036&query=%27NoneType%27&referrer=issue-stream&statsPeriod=7d&stream_index=4&utc=true) - not in this task exactly - that show `dbsession` being `None`

PR #699, opened 2024-09-10
feat: upsert daily test rollup when we process a test instance

joseph-sentry:
daily test rollups are unique by repo, test, branch and date, the goal is to have a summary of all instances that ran for a test on a given day. We're basically frontloading part of the aggregation of the test instances

depends on: https://github.com/codecov/shared/pull/352


tasks/test_results_processor.py
```diff
@@ -143,12 +146,105 @@ def _bulk_write_tests_to_db(
                     )
                 )
 
+                def update_daily_total():
+                    daily_totals[test_id]["last_duration_seconds"] = duration_seconds
```
adrian-codecov, line 150: Nit: change these to daily_totals.get(test_id, {}).get("last_duration_seconds") to prevent null errors

adrian-codecov (approved):
Approving, with caveat of considering the caveats of having foreign keys in the partitioned models. Preferably merge after hours w/ infra to anticipate migration hiccups

PR #703, opened 2024-09-11
add catchall exception handling to build_report_from_raw_content()

matt-codecov:
https://github.com/codecov/engineering-team/issues/1520 (or starts to, anyway)

after this deploys we should watch GCP logs to see what types of unrecognized exceptions are thrown and add catch cases for them + new error codes. then we can display UIs more consistently/descriptively in the UI.

- shared PR https://github.com/codecov/shared/pull/355
- api PR https://github.com/codecov/codecov-api/pull/820
- gazebo PR https://github.com/codecov/gazebo/pull/3193


Swatinem (commented):
this makes sense, and I think you can also clean up some error handling in the caller function as well, as that has some try/except logic as well.

services/report/__init__.py
```diff
@@ -986,6 +986,21 @@ def build_report_from_raw_content(
...
+        except Exception as e:
+            log.exception(
+                "Unknown error when fetching raw report from storage",
+                extra=dict(
+                    repoid=commit.repoid,
+                    commit=commit.commitid,
+                    archive_path=archive_url,
+                ),
+            )
+            result.error = ProcessingError(
```
adrian-codecov, line 998: Maybe not the scope of this PR, but do we intend to eventually set the upload's state also to some error? A state seems to also be "error", while we also have specific errors tied to an upload, so not sure what's the expected result here

matt-codecov: @adrian-codecov returning a ProcessingError is how the state gets uploaded correctly ([here](https://github.com/codecov/worker/blob/7ebdf9f9140a619b71347c4597d6d68079cb2850/services/report/__init__.py#L1040)). this PR basically plugs holes where we failed to return a ProcessingError when an error occurred

adrian-codecov: That makes sense, ty for pointing it 👌

adrian-codecov (approved):

Swatinem (approved):

Swatinem (approved):

PR #704, opened 2024-09-12
Avoid copying raw upload for parallel verification

Swatinem:
Instead of creating a copy of a raw upload for consumption in the parallel verification task, this will only schedule parallel verification in case the uploads are not being deleted after processing.

That way, the uploads are guaranteed to still be there for verification, so no copy is needed.

### DIFF
--- a/helpers/reports.py
+++ b/helpers/reports.py
@@ -1,7 +1,20 @@
 from collections import namedtuple
 
+from shared.config import get_config
+from shared.yaml import UserYaml
+
+from services.yaml.reader import read_yaml_field
+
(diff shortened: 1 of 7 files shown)


giovanni-guidini (approved):
LGTM.
Just pointing out that processing might change the format of the raw report to a "readable format" (legacy) if it was uploaded as v1 format. I don't think this matters much, but it might affect performance metrics if we are collecting those.

Swatinem:
Indeed. We might end up seeing some more Sentry logs about wrong report format, though I changed those to be compatible with both formats anyway.

PR #705, opened 2024-09-12
Update flaky fail count in the process flakes task

joseph-sentry:
depends on: https://github.com/codecov/worker/pull/699 and https://github.com/codecov/shared/pull/356


tasks/process_flakes.py
```diff
@@ -118,16 +114,22 @@ def generate_flake_dict(repo_id: int) -> FlakeDict:
...
     flake.recent_passes_count += 1
+
     if flake.recent_passes_count == FLAKE_EXPIRY_COUNT:
-        flake.end_date = dt.datetime.now(tz=dt.UTC)
+        flake.end_date = test_instance.created_at
+
     flake.save()
 
 
 def upsert_failed_flake(
```
ajay-sentry, line 129: nit: we should try and be persistent about ordering of params across functions belonging to the same module

test_instance,flake,repo_id vs. test_instance,repo_id,flake

And Ideally the most common params occur first

tasks/process_flakes.py
```diff
@@ -136,10 +138,32 @@ def upsert_failed_flake(
...
-            start_date=dt.datetime.now(dt.UTC),
+            start_date=test_instance.created_at,
             recent_passes_count=0,
         )
         flake.save()
+
+        # retroactively mark newly caught flake as flaky failure in its rollup
+        rollup = DailyTestRollup.objects.filter(
+            repoid=repo_id,
+            date=test_instance.created_at.date(),
```
ajay-sentry, line 149: curious about this .date() fn, what type does that cast to?

Saw it in the log as well and wondering why we'd want to cast it there vs just returning test_instance.created_at

tasks/test_results_processor.py
```diff
@@ -68,8 +74,21 @@ def run_impl(
         testrun_dict_list = []
         upload_list = []
 
+        f = db_session.query(Flake).all()
```
ajay-sentry, line 77: technically this is flakes

tasks/test_results_processor.py
```diff
@@ -68,8 +74,21 @@ def run_impl(
...
+
+        flakes = (
```
ajay-sentry, line 79: this is repo flakes

tasks/test_results_processor.py
```diff
@@ -107,167 +126,147 @@ def _bulk_write_tests_to_db(
...
-                    daily_totals[test_id]["avg_duration_seconds"] = (
-                        daily_totals[test_id]["avg_duration_seconds"]
-                        * (
-                            daily_totals[test_id]["pass_count"]
-                            + daily_totals[test_id]["fail_count"]
-                        )
-                        + duration_seconds
-                    ) / (
+            def update_daily_total():
+                daily_totals[test_id]["last_duration_seconds"] = duration_seconds
```
ajay-sentry, line 168: Can we add a comment here for what the formula is? It's a bit hard to parse otherwise

tasks/test_results_processor.py
```diff
@@ -107,167 +126,147 @@ def _bulk_write_tests_to_db(
...
+                    "repoid": repoid,
+                    "last_duration_seconds": duration_seconds,
+                    "avg_duration_seconds": duration_seconds,
+                    "pass_count": 1 if outcome == str(Outcome.Pass) else 0,
+                    "fail_count": 1
+                    if outcome == str(Outcome.Failure) or outcome == str(Outcome.Error)
+                    else 0,
+                    "skip_count": 1 if outcome == str(Outcome.Skip) else 0,
+                    "flaky_fail_count": 1
+                    if test_id in flaky_test_set and outcome == str(Outcome.Failure)
```
ajay-sentry, line 201: would an error be considered a flake too?

tasks/test_results_processor.py
```diff
@@ -107,167 +126,147 @@ def _bulk_write_tests_to_db(
...
+
+        # Save Tests
+        insert_on_conflict_do_nothing = (
+            insert(Test.__table__).values(test_data).on_conflict_do_nothing()
+        )
+        db_session.execute(insert_on_conflict_do_nothing)
+        db_session.flush()
+
+        # Upsert Daily Test Totals
+        rollup_table = DailyTestRollup.__table__
```
ajay-sentry, line 227: I may need you to run me through this on Monday 😅 

Particularly why we need to do the calculations again on conflict; I figured they'd be computed in daily_totals

tasks/process_flakes.py
joseph-sentry: date is a PG `date` type and created_at is a PG `timestamp with TZ` type, in python the equivalents are `datetime.date` and `datetime.datetime`. The `date()` method is getting the date component of the `datetime.datetime`

tasks/test_results_processor.py
joseph-sentry: oops, this is leftover from me debugging i think

tasks/test_results_processor.py
joseph-sentry: yeah for flaky test detection we consider an error a failure, me omitting Error here is a mistake

tasks/test_results_processor.py
joseph-sentry: ah the reason we do them is because the dictionary is just an intermediate holding container, if we came across two  test instances that map to the same rollup, we can't rely on the database to do the aggregating for us because there's a restriction that an `insert on conflict do update` can't have two entries that would insert/update the same row. So in the dictionary above we're duplicating this logic

tasks/process_flakes.py
ajay-sentry: got it! Thanks for the clarification

tasks/tests/unit/test_process_flakes.py
```diff
@@ -51,10 +51,51 @@ def add_test_instance(self, c, outcome=TestInstance.Outcome.PASS.value):
...
+                "commits_where_fail": [],
+            },
+        )
+
+        match outcome:
+            case TestInstance.Outcome.PASS.value:
+                rollup.pass_count += 1
+            case TestInstance.Outcome.SKIP.value:
+                rollup.skip_count += 1
+            case _:
```
ajay-sentry, line 78: is this identical to "default" in other languages?

joseph-sentry: yep

tasks/tests/unit/test_process_flakes.py
```diff
@@ -400,3 +440,60 @@ def test_it_creates_flakes_expires(transactional_db):
...
+    rs.add_test_instance(c1, outcome=TestInstance.Outcome.FAILURE.value)
+
+    ProcessFlakesTask().run_impl(
+        None,
+        repo_id=rs.repo.repoid,
+        commit_id_list=[c1.commitid],
+        branch=rs.repo.branch,
+    )
+
+    traveller.stop()
```
ajay-sentry, line 462: why is it checking for code coverage in a test file lol

joseph-sentry: i actually don't know

ajay-sentry (approved):
looks good, nice work man!

PR #707, opened 2024-09-13
perf: use lpop with count and don't fetch upload in Upload task

joseph-sentry:
we don't need to fetch the report upload because all we're doing is setting the upload_pk to upload_id in the normalized args dict

we also can batch our redis lpops into groups of 50 so we don't have to make as many calls to redis


Swatinem (approved):
looks great!

PR #712, opened 2024-09-16
feat: notify status to multiple shas (gitlab)

giovanni-guidini:
related to https://github.com/codecov/engineering-team/issues/2076

Now that we are collecting different SHAs that we potentially need to notify for Gitlab,
the only missing piece was actually doing the notification in multiple SHAs.

Some notes:
* The `checks` notifier is exclusive for GitHub, so no need to change that
* "notifying to multiple shas" is done in the most transparent way possible
* I created `_send_notification_with_metrics` just to wrap the timer, but it would probably be better to
add it in the Torngit adapter directly
* We don't need to do this for the PR comment, just the status

### DIFF
--- a/services/notification/notifiers/status/base.py
+++ b/services/notification/notifiers/status/base.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 from typing import Dict
 
@@ -16,9 +17,7 @@ from services.notification.notifiers.base import (
     Comparison,
     NotificationResult,
 )
-from services.repository import (
-    get_repo_provider_service_for_specific_commit,
-)
+from services.repository import get_repo_provider_service_for_specific_commit
 from services.urls import get_commit_url, get_pull_url
 from services.yaml import read_yaml_field
 from services.yaml.reader import get_paths_from_flags
@@ -160,7 +159,7 @@ class StatusNotifier(AbstractBaseNotifier):
         )
         return selected_installation_id
 
-    async def notify(self, comparison: Comparison):
+    async def notify(self, comparison: ComparisonProxy):
         payload = None
         if not self.can_we_set_this_status(comparison):
             return NotificationResult(
@@ -273,7 +272,7 @@ class StatusNotifier(AbstractBaseNotifier):
         return f"codecov/{self.context}{status_piece}"
 
     async def maybe_send_notification(
-        self, comparison: Comparison, payload: dict
+        self, comparison: ComparisonProxy, payload: dict
     ) -> NotificationResult:
         base_commit = (
             comparison.project_coverage_base.commit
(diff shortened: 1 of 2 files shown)


services/notification/notifiers/status/base.py
```diff
@@ -273,7 +272,7 @@ def get_status_external_name(self) -> str:
         return f"codecov/{self.context}{status_piece}"
 
     async def maybe_send_notification(
-        self, comparison: Comparison, payload: dict
+        self, comparison: ComparisonProxy, payload: dict
```
adrian-codecov, line 275: Has this always been Proxy and we never realized?

services/notification/notifiers/status/base.py
```diff
@@ -336,26 +357,37 @@ async def send_notification(self, comparison: Comparison, payload):
...
-                        commit=head.commitid,
-                        status=state,
-                        context=title,
-                        coverage=(
-                            float(head_report.totals.coverage) if head_report else 0
-                        ),
-                        description=message,
-                        url=url,
-                    )
+                all_results = await asyncio.gather(*all_notify_tasks)
```
adrian-codecov, line 382: Is there any overhead of doing gather vs await if all_notify_tasks has more often than not length = 1?

services/notification/notifiers/status/base.py
```diff
@@ -336,26 +357,37 @@ async def send_notification(self, comparison: Comparison, payload):
                 "state": state,
                 "message": message,
             }
+
+            all_shas_to_notify = [head_commit_sha] + list(
+                comparison.context.gitlab_extra_shas or set()
```
adrian-codecov, line 362: Optional: leave a comment explaining why we have to do this, also optional since you already left a comment in the gitlab_extra_shas definition

services/notification/notifiers/status/base.py
```diff
@@ -336,26 +357,37 @@ async def send_notification(self, comparison: Comparison, payload):
...
-                        status=state,
-                        context=title,
-                        coverage=(
-                            float(head_report.totals.coverage) if head_report else 0
-                        ),
-                        description=message,
-                        url=url,
-                    )
+                all_results = await asyncio.gather(*all_notify_tasks)
+                res = all_results[0]
```
adrian-codecov, line 383: Would we only care of the 1st result here? I guess cause the second+ would be the extra gitlab ones?

adrian-codecov (commented):
Left some questions 👀

services/notification/notifiers/status/base.py
giovanni-guidini: I don't know. But this seems to be the best way to do it.
It probably has some overhead, but I doubt it's relevant

services/notification/notifiers/status/base.py
giovanni-guidini: I was hoping that having a comment both in the `gitlab_extra_shas` definition and the function that calculates those extra shas would suffice.

services/notification/notifiers/status/base.py
giovanni-guidini: Exactly. The 1st result matches the notification we were doing before. All others should be as transparent as possible, I suppose

adrian-codecov (approved):

PR #713, opened 2024-09-17
Make `session.id == upload.id`

Swatinem:
Instead of trying very hard to allocate contiguous `session.id`s, just use the `upload.id` as the `session.id`.

The `upload.id` is guaranteed to be unique, so there should be no collisions. It is also very large, so there should not be any collisions with existing sessions from merged/carry-forwarded reports.


services/report/raw_upload_processor.py
```diff
@@ -22,23 +22,9 @@
...
-        SpecialLabelsEnum.CODECOV_ALL_LABELS_PLACEHOLDER.corresponding_index: SpecialLabelsEnum.CODECOV_ALL_LABELS_PLACEHOLDER.corresponding_label
-    }
-
-
-def invert_pattern(string: str) -> str:
-    if string.startswith("!"):
-        return string[1:]
-    else:
-        return "!%s" % string
+DEFAULT_LABEL_INDEX = {
```
giovanni-guidini, line 25: Is it fine to use a shared reference as the label index?

giovanni-guidini (approved):
LGTM, left one comment.

services/report/raw_upload_processor.py
Swatinem: actually you are right, I fixed this to default the `labels_index` to a copy of that global.
I might have been confused by the comment, plus this being used for a comparison later on which in python is structural equality for dicts.

PR #715, opened 2024-09-17
feat: add migrate.sh to worker

joseph-sentry:


adrian-codecov (approved):
As a future better way, we should add this script to Shared and use this for worker + api

migrate.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+# Command ran by k8s to run migrations for api
```
adrian-codecov, line 3: Worker instead of api

PR #723, opened 2024-09-18
bundle analysis: remove unnecessary save measurements

JerrySentry:
closes https://github.com/codecov/engineering-team/issues/1875

No need to save every bundle of a commit, just save the processed bundle instead, otherwise it is inserting duplicate measurements to timeseries

### Legal Boilerplate

### DIFF
--- a/services/bundle_analysis/report.py
+++ b/services/bundle_analysis/report.py
@@ -354,7 +354,9 @@ class BundleAnalysisReportService(BaseReportService):
         db_session.flush()
 
     @sentry_sdk.trace
-    def save_measurements(self, commit: Commit, upload: Upload) -> ProcessingResult:
+    def save_measurements(
+        self, commit: Commit, upload: Upload, bundle_name: str
+    ) -> ProcessingResult:
         """
         Save timeseries measurements for this bundle analysis report
         """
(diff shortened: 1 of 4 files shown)


services/bundle_analysis/report.py
```diff
@@ -372,50 +374,51 @@ def save_measurements(self, commit: Commit, upload: Upload) -> ProcessingResult:
             ]
 
             db_session = commit.get_db_session()
```
JerrySentry, line 376: The diff is basically from:
```
for bundle_report in bundle_analysis_report.bundle_reports():
    <all the saving code>
```
to
```
bundle_report = bundle_analysis_report.bundle_report(bundle_name)
<all the saving code>
```

services/bundle_analysis/report.py
```diff
@@ -369,50 +371,51 @@ def save_measurements(self, commit: Commit, upload: Upload) -> ProcessingResult:
...
-                # For overall bundle size
-                if MeasurementName.bundle_analysis_report_size.value in dataset_names:
-                    self._save_to_timeseries(
-                        db_session,
-                        commit,
-                        MeasurementName.bundle_analysis_report_size.value,
-                        bundle_report.name,
-                        bundle_report.total_size(),
-                    )
+            bundle_report = bundle_analysis_report.bundle_report(bundle_name)
```
giovanni-guidini, line 374: it seems that this can be `None`. We could early-return here if that's the case

services/bundle_analysis/report.py
giovanni-guidini: Thank you for this comment, most helpful

giovanni-guidini (approved):
LGTM, left a comment

PR #725, opened 2024-09-19
Better detect corrupted Go report lines

Swatinem:
It appears that Go coverage reports are very susceptible to mangled lines. This makes sure we encapsulate all of the parsing for a coverage line within a try/except block, or skip over other invalid lines, instead of raising unexpected errors.

fixes [WORKER-MB9](https://codecov.sentry.io/issues/5165160706/), fixes [WORKER-P4S](https://codecov.sentry.io/issues/5787850847/), fixes [WORKER-P8D](https://codecov.sentry.io/issues/5826394496/), fixes [WORKER-P7Z](https://codecov.sentry.io/issues/5821076186/), fixes [WORKER-PAJ](https://codecov.sentry.io/issues/5860678163/), fixes [WORKER-PAH](https://codecov.sentry.io/issues/5860453353/), fixes [WORKER-P56](https://codecov.sentry.io/issues/5792037589/), fixes [WORKER-P7W](https://codecov.sentry.io/issues/5820563536/), fixes [WORKER-PBC](https://codecov.sentry.io/issues/5868077255/)

### DIFF
--- a/services/report/languages/go.py
+++ b/services/report/languages/go.py
@@ -7,8 +7,8 @@ from shared.utils import merge
 from shared.utils.merge import LineType, line_type, partials_to_line
 
 from helpers.exceptions import CorruptRawReportError
-from services.path_fixer import PathFixer
 from services.report.languages.base import BaseLanguageProcessor
+from services.report.languages.helpers import Region, SourceLocation
 from services.report.report_builder import ReportBuilderSession
 
 
@@ -30,11 +30,13 @@ def from_txt(string: bytes, report_builder_session: ReportBuilderSession) -> Non
     )
 
     # Process the bytes from uploaded report to intermediary representation
-    # files: {new_name: <lines defaultdict(list)>}
-    files = process_bytes_into_files(string, report_builder_session.path_fixer)
-    # create a file
+    files = process_bytes_into_files(string)
+
     for filename, lines in files.items():
-        _file = report_builder_session.create_coverage_file(filename, do_fix_path=False)
+        _file = report_builder_session.create_coverage_file(filename)
+        if _file is None:
+            continue
+
         for ln, partials in lines.items():
             best_in_partials = max(map(lambda p: p[2], partials))
             partials = combine_partials(partials)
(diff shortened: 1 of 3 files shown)


services/report/languages/go.py
```diff
@@ -80,71 +78,66 @@ def process_bytes_into_files(
...
-        line_start, line_end = columns.split(",", 1)
-        line_start, sc = list(map(int, line_start.split(".", 1)))
+        filename = split[0]
         try:
-            line_end, ec = list(map(int, line_end.split(".", 1)))
+            region = parse_coverage(split[1])
         except ValueError:
+            # FIXME: do we actually want to raise an error here?
+            # Why not just skip over invalid lines, as the coverage file likely
+            # contains other valid lines we can use.
```
Swatinem, line 101: this is a valid question, and I believe we should indeed "silently" ignore these corrupt lines, though more discussion would be needed here.

giovanni-guidini: I think a good compromise for now would be to emit logs that lines have been ignored.

Particularly if coverage starts to fluctuate and questions arise as to "why is Codecov saying my coverage has go up / down?" we can at least point to "your report couldn't be fully parsed"

Another thing we maybe should verify is if the lines are indeed corrupted or if the standard changed a little bit over time and we didn't update our processors accordingly

I don't think being totally silent is the best approach, but I think we should be subtle. As you said, "silently" haha, I agree with that.

giovanni-guidini (approved):

services/report/languages/go.py
Swatinem: Indeed, it is hard to tell "complete garbage" from "format has changed".
I’m not quite sure whether using logs (either using `logging` or Sentry) would help us here.
We are dealing with "untrusted user data" in the end, so I also don’t want to be DOS-d by fabricated data.

So for in all cases I have seen, and which have generated internal errors so far, it was "complete garbage".

I think changing the status reporting to:
- File format not recognized at all
- File format fully parsed without errors
- File format parsed, with errors

Having metrics for these cases would also be a good thing as well.

PR #728, opened 2024-09-19
fix: add more information to detected flakes logs

joseph-sentry:


tasks/test_results_finisher.py
```diff
@@ -357,6 +357,7 @@ def process_impl_within_lock(
                 "Detected failure on test that has been identified as flaky",
                 extra=dict(
                     success=success,
+                    notifier_result=notifier_result.value,
```
michelletran-codecov, line 360: Would using the `.name` be more readable?

joseph-sentry: the definition of the notifier result looks like: 
```
class NotifierResult(Enum):
    COMMENT_POSTED = "comment_posted"
    TORNGIT_ERROR = "torngit_error"
    NO_PULL = "no_pull"
```
so i think either name or value should be fine

michelletran-codecov (approved):

PR #729, opened 2024-09-19
Run django db migrations for tests

michelletran-codecov:
This ensures that the table that the models run on are the Django models, rather than the slightly less accurate SQLAlchemy models. 

The first 3 commits are related to getting the SQLAlchemy tables created with Django migrations. I've decided to keep the databases for each test separate (because Django transactions tests won't work properly with a persistently open SQLAlchemy DB connection). 

Some of the common fixes to the tests:
* add timezones
* set `created_at` timestamp for `owner` table to be before the patch centric YAML change (the default DB behaviour is to set that time to now, which breaks old legacy tests that check YAML)
* fix issues with missing not-null fields or unique references

### Legal Boilerplate

### DIFF
--- a/database/tests/factories/core.py
+++ b/database/tests/factories/core.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timezone
 from hashlib import sha1
 from uuid import uuid4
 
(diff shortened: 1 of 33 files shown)


database/tests/factories/core.py
```diff
@@ -71,9 +71,11 @@ class Meta:
 
     owner = factory.SubFactory(OwnerFactory)
     bot = None
-    updatestamp = factory.LazyAttribute(lambda o: datetime.now())
+    updatestamp = factory.LazyAttribute(lambda o: datetime.now(tz=timezone.utc))
     languages = []
-    languages_last_updated = factory.LazyAttribute(lambda o: datetime.now())
+    languages_last_updated = factory.LazyAttribute(
+        lambda o: datetime.now(tz=timezone.utc)
```
Swatinem, line 77: Sentry uses `django.utils.timezone.now()` for these. Would that make sense here as well?

Swatinem (approved):

database/tests/factories/core.py
michelletran-codecov: Hmm, these are factories to create the SQLAlchemy models, so I don't think using a Django Utils is appropriate.

PR #730, opened 2024-09-20
WIP: Make `ComparisonProxy` sync

Swatinem:


tasks/status_set_error.py
```diff
@@ -59,18 +59,9 @@ def run_impl(self, db_session, repoid, commitid, *, message=None, **kwargs):
                             message or "Coverage not measured fully because CI failed"
                         )
                         if context in statuses:
-                            # async_to_sync has its own "context" argument so we
```
giovanni-guidini, line 62: Is this comment wrong? Or simply by passing the argument as `arg` (instead of `kwarg`) we sidestep `async_to_sync` having a context?

giovanni-guidini (commented):
🎉 LGTM, very nice

giovanni-guidini (approved):

tasks/status_set_error.py
Swatinem: > simply by passing the argument as arg (instead of kwarg) we sidestep async_to_sync having a context?

yes, exactly that. the problem is that it internally spreads the `kwargs` into a call that already has a `context` parameter.

I stumbled across this bug in `async_to_sync` myself in a different part of this PR, and fixed that the same way.
It took me a while to figure that out btw :-(

PR #734, opened 2024-09-23
feat: add backfill one off scripts

joseph-sentry:
these are one off python scripts to be run directly that will modify the database

the first backfills the daily test rollups starting on a given day, and ending on a given date, starting for a given repo (it will backfill in repo id order)

the second backfills the TestFlagBridge objects for all repos with the option to focus on a given repo, specified by repoid


one_off_scripts/backfill_daily_test_rollups.py
```diff
@@ -0,0 +1,191 @@
...
+            skip_count=obj.skip_count,
+            flaky_fail_count=obj.flaky_fail_count,
+            commits_where_fail=list(obj.commits_where_fail),
+            latest_run=obj.latest_run,
+            last_duration_seconds=obj.last_duration_seconds,
+            avg_duration_seconds=obj.sum_duration_seconds
+            / (obj.pass_count + obj.fail_count),
+        )
+
+        rollup.save()
```
Swatinem, line 111: django has a [`bulk_create`](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#bulk-create) method as well, though I recommend to always define a reasonable chunk size.

one_off_scripts/backfill_daily_test_rollups.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def run_impl(
```
Swatinem, line 114: please give this a more descriptive name :-)

one_off_scripts/backfill_daily_test_rollups.py
```diff
@@ -0,0 +1,191 @@
...
+    start_repoid: int | None = None,
+    start_date: str | None = None,  # default is 2024-07-16
+    end_date: str | None = None,  # default is 2024-09-17
+) -> dict[str, bool]:
+    log.info(
+        f"Updating test instances {start_repoid} {start_date} {end_date}",
+        extra=dict(start_repoid=start_repoid, start_date=start_date, end_date=end_date),
+    )
+    test_analytics_repos = get_test_analytics_repos(start_repoid)
```
Swatinem, line 123: you should probably also iterate over these in chunks, because you are also logging the whole list.
unless you know for certain there is only a small number of these?

one_off_scripts/backfill_test_flag_bridges.py
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        for test in tests:
+            TestFlagBridge.objects.filter(test == test).delete()
+
+            first_test_instance = TestInstance.objects.filter(
+                test_id=test.id
+            ).select_related("upload")[:1][0]
+
+            if first_test_instance is None:
```
Swatinem, line 36: I’m not 100% sure how this works with using the slicing syntax to define a limit, whether you actually get a list of appropriate length with `None` filled in?

I think this might just throw because you are indexing out of bounds if no such object exists?

one_off_scripts/backfill_test_flag_bridges.py
```diff
@@ -0,0 +1,49 @@
...
+                continue
+
+            flag_names = first_test_instance.upload.flag_names
+
+            for flag_name in flag_names:
+                new_bridge = TestFlagBridge(test=test, flag=flags[flag_name])
+                new_bridge.save()
```
Swatinem, line 43: again, `bulk_create` should be more appropriate here.

one_off_scripts/__init__.py
```diff
@@ -0,0 +1,6 @@
+import os
+
+import django
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "django_scaffold.settings")
+django.setup()
```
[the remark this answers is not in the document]
joseph-sentry: These scripts are meant to be run separately from the actual application running on a pod so we'd invoke it directly using `python one_off_script.py` and that means we have to setup django separately

one_off_scripts/backfill_daily_test_rollups.py
joseph-sentry: Yeah there aren't many `Repository` objects that would be returned from this function

one_off_scripts/backfill_test_flag_bridges.py
```diff
@@ -0,0 +1,54 @@
...
+
+        bridges_to_create = []
+        for test in tests:
+            TestFlagBridge.objects.filter(test=test).delete()
+
+            try:
+                first_test_instance = (
+                    TestInstance.objects.filter(test_id=test.id)
+                    .select_related("upload")
+                    .first()
```
Swatinem, line 38: my comment might have been confusing. I believe `first` returns `None` if the element does not exist instead of the `IndexError`.

one_off_scripts/backfill_daily_test_rollups.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def get_test_analytics_repos(start_repoid):
+    # get all repos that have test_analytics_enabled == True
+    test_analytics_repos = Repository.objects.filter(
+        test_analytics_enabled=True
+    ).order_by("repoid")
+
+    if start_repoid is not None:
+        test_analytics_repos = test_analytics_repos.filter(repoid__gte=start_repoid)
```
nora-codecov, line 37: smart

nora-codecov (approved):

PR #737, opened 2024-09-24
Remove ATS setup

Swatinem:
ATS (automatic test selection) is not functional in this repo (anymore).

It was previously used using the `onlysomeflags` flag, which was not used for uploads since quite some time, but which seems to still be carry-forwarded in reports to this day.

As a driveby change, this also cleans up configuration for `mutmut` which has been disabled/removed for a while, as well as other codecov beta features which have not made it out of beta.


adrian-codecov (approved):
Are we for sure not using ATS anymore? Please confirm if you haven't, and if so ✅

Swatinem:
The `run-ats` GHA workflow is only referenced in `api`, and there it is also commented out: https://github.com/search?q=org%3Acodecov%20run-ats&type=code

So I’m quite confident we are good here. Also judging by the database entries, the last one of these carried-forward reports is from february this year.

PR #739, opened 2024-09-25
Create an E2E test for Carryforwarding

Swatinem:
This adds another E2E-style test which runs the "full" upload process from Redis through the various Upload Tasks, using two consecutive commits that will carry-forward a couple of sessions from the first to the second commit.


giovanni-guidini (approved):
nice 🎉

PR #740, opened 2024-09-25
Create TestFlagBridge during test results processing

joseph-sentry:
we want to connect test objects to the RepositoryFlag objects


tasks/test_results_processor.py
```diff
@@ -150,6 +182,12 @@ def _bulk_write_tests_to_db(
                 )
             )
 
+            test_flag_bridge_data += [
+                {"test_id": test_id, "repoid": repoid, "flag_id": repo_flags[flag]}
```
ajay-sentry, line 186: Because test_id is not dependent on flag in flags I think we can move that outside of the list comprehension and reduce some cycles spent evaluating that for every flag right?

joseph-sentry: at some point we'd have to turn it into a list of dicts with each value so sqlalchemy can insert it

ajay-sentry (approved):

PR #742, opened 2024-09-25
feat: update test_results_parser version and start computing names

joseph-sentry:
This commit updates the version of the test_results_parser library
that we are using:
- no more support for parsing vitest.json or pytest-reportlog files
- parse_junit_xml now returns a ParsingInfo class, the type signature
  looks something like:

    class ParsingInfo:
        testruns: list[Testrun]
        framework: framework

  where Framework is:

    class Framework(Enum):
        Pytest
        Jest
        Vitest
        PHPUnit

- Testrun also now contains the filename field which is an optional
  string and a classname field

We've also started computing the name of the test which just means
depending on the framework we've detected we transform the raw name
and classname from the JUnit file to something more readable and if
that computed name is available it's what we should show in the PR
comment


tasks/test_results_processor.py
```diff
@@ -376,35 +443,23 @@ def process_individual_arg(self, upload: Upload, repository) -> List[Testrun]:
...
-            ) from e
-
         try:
             file_content = file_bytes.read()
             with metrics.timer("test_results.processor.file_parsing"):
-                res = parsing_function(file_content)
+                res = parse_junit_xml(file_content)
+                print("BBBBBBBBBBBBBBBBBBBBBBBBBBBBB")
         except ParserError as e:
+            print("AAAAAAAAAAAAAAAAAAAAAAAAAAAAA")
```
michelletran-codecov: Remove print statements

tasks/tests/samples/sample_test_network.txt
```diff
@@ -0,0 +1 @@
+{"network_files": ["api/temp/calculator/test_calculator.py"], "test_results_files": [{"filename": "codecov-demo/temp.junit.xml", "format": "base64+compressed", "data": "eJy1VMluwjAQvfMVI1dCoBbHZiklJEFVS4V66Kkqx8okBqw6i2KHwt/XWSChnCrRXDLjefNm8Uuc2T6UsOOpEnHkIooJAh75cSCijYsyve49oJnnaK60yoR5NWyIWMhdlBzyE5OWpnGqXGQY1kzILOXGoQjUl0gSHhSBItdFQ2OJPJdgMuqXjtIsTFzUJ/1Bj9IeuX+n1KZjmwwxoZSMDWwbK13W/HhZvC1fn5eLCZaxzyQq2/KZ4uBLplQJY4nAmocJNhA/k0zHKc5xn7WPqimKYxYEjc6Iad66DrHKVjplvv4f9jCTWiTy0GQnV2MPxE4E/Lxzz6muGMzFKbbJaZXiqQajIHBdIHjUvqFkCrcA31tugEUA2lJP11nkayM3eKobKIsA00D2lAz9CV9NSHujpx16B/3uievI9mceU/sChryAr6ExZKdrtwpw+VQjXeSVPVVjtubn6HoBp8iVlnDGd9VFtFpGFFYuCqsWgfVLFDg52ANiw2Mxpyk3zz94x6qU4DnWUW2VWfwkmrbyfgBbcXMH", "labels": ""}], "metadata": {}}
```
michelletran-codecov, line 1: nit: Can we format these json better? This will make it easier to read and debug errors.

tasks/test_results_processor.py
```diff
@@ -360,14 +425,16 @@ def process_individual_arg(self, upload: Upload, repository) -> List[Testrun]:
         payload_bytes = archive_service.read_file(upload.storage_path)
         data = json.loads(payload_bytes)
 
-        testrun_list = []
+        parsing_results: list[ParsingInfo] = []
+
+        # TODO: this is bad
+        self.network = data.get("network_files")
```
michelletran-codecov, line 432: Can you add more context into why this is bad? I see that `network_files` is supposed to be in the uploaded test file, but I don't have much context into why getting this here is a bad thing?

michelletran-codecov (commented):
Just a few comments

tasks/test_results_processor.py
joseph-sentry: ah, this was a note for myself to revisit, the alternative to this is to return the network from the function, which i think makes more sense, but i wanted to quickly make sure it worked so i just set it as an attribute on the task itself, will fix this with the next commit

tasks/test_results_processor.py
```diff
@@ -76,9 +76,55 @@ def get_repo_flags(
...
+        framework: Framework,
+        raw_classname: str,
+        raw_name: str,
+        filename: str | None,
+        network: list[str] | None,
+    ) -> str:
+        match framework:
+            case Framework.Jest:
+                name = raw_name
+                return name
```
michelletran-codecov, line 105: simplification suggestion:
```suggestion
                return raw_name
```

tasks/test_results_processor.py
```diff
@@ -149,125 +195,149 @@ def _bulk_write_tests_to_db(
...
+        network: list[str] | None,
         flaky_test_set: set[str],
         flags: list[str],
     ):
-        test_data = []
+        test_data = {}
         test_instance_data = []
         test_flag_bridge_data = []
-        daily_totals = dict()
+        daily_totals: dict[str, dict[str, str | int | list[str]]] = dict()
```
michelletran-codecov, line 206: Not something to do in this PR, but I would consider creating a custom object/type/dataclass to simplify this complex type annotation.

tasks/test_results_processor.py
```diff
@@ -76,9 +76,55 @@ def get_repo_flags(
...
 
 
+@dataclass
+class PytestName:
+    actual_class_name: str
+    test_file_path: str
+
+
+@dataclass
+class argProcessingResult:
```
michelletran-codecov, line 86: Is this supposed to be `ProcessingResult` or `ArgProcessingResult`?

michelletran-codecov (commented):
Generally LGTM! Just a question about one of the class names.

michelletran-codecov (approved):

PR #745, opened 2024-09-27
Clear CFF `Upload`s as they are overwritten

Swatinem:
The `Upload`s in the database should mirror the `Session`s within a stored `Report`.

However, so far this was not the case for carryforwarded sessions/uploads. The `Session`s were removed from the `Report` as they were being overridden by new uploads, but that was not the case for the `Upload` rows in the database which were created from the carry-forwarded `Session`s on initial `Report` carry-forwarding.

This change will now make sure that the `Upload`s in the database match the `Session`s in the `Report`.

### DIFF
--- a/services/report/__init__.py
+++ b/services/report/__init__.py
@@ -12,7 +12,6 @@ import sentry_sdk
 from asgiref.sync import async_to_sync
 from celery.exceptions import SoftTimeLimitExceeded
 from shared.django_apps.reports.models import ReportType
-from shared.metrics import metrics
 from shared.reports.carryforward import generate_carryforward_report
 from shared.reports.editable import EditableReport
 from shared.reports.enums import UploadState, UploadType
@@ -32,6 +31,7 @@ from database.models.reports import (
     ReportLevelTotals,
     RepositoryFlag,
     UploadLevelTotals,
+    uploadflagmembership,
 )
 from helpers.exceptions import (
     OwnerWithoutValidBotError,
@@ -52,7 +52,10 @@ from services.report.prometheus_metrics import (
     RAW_UPLOAD_RAW_REPORT_COUNT,
     RAW_UPLOAD_SIZE,
 )
-from services.report.raw_upload_processor import process_raw_upload
+from services.report.raw_upload_processor import (
+    SessionAdjustmentResult,
+    process_raw_upload,
+)
 from services.repository import get_repo_provider_service
 from services.yaml.reader import get_paths_from_flags, read_yaml_field
 
@@ -72,6 +75,7 @@ class ProcessingResult:
     session: Session
     report: Report | None = None
     error: ProcessingError | None = None
+    session_adjustment: SessionAdjustmentResult | None = None
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


services/report/__init__.py
```diff
@@ -936,17 +941,18 @@ def build_report_from_raw_content(
 
         log.debug("Retrieved report for processing from url %s", archive_url)
         try:
-            with metrics.timer(f"{self.metrics_prefix}.process_report") as t:
```
adrian-codecov, line 939: May be out of the loop, but we're no longer using these metrics?

services/report/__init__.py
```diff
@@ -1239,3 +1255,31 @@ def save_parallel_report_to_archive(
             "chunks_path": chunks_url,
             "files_and_sessions_path": files_and_sessions_url,
         }
+
+
+@sentry_sdk.trace
+def delete_uploads_by_sessionid(upload: Upload, session_ids: list[int]):
```
adrian-codecov, line 1261: 3 things here: 

1) I could see this taking some time because some customers have a gargantuan number of CFFs in their commits, so this might benefit from making a standalone task - unless we're wanting to do this synchronously. In which case, we'd take a bit extra time processing, so something to contemplate/measure in some way if we can
2) How could this react with other sessions being processed in parallel? Not sure if while deleting we could run into errors if two or more processes are trying to dleete the same entry
3) How could we test this in practice? I just want to make sure we aren't silently corrupting data by not acknowledging our reports were using data from cff's we deleted or something like that

services/report/__init__.py
```diff
@@ -1239,3 +1255,31 @@ def save_parallel_report_to_archive(
...
+    """
+    This deletes all the `Upload` records corresponding to the given `session_ids`.
+    """
+    db_session = upload.get_db_session()
+    uploads = (
+        db_session.query(Upload.id_)
+        .filter(Upload.report == upload.report, Upload.order_number.in_(session_ids))
```
adrian-codecov, line 1268: What's order_number? Don't think I've ever run into this one 😅

services/report/__init__.py
Swatinem: I think tracing is a much better alternative to this. Also, I haven’t seen any dashboards actually using these metrics?

services/report/__init__.py
Swatinem: It is the `session_id`, so the incremental ID within the `report_json`.

services/report/__init__.py
Swatinem: 1. not quite sure the overhead of a separate task (both complexity and runtime) is reasonable for that. but you are right that this might be quite slow. However it is limited to only those uploads/sessions which have been overridden by an upload, so that should be rather limited.
2. we are behind the "modify report" (upload processing) lock, so there is no parallelism here. also concurrent deletes shouldn’t be a problem either way, as we filter things by `IN`, which sql just ignores when no matching rows exist.
3. good point, I don’t have a good idea for this yet. Though this change aims to align data in SQL with data inside of the `report_json`. So this is actually fixing a pre-existing "data corruption" if you think about it that way.

with these questions, you have brought up another good point: do we need to care about SQL-level locking here? I don’t think we need to, as the deletes are only locking the effected rows for the duration of the transaction (which I guess spans the whole task?), and those particular rows are only touched behind the upload processing lock.

adrian-codecov: Assuming the deletions take place a short amount of time, I don't think we'd need to worry

adrian-codecov (approved):

adrian-codecov (approved):
When merged, lets keep a close look at how the DB and customers react to the data (e.g. be aware if there's any huge complains from our customers, if we hear anything from support, etc). I'd post a brief message in the eng/support channel to explain the change and have everyone ready

PR #753, opened 2024-10-01
feat: add script for rerunning test result uploads

joseph-sentry:
we had an issue from the 25th to the 28th where the test results processing step was failing

this one off script looks for all commit reports created between those days that is missing a TestResultReportTotals object (which insinutates that the test result processing failed) and tries to run it again using a celery chord

### DIFF
--- a/one_off_script.py
+++ b/one_off_script.py
@@ -5,9 +5,12 @@ import django
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "django_scaffold.settings")
 django.setup()
 
-if __name__ == "__main__":
-    from one_off_scripts.backfill_daily_test_rollups import run_impl
-    from one_off_scripts.backfill_test_flag_bridges import backfill_test_flag_bridges
 
-    run_impl()
-    backfill_test_flag_bridges()
+if __name__ == "__main__":
+    # from one_off_scripts.backfill_daily_test_rollups import run_impl
+    from one_off_scripts.rerun_uploads import rerun_test_results_uploads
+
+    rerun_test_results_uploads("2024-09-25", "2024-09-27")
+
+    # run_impl()
+    # backfill_test_flag_bridges()
--- a/one_off_scripts/rerun_uploads.py
+++ b/one_off_scripts/rerun_uploads.py
@@ -1,0 +1,52 @@
+import datetime as dt
+
+from celery import chord
+from shared.django_apps.reports.models import (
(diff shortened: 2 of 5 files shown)


one_off_script.py
```diff
@@ -5,9 +5,12 @@
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "django_scaffold.settings")
 django.setup()
 
+
 if __name__ == "__main__":
-    from one_off_scripts.backfill_daily_test_rollups import run_impl
-    from one_off_scripts.backfill_test_flag_bridges import backfill_test_flag_bridges
+    # from one_off_scripts.backfill_daily_test_rollups import run_impl
```
Swatinem, line 10: Might be good to somehow control which scripts are run, instead of just commenting things out.
Otherwise, just removing the code for the ones that have been executed is also good.

one_off_scripts/rerun_uploads.py
```diff
@@ -0,0 +1,52 @@
...
+)
+
+from one_off_scripts.celery_stuff import create_signature
+
+
+def rerun_test_results_uploads(start_date, end_date):
+    relevant_reports = (
+        CommitReport.objects.filter(
+            created_at__gt=dt.datetime(2024, 9, 25, 0, 0, 0, 0, dt.UTC),
+            created_at__lt=dt.datetime(2024, 9, 28, 0, 0, 0, 0, dt.UTC),
```
Swatinem, line 16: you should probably use the `start_date`/`end_date` arguments here :-)

one_off_scripts/rerun_uploads.py
```diff
@@ -0,0 +1,52 @@
...
+        )
+        .select_related("testresultreporttotals")
+        .filter(testresultreporttotals__isnull=True)
```
Swatinem, line 19: does `select_related` even have a meaning if you select the ones that are `NULL`?

one_off_scripts/rerun_uploads.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+
+    for report in relevant_reports:
+        print(report)
+        commit = report.commit
+        relevant_uploads = ReportSession.objects.filter(report=report)
+
+        print(relevant_uploads)
```
Swatinem, line 27: do you want to keep these `print` statements? I guess if you run this script manually, it does provide value?

PR #760, opened 2024-10-03
Replace sentry_sdk metrics with statsd metrics

tony-codecov:
This PR replaces the deprecated `sentry_sdk.metrics` with `statsd` metrics throughout the worker codebase. This change is necessary as Sentry SDK metrics are being phased out.

Key changes:
- Removed all instances of sentry_sdk.metrics and replaced them with statsd metrics (imported as `metrics` from `helpers.metrics`)
- Converted tagged sentry_sdk metrics to use dot-separated naming conventions
- Ensured that the metrics object is properly imported where needed

### Legal Boilerplate


adrian-codecov (commented):
I believe we're also wanting to get rid of statsd metrics altogether. Is this an intermediate step towards prometheus? Also some tests are failing but feel free to ping me when those are addressed 👍

PR #764, opened 2024-10-07
Keep track of in-progress Uploads

Swatinem:
This introduces a new component that is responsible for keeping track of the processing status of individual uploads.

The goal here is that this state tracking can make batched Report merging and debounced notifications a lot more intentional.

The state tracking is achieved by keeping sets of `upload_id`s in Redis, one for in-progress (currently being parsed) Uploads, another for processed (and waiting to be merged) Uploads.

---

This is part of https://github.com/codecov/engineering-team/issues/2618

### DIFF
--- a/tasks/upload_processor.py
+++ b/tasks/upload_processor.py
@@ -25,6 +25,7 @@ from helpers.parallel_upload_processing import (
 )
 from helpers.reports import delete_archive_setting
 from helpers.save_commit_error import save_commit_error
+from services.processing.state import ProcessingState
 from services.redis import get_redis_connection
 from services.report import ProcessingResult, RawReportInfo, Report, ReportService
 from services.report.parser.types import VersionOneParsedRawReport
@@ -218,7 +219,7 @@ class UploadProcessorTask(BaseCodecovTask, name=upload_processor_task_name):
                 pr = arguments.get("pr")
                 upload_obj = (
                     db_session.query(Upload)
-                    .filter_by(id_=arguments.get("upload_pk"))
+                    .filter_by(id_=arguments["upload_pk"])
                     .first()
                 )
                 log.info(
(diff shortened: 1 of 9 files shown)


Swatinem:
We discussed this a bit with @giovanni-guidini :

The solution right now is a bit brittle in regards to errors happening in tasks, or tasks just completely vanishing.
The problem is that we might get into a bad state if some uploads have been scheduled to be processed, but the processing never actually completes for whatever reason.

Since this is using different sets of `upload_id`s, I believe one solution would be:
- At the beginning of processing, add the scheduled `upload_ids` to the "processing" set.
  This is a noop in the normal case as the `Upload` task has already put them into this set.
  However, this is relevant when retrying the `Upload` task for whatever reason.
- At the end of processing, remove the scheduled `upload_ids` from the "processing" set.
  This should also be a noop in the normal case, as the `upload_ids` were moved to the "processed" set.
  However, this again in relevant in the retry / max retries reached case.

This should make things a bit less brittle in terms of retries and tasks just dying completely.
However, we could still end up in the following case:
- We have scheduled some tasks to be processed
- The first one succeeds, but final merging / notification is skipped as there is more tasks in the queue
- The second task completely dies, never advancing to the next stage that would trigger merge / notify
  We would thus end up in a state where the final steps are being skipped.
  I’m unsure whether this already happens today in cases of "max retries exceeded": will the `chain`/`chord` completely die in such a case as well?

tasks/upload_processor.py
```diff
@@ -357,6 +368,10 @@ def process_upload(
                 extra=dict(repoid=repoid, commit=commitid),
             )
             raise
+        finally:
+            # this is a noop in the success case, but makes sure unrecoverable errors
+            # are not blocking later merge/notify stages
+            state.clear_in_progress_uploads(upload_ids)
```
giovanni-guidini, line 374: could we have metrics here to know how often this happens?

giovanni-guidini (approved):
There might be some wrinkles to iron out, but I'm generally in favor of keeping track of uploads, so let's go o/

Left 1 comment tho

PR #767, opened 2024-10-08
Move loading and merging intermediate Reports to its own file

Swatinem:
This should make the high level steps of parallel processing a bit more obvious and easier to grasp.

### DIFF
--- a/services/processing/merging.py
+++ b/services/processing/merging.py
@@ -1,0 +1,112 @@
+from dataclasses import dataclass
+
+import sentry_sdk
+from shared.reports.editable import EditableReport, EditableReportFile
+from shared.reports.enums import UploadState
+from shared.reports.resources import Report
+from shared.yaml import UserYaml
+from sqlalchemy.orm import Session as DbSession
+
+from database.models.reports import Upload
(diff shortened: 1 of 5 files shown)


services/processing/merging.py
```diff
@@ -0,0 +1,112 @@
...
+
+@sentry_sdk.trace
+def update_uploads(db_session: DbSession, merge_result: MergeResult):
+    for upload_id, session_id in merge_result.session_mapping.items():
+        upload = db_session.query(Upload).filter(Upload.id_ == upload_id).first()
+        upload.state_id = UploadState.PROCESSED.db_id
+        upload.state = "processed"
+        upload.order_number = session_id
+
+    if upload:
```
giovanni-guidini, line 72: is this supposed to be inside the `for`? It seems to only affect the last upload.
From it being a loop it looks like there might be more than 1 upload

Swatinem: that is indeed confusing, I admit. The `Upload` is really only used to get the DB session, as well as joining the `Upload.report`. It is pretty much querying all the sibling uploads for the same report. so any `Upload` (for the same report, which is true here) will do.

PR #772, opened 2024-10-09
Switch to `orjson` as faster JSON implementation

Swatinem:
`orjson` is supposed to be a faster JSON parser/serializer, with benchmark claiming its ~3x faster at parsing, and up to ~10x faster serializing.

Here I mostly changed the `loads` codepaths, and most of the `dumps` ones that do not define a custom serializer implementation.

### DIFF
--- a/database/utils.py
+++ b/database/utils.py
@@ -1,8 +1,8 @@
-import json
 import logging
 import time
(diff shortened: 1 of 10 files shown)


giovanni-guidini (approved):
nice!

PR #777, opened 2024-10-10
chore: fix (some) type errors in BA notify

giovanni-guidini:
This was not the main goal I had, but these errors were really distracting,
so fixing some type errors in BA notifiers.

One side-effect is that when loading the user_config we are now expecting the comment
config to be `bool | dict` (not `None`). This is in fact the case because we have default
config in place, but the tests didn't know that. Now they do.

I might have also caught a bug in the user activation in terms of the func args.

### Legal Boilerplate

### DIFF
--- a/services/bundle_analysis/notify/__init__.py
+++ b/services/bundle_analysis/notify/__init__.py
@@ -1,5 +1,5 @@
 import logging
-from typing import NamedTuple
+from typing import NamedTuple, Never
 
 import sentry_sdk
 from shared.yaml import UserYaml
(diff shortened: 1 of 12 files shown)


services/bundle_analysis/notify/__init__.py
```diff
@@ -35,9 +35,9 @@ class NotificationFullContext(NamedTuple):
 
 
 class BundleAnalysisNotifyReturn(NamedTuple):
-    notifications_configured: tuple[NotificationType]
-    notifications_attempted: tuple[NotificationType]
-    notifications_successful: tuple[NotificationType]
+    notifications_configured: tuple[NotificationType, ...]
+    notifications_attempted: tuple[NotificationType, ...] | tuple[Never]
+    notifications_successful: tuple[NotificationType, ...] | tuple[Never]
```
michelletran-codecov, line 40: Is there a reason to specify `tuple[Never]`? I believe `tuple[NotificationType, ...]` is inclusive of empty tuples as well.

services/bundle_analysis/notify/contexts/__init__.py
```diff
@@ -188,14 +184,13 @@ def load_user_config(self) -> "NotificationContextBuilder":
...
+        required_changes: bool | Literal["bundle_increase"]
+        if isinstance(comment_config, bool):
+            required_changes = comment_config
             required_changes_threshold = BundleThreshold("absolute", 0)
         else:
-            required_changes: bool | Literal["bundle_increase"] = comment_config.get(
-                "require_bundle_changes", False
-            )
-            required_changes_threshold: int | float = comment_config.get(
+            required_changes = comment_config.get("require_bundle_changes", False)
```
michelletran-codecov, line 192: nit: can we have `required_changes` not return a boolean? I think returning `None` is a bit more intuitive, especially since the other type is a string literal.

michelletran-codecov (approved):
Generally LGTM. Just a few minor comments.

PR #779, opened 2024-10-11
Revert "chore: update shared (get_PR changed to use graphQL) (#776)"

giovanni-guidini:
This reverts commit a4a3d79e185422bb727eb71f9789e823fdd53b42.

### Legal Boilerplate


nicholas-codecov (approved):
send it

PR #782, opened 2024-10-14
Remove all "labels"-related code

Swatinem:
The concepts of `labels`, `labels_index` and `datapoints` are all related to the defunct ATS (automatic test selection) product which was never adopted.

This PR starts out by removing the `label_index` feature, and then progressively removes *all* labels related code, along with never writing out new `datapoints`.


giovanni-guidini (approved):
LGTM

PR #786, opened 2024-10-15
Retry `UploadFinisher` when Report is locked

Swatinem:
Instead of failing with a `LockError`, this copy-pasted the same retry logic as the `UploadProcessor` has.

---

Fixes [WORKER-PJE](https://codecov.sentry.io/issues/5974196957/)

### DIFF
--- a/tasks/upload_finisher.py
+++ b/tasks/upload_finisher.py
@@ -1,5 +1,6 @@
 import contextlib
 import logging
+import random
 import re
 from enum import Enum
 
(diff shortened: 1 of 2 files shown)


tasks/upload_finisher.py
```diff
@@ -154,71 +159,85 @@ def run_impl(
...
+                                commitid=commitid,
+                                commit_yaml=commit_yaml,
+                                report_code=report_code,
+                                parallel_paths=parallel_paths,
+                                processing_results=processing_results,
+                            ),
+                        )
+            except LockError:
+                max_retry = 200 * 3**self.request.retries
+                retry_in = min(random.randint(max_retry // 2, max_retry), 60 * 60 * 5)
```
adrian-codecov, line 229: So 60*60*5 is 5 hours correct? Are we okay that being the minimum number to retry for big numbers of self.request.retries?

adrian-codecov (approved):

tasks/upload_finisher.py
Swatinem: So far this was just copied from the other place that does locking. I would like to revisit this at some point, and reduce the retry intervals, as well as use try-lock instead of blocking.

PR #790, opened 2024-10-15
Created prometheus handler class

adrian-codecov:
We want to separate prometheus responsibilities from the checkpoint responsibilities to later extend its functionality. This PR just rearranges logic and doesn't include new one, to separate reviews into more manageable chunks. 

This PR mainly creates the PrometheusCheckpointLoggerHandler class, which has log_X methods for logging X things. This class is then used in the checkpoint_logger.__init__.py file

### Legal Boilerplate

### DIFF
--- a/helpers/checkpoint_logger/__init__.py
+++ b/helpers/checkpoint_logger/__init__.py
@@ -17,71 +17,14 @@ from typing import (
 )
 
 import sentry_sdk
-from shared.metrics import Counter, Histogram, metrics
+from shared.metrics import metrics
+
+from helpers.checkpoint_logger.prometheus import PROMETHEUS_HANDLER
 
 logger = logging.getLogger(__name__)
 
-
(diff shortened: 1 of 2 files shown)


matt-codecov (approved):

helpers/checkpoint_logger/__init__.py
```diff
@@ -17,79 +17,22 @@
 )
 
 import sentry_sdk
-from shared.metrics import Counter, Histogram, metrics
+from shared.metrics import metrics
```
Swatinem, line 20: I guess you can also use the opportunity to remove all these statsd metrics

Swatinem (approved):

helpers/checkpoint_logger/__init__.py
adrian-codecov: I have a followup PR incoming, I'll remove them in that one 👌

PR #791, opened 2024-10-16
Use correct signatures for Celery Task Hooks

Swatinem:
This explicitly declares and forwards all the arguments to the `on_success/retry/failure` task hooks, as they are documented on https://docs.celeryq.dev/en/main/_modules/celery/app/task.html#Task.on_success

The reason being that the Sentry tags set via the `MetricContext` constructor are not making their way to Sentry, possibly because the `kwargs` were misused in the previous hooks, and were overwriting those tag values with `None`.

So this should ideally solve that mystery.

As a driveby change, I also took the liberty of removing all the deprecated statsd `metrics` calls. All of the relevant metrics also have prometheus equivalents.

### DIFF
--- a/tasks/base.py
+++ b/tasks/base.py
@@ -24,7 +24,6 @@ from helpers.logging_config import (
     log_set_task_id,
     log_set_task_name,
 )
-from helpers.metrics import metrics
 from helpers.telemetry import MetricContext, TimeseriesTimer
 from helpers.timeseries import timeseries_enabled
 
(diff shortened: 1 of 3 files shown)


tasks/base.py
```diff
@@ -283,35 +273,31 @@ def run(self, *args, **kwargs):
             metric_context, f"{self.metrics_prefix}.full_runtime", sync=True
         ):
```
matt-codecov, line 274: this timer will insert a SQL row when it's over, but the db commit happens inside the timed code. so the inserted row happens in a separate transaction that is presumably rolled back when the connection drops, or something. so i think if you want to clean things up you can kill this outer timer while you're at it

matt-codecov (approved):
did you observe that this behaves differently? it's a little surprising to me

matt-codecov (commented):
`self.metrics_prefix` might also be dead code now too

PR #795, opened 2024-10-17
Refactored repository services tests to use fixtures

michelletran-codecov:
This helps reduce database key conflicts by reusing existing objects for tests

This refactor was pulled out of https://github.com/codecov/worker/pull/729 to make changes more manageable. It solved a problem where recreating the objects had conflicts when I switched to Django models (duplicate key errors).

This is also significantly easier to review if you set "hide whitespace" to view the diff because I removed test classes.

### Legal Boilerplate

### DIFF
--- a/services/tests/test_repository_service.py
+++ b/services/tests/test_repository_service.py
@@ -24,6 +24,8 @@ from database.models import Owner
 from database.models.core import (
     GITHUB_APP_INSTALLATION_DEFAULT_NAME,
     GithubAppInstallation,
+    Pull,
+    Repository,
 )
 from database.tests.factories import (
     CommitFactory,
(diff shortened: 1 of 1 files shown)


services/tests/test_repository_service.py
```diff
@@ -1605,24 +1607,22 @@ async def test_fetch_and_update_pull_request_information_torngitexception(
...
         )
         compared_to_commit = CommitFactory.create(
-            repository=repository, branch="master", merged=True
+            repository=repo, branch="master", merged=True
         )
         dbsession.add(commit)
         dbsession.add(compared_to_commit)
         dbsession.flush()
+
+        commit.pullid = "123"
```
joseph-sentry, line 1625: why is this needed?

services/tests/test_repository_service.py
```diff
@@ -2076,6 +2031,10 @@ def test_fetch_commit_yaml_and_possibly_store_commit_yaml_invalid_commit_yaml(
     mock_configuration.set_params({"site": {"comment": {"behavior": "new"}}})
     commit = CommitFactory.create(
         repository__owner__yaml={"coverage": {"precision": 2}},
+        # User needs to be less than PATCH_CENTRIC_DEFAULT_TIME_START
```
joseph-sentry, line 2034: why is this needed?

adrian-codecov (approved):

joseph-sentry (approved):
looks good for the most part, just added a couple comments where i noticed some differences with the original tests

services/tests/test_repository_service.py
michelletran-codecov: Ah, it's a remnant from https://github.com/codecov/worker/pull/729 . Basically, the new Django models adds a default time of now (SQLAlchemy models will just have no default time for the owner object). If the owner is created _after_ the 'PATCH_CENTRIC_DEFAULT_TIME_START', the default owner YAML is actually different and messed up with the tests. So, setting the timestamp is to preserve the test functionality of using the pre-patch_centric yaml behaviour. This doesn't really do anything _bad_ right now, and will be useful later when I start creating the tables with Django later.

services/tests/test_repository_service.py
michelletran-codecov: This was extracted from line 1615. If I leave it in the factory create, it will try to link a pull object and was failing. Since we just needed it to be not null but not actually exist, it was just easier to set it outside the factory call. I don't know why it started failing after I switched to using Django to create the tables 🤷

PR #800, opened 2024-10-18
2686 add checkpoint with repoid

adrian-codecov:
We're extending our checkpoint metrics to add repo specific metrics. The original approach considered adding a repo_id label to the existing checkpoint logger, but this is an ask for a specific repository. This approach isn't necessarily scalable with many repositories as repo_id has high cardinality, which isn't very well suited for prometheus. The better approach would be to leverage sql metrics and extend their functionality, but this approach was chosen for times sake.

This PR

- Adds a new Repository based checkpoints that expect the repo_id
- Added a checkpoint context variable to supply a typed context to metrics
  - Adjusted files that instantiated a checkpoint class to provide said context
- Removes statsd and tests

### Legal Boilerplate


helpers/checkpoint_logger/__init__.py
```diff
@@ -272,30 +271,28 @@ class MyEnum(str, Enum):
     aren't instrumented.
     """
 
-    def log_counters(obj: T) -> None:
-        metrics.incr(f"{klass.__name__}.events.{obj.name}")
-        PROMETHEUS_HANDLER.log_checkpoints(flow=klass.__name__, checkpoint=obj.name)
+    def log_counters(obj: T, context: CheckpointContext = None) -> None:
```
Swatinem, line 274:
```suggestion
    def log_counters(obj: T, context: CheckpointContext | None = None) -> None:
```
the same pattern repeats a bunch of times below. If the default is `None`, then `None` has to appear in the type as well.

helpers/checkpoint_logger/__init__.py
```diff
@@ -358,11 +360,13 @@ def __init__(
         cls: type[T],
         data: Optional[MutableMapping[T, int]] = None,
         strict=False,
+        context: CheckpointContext = None,
     ):
         self.cls = cls
         self.data = data if data else {}
         self.kwargs_key = _kwargs_key(self.cls)
         self.strict = strict
+        self.context = context if context else {}
```
Swatinem, line 369: This should just be `context` in this case, as defaulting to `{}` does not really make sense.

helpers/checkpoint_logger/prometheus.py
```diff
@@ -67,28 +126,46 @@ class PrometheusCheckpointLoggerHandler:
     methods in this class are mainly used by the CheckpointLogger class.
     """
 
-    def log_begun(self, flow: str):
+    def log_begun(self, flow: str, repo_id: int = None):
```
Swatinem, line 129:
```suggestion
    def log_begun(self, flow: str, repo_id: int | None = None):
```
same here, the argument type needs to match up with the default value.

helpers/checkpoint_logger/__init__.py
```diff
@@ -432,15 +436,22 @@ def submit_subflow(self: _Self, metric: str, start: T, end: T) -> _Self:
...
 
         return self
 
 
 def from_kwargs(
-    cls: type[T], kwargs: MutableMapping[str, Any], strict: bool = False
+    cls: type[T],
+    kwargs: MutableMapping[str, Any],
+    strict: bool = False,
+    context: CheckpointContext = None,
```
Swatinem, line 454: for this particular case, you could just pick `repo_id` (aka `repoid`?) out of the `kwargs`.

adrian-codecov: I don't think that would work off the bat as the `repoid` has been specified as a keyword parameter, thus not belonging to **kwargs when it's supplied to the checkpoints functionality, for instance here https://github.com/codecov/worker/blob/5514b97f08b80b0615ea4122e1c3edab89dd3d45/tasks/upload.py#L300-L301. Unless you meant re-adding it to the kwargs object before we supply it. As an add-on, I created the `context` to a) type the supplied params and b) serve as an object that has items additional to the checkpoint flow - I'm open to a different approach, I chose this one for readability + separation of concerns. (Although I will rename repo_id to repoid to be consistent with our model definition)

adrian-codecov: Addressed your other suggestions, thanks 🙏

helpers/tests/unit/test_checkpoint_logger.py
```diff
@@ -497,11 +480,273 @@ def test_reliability_counters(self):
...
-        assert metrics.data["DecoratedEnum.events.BRANCH_2_SUCCESS"] == 1
-        assert metrics.data["DecoratedEnum.total.begun"] == 1
-        assert metrics.data["DecoratedEnum.total.failed"] == 1
-        assert metrics.data["DecoratedEnum.total.ended"] == 3
-        assert metrics.data["DecoratedEnum.total.succeeded"] == 2
+
+    @patch("helpers.checkpoint_logger.prometheus.CHECKPOINT_ENABLED_REPOSITORIES")
+    def test_reliability_counters_with_context(self, mock_object):
+        repoid = 123
+        mock_object.return_value = repoid
```
matt-codecov, line 487: `mock_object` is the feature object, not its `check_value()` function. i wonder why it appears to work anyway?

helpers/tests/unit/test_checkpoint_logger.py
```diff
@@ -497,11 +480,273 @@ def test_reliability_counters(self):
...
+            CounterAssertion(
+                "worker_repo_checkpoints_failed_total",
+                {"flow": "DecoratedEnum", "repoid": "123"},
+                0,
+            ),
+            CounterAssertion(
+                "worker_checkpoints_ended_total", {"flow": "DecoratedEnum"}, 0
+            ),
+            CounterAssertion(
+                "worker_reopo_checkpoints_ended_total",
```
matt-codecov, line 589: typo `reopo`

matt-codecov (changes requested):

helpers/checkpoint_logger/prometheus.py
```diff
@@ -67,28 +126,50 @@ class PrometheusCheckpointLoggerHandler:
     methods in this class are mainly used by the CheckpointLogger class.
     """
 
-    def log_begun(self, flow: str):
+    def log_begun(self, flow: str, repoid: int = None):
+        print("here 111", flow, repoid)
```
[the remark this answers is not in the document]
adrian-codecov: 💀

helpers/checkpoint_logger/__init__.py
```diff
@@ -432,15 +436,22 @@ def submit_subflow(self: _Self, metric: str, start: T, end: T) -> _Self:
...
 
         return self
 
 
 def from_kwargs(
-    cls: type[T], kwargs: MutableMapping[str, Any], strict: bool = False
+    cls: type[T],
+    kwargs: MutableMapping[str, Any],
+    strict: bool = False,
+    context: CheckpointContext | None = None,
```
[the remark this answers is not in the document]
adrian-codecov: So I had to play with this a bit to make it work. The `kwargs.get(_kwargs_key(cls), {}).copy()` is an empty object the first time a flow is ran, so instead I did 

``` python

def from_kwargs(
    cls: type[T],
    kwargs: MutableMapping[str, Any],
    strict: bool = False,
) -> CheckpointLogger[T]:
    context = kwargs.pop("context", CheckpointContext())
    data = kwargs.get(_kwargs_key(cls), {})

    # kwargs has been deserialized into a Python dictionary, but our enum values
    # are deserialized as simple strings. We need to ensure the strings are all
    # proper enum values as best we can, and then downcast to enum instances.
    deserialized_data = {}
    deserialized_data["context"] = context

    for checkpoint, timestamp in data.items():

```

it could be prettier but that's what I could think, wdyt?

helpers/tests/unit/test_checkpoint_logger.py
adrian-codecov: Huh didn't post yesterday. I had a hard time mocking this but ultimately this did the trick, not 100% sure why

helpers/checkpoint_logger/__init__.py
```diff
@@ -426,21 +428,29 @@ def submit_subflow(self: _Self, metric: str, start: T, end: T) -> _Self:
...
 ) -> CheckpointLogger[T]:
+    context = kwargs.pop("context", CheckpointContext())
     data = kwargs.get(_kwargs_key(cls), {})
 
     # kwargs has been deserialized into a Python dictionary, but our enum values
     # are deserialized as simple strings. We need to ensure the strings are all
     # proper enum values as best we can, and then downcast to enum instances.
     deserialized_data = {}
+    deserialized_data["context"] = context
+
```
matt-codecov, line 453: there's a subtle bug here. you have to do `data.pop("context")` because otherwise the for loop later in this function that iterates over `data.items()` will explode when trying to cast `"context"` to an `UploadFlow` enum (or whichever flow). also i think `kwargs.pop("context")` will mutate the passed-in kwargs which might not be safe

taking a step back, i think we don't have a tidy way to manage this context. if we store it in `self.data`, then each flow's `self.data` has its own copy and that's redundant. if we put a single copy in various tasks' `kwargs`, we have to remember to manually pass it and make sure it's included in kwargs when we retry and yadda yadda. since we want to do similar things for `log.info()` calls (task_name, task_id) and SQL metrics (repo_id, owner_id, commit_sha), i wrote https://github.com/codecov/worker/pull/810 which i think can neatly take care of it for all of them

matt-codecov (approved):

PR #804, opened 2024-10-22
Remove some test usage of `ReportDetails`

Swatinem:
A lot of places and assertions still exist.
As long as we are still writing out the `ReportDetails` and `files_array`, it still makes sense to assert that we do not regress.


Swatinem:
Closing in favor of https://github.com/codecov/worker/pull/840

PR #805, opened 2024-10-22
build: update junit xml generation

joseph-sentry:


tasks/tests/unit/test_test_results_processor_task.py
```diff
@@ -20,6 +20,78 @@
 
 
 class TestUploadTestProcessorTask(object):
+    def test_compute_name(self):
+        assert (
+            TestResultsProcessorTask().compute_name(
```
Swatinem, line 25: as a followup, it would be nice to move this method out of `TestResultsProcessorTask`

Swatinem (approved):

tasks/tests/unit/test_test_results_processor_task.py
joseph-sentry: oh i think that second commit wasn't actually supposed to make it in to this PR, i will fix that

PR #806, opened 2024-10-22
fix: update compute_name and add tests

joseph-sentry:
we want to handle the case where there is no actual class name when computing the name for a Pytest test.

this commit also adds unit tests for the compute_name function in test_test_results_processor_task.py

### DIFF
--- a/tasks/test_results_processor.py
+++ b/tasks/test_results_processor.py
@@ -93,7 +93,7 @@ class TestResultsProcessorTask(BaseCodecovTask, name=test_results_processor_task
 
     def compute_name(
         self,
-        framework: Framework,
+        framework: Framework | None,
         raw_classname: str,
         raw_name: str,
         filename: str | None,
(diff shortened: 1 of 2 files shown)


ajay-sentry (approved):
solid

tasks/test_results_processor.py
```diff
@@ -117,7 +117,7 @@ def compute_name(
                     if candidate.test_file_path == filename or (
                         network is not None and candidate.test_file_path in network
                     ):
-                        return f"{candidate.test_file_path}::{candidate.actual_class_name}::{raw_name}"
+                        return f"{candidate.test_file_path}::{f"{candidate.actual_class_name}::" if candidate.actual_class_name else ""}{raw_name}"
```
Swatinem, line 120: nesting these format strings is a bit hard to read, can you move that into a separate variable?

Swatinem (approved):

PR #810, opened 2024-10-22
create LogContext ContextVar and populate in tasks/base.py

matt-codecov:
new functions for managing log context:
- `get_log_context()`
- `set_log_context(LogContext())`
- `update_log_context({"repo_id": 5})`

`tasks/base.py` / `helpers/logging_config.py` use this new `ContextVar` log context instead of globals to put the task name and id on each log message. in separate PRs, we can continue on to:
- automatically add `repo_id`/`owner_id`/`commit_sha` to log messages when available
- get rid of `MetricContext`, making SQL metrics much less cumbersome to work with
- store `CheckpointLogger` data in the `ContextVar` to clean up usage a bit

### history
the globals were put in because previously we got the task details from `get_current_task()` and its `request` member. the request is definitely a thread-local variable and `get_current_task()` may also rely on thread-local state, so when we started pushing things into `async_to_sync()` which runs on a worker thread that context was getting dropped

this PR uses a `ContextVar` instead which is similar to thread locals but should work with Python's non-thread-based cooperative multitasking

caveat: for things run via `async_to_sync()` and `asyncio.run()`, they only inherit copies of the parent's context. so they should get the context set by the parent, but if they make changes the parent will not be able to see them.

### DIFF
--- a/helpers/log_context.py
+++ b/helpers/log_context.py
@@ -1,0 +1,83 @@
+import contextvars
+import logging
+from dataclasses import dataclass, replace
+
+from database.models.core import Commit, Repository
+
+log = logging.getLogger("log_context")
+
+
+@dataclass
+class LogContext:
+    """
+    Class containing all the information we may want to add in logs and metrics.
+    """
+
+    task_name: str = "???"
+    task_id: str = "???"
+
+    _populated_from_db = False
+    owner_id: int | None = None
+    repo_id: int | None = None
+    commit_sha: str | None = None
+    commit_id: int | None = None
+
+    def populate_from_sqlalchemy(self, dbsession):
+        """
+        Attempt to use the information we have to fill in other context fields. For
+        example, if we have `self.repo_id` but not `self.owner_id`, we can look up
+        the latter in the database.
+
+        Ignore exceptions; no need to fail a task for a missing context field.
+        """
+        if self._populated_from_db:
+            return
+
+        try:
+            if self.repo_id:
+                if not self.owner_id:
+                    self.owner_id = (
+                        dbsession.query(Repository.ownerid)
+                        .filter(Repository.repoid == self.repo_id)
+                        .first()[0]
+                    )
+
+                if self.commit_sha and not self.commit_id:
+                    self.commit_id = (
+                        dbsession.query(Commit.id_)
+                        .filter(
+                            Commit.repoid == self.repo_id,
+                            Commit.commitid == self.commit_sha,
+                        )
+                        .first()[0]
+                    )
+        except Exception:
+            log.exception("Failed to populate log context")
+
+        self._populated_from_db = True
+
+
+_log_context = contextvars.ContextVar("log_context", default=LogContext())
+
+
+def set_log_context(context: LogContext):
+    """
+    Overwrite whatever is currently in the log context.
+    """
+    _log_context.set(context)
+
+
+def update_log_context(context: dict):
+    """
+    Add new fields to the log context without removing old ones.
+    """
+    current_context: LogContext = _log_context.get()
+    new_context = replace(current_context, **context)
+    set_log_context(new_context)
+
+
+def get_log_context() -> LogContext:
+    """
+    Access the log context.
+    """
+    return _log_context.get()
--- a/helpers/tests/unit/test_log_context.py
+++ b/helpers/tests/unit/test_log_context.py
@@ -1,0 +1,112 @@
+import asyncio
+
+from asgiref.sync import async_to_sync
+from sqlalchemy.exc import IntegrityError
+
+from database.tests.factories.core import CommitFactory, OwnerFactory, RepositoryFactory
+from helpers.log_context import (
+    LogContext,
+    get_log_context,
+    set_log_context,
+    update_log_context,
+)
+
+
+def create_db_records(dbsession):
+    owner = OwnerFactory.create(
+        service="github",
+        username="codecove2e",
+        unencrypted_oauth_token="test76zow6xgh7modd88noxr245j2z25t4ustoff",
+    )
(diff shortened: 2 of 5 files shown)


helpers/log_context.py
```diff
@@ -0,0 +1,83 @@
...
+    Class containing all the information we may want to add in logs and metrics.
+    """
+
+    task_name: str = "???"
+    task_id: str = "???"
+
+    _populated_from_db = False
+    owner_id: int | None = None
+    repo_id: int | None = None
+    commit_sha: int | None = None
```
adrian-codecov, line 22: This should be a string?

helpers/log_context.py
```diff
@@ -0,0 +1,83 @@
...
+        example, if we have `self.repo_id` but not `self.owner_id`, we can look up
+        the latter in the database.
+
+        Ignore exceptions; no need to fail a task for a missing context field.
+        """
+        if self._populated_from_db:
+            return
+
+        try:
+            if self.repo_id:
```
adrian-codecov, line 37: So this approach relies on for sure having the repo_id? I suppose you couldn't really find the commit/repo from an owner id, although you could find a repo and owner from a commit info

helpers/tests/unit/test_log_context.py
```diff
@@ -0,0 +1,112 @@
...
+
+    assert log_context == LogContext(repo_id=repo.repoid, owner_id=owner.ownerid)
+
+
+def test_populate_just_commit_sha(dbsession):
+    _owner, _repo, commit = create_db_records(dbsession)
+    log_context = LogContext(commit_sha=commit.commitid)
+    log_context.populate_from_sqlalchemy(dbsession)
+
+    assert log_context == LogContext(commit_sha=commit.commitid)
```
adrian-codecov, line 66: Couldn't you in theory know the repo/owner from the commit sha? Or are you assuming sha's here shouldn't be unique, hence you could find multiple records with the same sha? If so, what about a commit sha/id combo?

tasks/base.py
```diff
@@ -250,27 +245,32 @@ def _emit_queue_metrics(self):
...
+                log_context.task_name = task.name
+                log_context.task_id = task.request.id
 
-        metric_context = MetricContext(
-            commit_sha=kwargs.get("commitid"),
-            repo_id=kwargs.get("repoid"),
-            owner_id=kwargs.get("ownerid"),
-        )
+            log_context.populate_from_sqlalchemy(db_session)
+            set_log_context(log_context)
```
adrian-codecov, line 263: Would it make sense to call this `set_log_context()` every time inside the `populate_from_sqlalchemy` instead? I worry that people would subtly forget immediately add it after calling the `log_context.populate_from_sqlalchemy(db_session)` fn

helpers/log_context.py
matt-codecov: `LogContext` doesn't care if you don't even have the repo id. it just figures out whatever it can with whatever it has
- `owner_id` is not enough to figure out repo or commit
- `repo_id` is enough to find owner but not commit
- `repo_id` + `commit_sha` is enough to find owner + commit
- `commit_id` on its own is enough to find everything, but we never have it on its own so that case was never implemented
- `commit_sha` on its own is not enough to find anything because forks/mirrors will each have their own copy of the same SHAs

helpers/tests/unit/test_log_context.py
matt-codecov: mirrors/forks will have their own copies of the same SHAs so just SHA is not enough. if you have the commit's DB ID it's enough to find everything else, but that case was not implemented because it never comes up in our actual code. we don't really pass around the commit's DB ID ever

helpers/log_context.py
matt-codecov: whoops lol good catch

tasks/base.py
matt-codecov: i can make `set_log_context()` call `populate_from_sqlalchemy()` internally

`populate_from_sqlalchemy()` sets/checks a `self._populated_from_db` flag that will skip the queries if it has already been run. i figure that when the task begins you already know the relevant owner, repo, and commit, if they exist. so you can construct the `LogContext` with what you have, fill in any missing info you can from the database, and then lock it in.

tasks/base.py
```diff
@@ -291,10 +291,9 @@ def run(self, *args, **kwargs):
                 self._rollback_django()
                 self.retry()
             finally:
-                log_set_task_name(None)
-                log_set_task_id(None)
                 self.wrap_up_dbsession(db_session)
                 self._commit_django()
+                set_log_context(LogContext())
```
matt-codecov, line 296: note: i delayed clearing the log context because `wrap_up_dbsession` and `_commit_django` both have logs in them. may as well keep the task name, etc

tasks/base.py
matt-codecov: oh actually i don't know if i want to do that

`set_log_context()` is called after `wrap_up_dbsession(dbsession)` so i don't know if it is safe to call `populate_from_sqlalchemy()` at that point. it would be a no-op because it's called with an empty `LogContext`, but that feels sketchy to rely on

helpers/tests/unit/test_log_context.py
adrian-codecov: Makes sense, hadn't considered forks

tasks/base.py
adrian-codecov: Does this mean you need to run `set_log_context()` every time after `log_context.populate_from_sqlalchemy(db_session)` is called?

matt-codecov: if you want to update the contextvar yes, but the db fetch probably doesn't need to happen anywhere except right here

PR #812, opened 2024-10-23
Add a metric to count `carryforward_mode=labels` usage

Swatinem:
This introduces a metric to count the various usages of `carryforward_mode=labels`. It also adds some comments to those places explaining whether the logic might be over- or undercounting the actual real-world usage.

### DIFF
--- a/tasks/upload_finisher.py
+++ b/tasks/upload_finisher.py
@@ -36,6 +36,7 @@ from services.processing.loading import (
     load_intermediate_reports,
 )
 from services.processing.merging import merge_reports, update_uploads
+from services.processing.metrics import LABELS_USAGE
 from services.processing.state import ProcessingState, should_trigger_postprocessing
 from services.redis import get_redis_connection
 from services.report import ReportService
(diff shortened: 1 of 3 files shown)


tasks/upload_finisher.py
```diff
@@ -401,6 +402,10 @@ def finish_reports_processing(
             commit.state = "skipped"
 
         if self.should_clean_labels_index(commit_yaml, processing_results):
+            # NOTE: according to tracing, the cleanup task is never actually run.
+            # reasons for that might be that we indeed *never* have any flags
+            # configured for `carryforward_mode=labels`, or the logic is somehow wrong?
```
giovanni-guidini, line 407: I think it _might_ be because no repos use the label index at all
(see [here](https://github.com/codecov/worker/blob/ad3621cbee3619ec1b47ee55a1050e8be4343dda/services/report/raw_upload_processor.py#L87))

giovanni-guidini (approved):
Excited to see how this turns out

tasks/upload_finisher.py
Swatinem: well yes, though the conditional above in `UploadFinisher` is not looking at that feature flag at all.

PR #814, opened 2024-10-23
Replace Sentry metrics with Prometheus metrics

tony-codecov:
Replace Sentry metrics with Prometheus metrics.

Closes https://github.com/codecov/infrastructure-team/issues/461.

### Legal Boilerplate


services/bundle_analysis/report.py
```diff
@@ -99,17 +99,14 @@ def update_upload(self, carriedforward: Optional[bool] = False) -> None:
...
-        try:
-            BUNDLE_ANALYSIS_REPORT_PROCESSOR_COUNTER.labels(
+        inc_counter(
+            BUNDLE_ANALYSIS_REPORT_PROCESSOR_COUNTER,
+            labels=dict(
                 result="upload_error" if self.error else "processed",
                 plugin_name="n/a",
                 repository=self.commit.repository.repoid,
-            ).inc()
-        except Exception:
```
Swatinem, line 108: I wonder if this is actually fallible?
having a fn indirection just to slap a `try/except` on this call seems a bit weird, in particular as the call with the extra `labels` argument is a bit more unwieldy than just using the prometheus API directly.

Swatinem (approved):

services/bundle_analysis/report.py
tony-codecov: That's a good point, I think the existing code with the try/except block should work as is, and the inc_counter change here is not necessary.

Another thing I'm curious about this metric though is that there are 4 different `results`, some number of unique `plugin_name` and so for each unique `repo_id` there will be `4*count(plugin_name)` time series created, and such high cardinality might be a performance issue with Prometheus. What do you think about this or am I overlooking anything?

Swatinem: You are absolutely right.
IMO its okay to just remove the `repo_id` tag here. You can also check in with @adrian-codecov who has recently worked on getting some limited form of repo-tagged metrics going using an allowlist to only tag very specific repos.

tony-codecov: I've removed `repo_id` for now. The allowlist method can probably be implemented later when we have determined  which label(s) a `repo_id` should be classified into.

PR #816, opened 2024-10-24
Fix notifications with an empty HEAD report

Swatinem:
There were a couple of errors spread around the notification code assuming that a `head.report` always exists, which is not necessarily the case.

In particular that would be the case when using `empty-upload` or `manual_trigger`.

---

Fixes [WORKER-P6A](https://codecov.sentry.io/issues/5801299347/)

### DIFF
--- a/services/notification/notifiers/codecov_slack_app.py
+++ b/services/notification/notifiers/codecov_slack_app.py
@@ -84,6 +84,7 @@ class CodecovSlackAppNotifier(AbstractBaseNotifier):
             message = "unknown"
             notation = ""
             comparison_url = None
+        head_report = comparison.head.report
         return {
             "url": comparison_url,
             "message": message,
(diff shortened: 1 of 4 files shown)


spalmurray-codecov (approved):
lgtm, but you may want a platform review as well - your call 🙂

services/notification/notifiers/codecov_slack_app.py
```diff
@@ -84,6 +84,7 @@ def build_payload(self, comparison: ComparisonProxy) -> dict:
             message = "unknown"
             notation = ""
             comparison_url = None
+        head_report = comparison.head.report
```
giovanni-guidini, line 87: Is `comparison.head` typed as a `FullCommit` in the code? Maybe it would be a good idea to update the type to account for that. I've seen examples where the `commit` might also be `None` and have a half-done PR to update it

giovanni-guidini (approved):

services/notification/notifiers/codecov_slack_app.py
Swatinem: I’m unsure about that one. I had the problem that `comparison` might be wrongly typed, as some tests were failing because it was actually a `FilteredComparison`, though I’m not sure if that only happened because of tests, or if that can also happen in real world code. In other parts of the notification infra, I have updated the types to take that into account.

PR #817, opened 2024-10-24
Only schedule parallel Upload Tasks

Swatinem:
With parallel upload processing fully rolled out, its time to start cleaning up the feature flag.

This first step is changing the initial `Upload` task to only schedule the parallel version of `UploadProcessor/Finisher`.

As there will probably be a bunch of already scheduled tasks at time of deploy, this does not yet change those tasks to assume they run in fully parallel mode.

### DIFF
--- a/tasks/upload_processor.py
+++ b/tasks/upload_processor.py
@@ -80,6 +80,7 @@ class UploadProcessorTask(BaseCodecovTask, name=upload_processor_task_name):
         commitid,
         commit_yaml,
         arguments_list,
+        arguments=None,
         report_code=None,
         **kwargs,
     ):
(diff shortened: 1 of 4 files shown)


tasks/upload_processor.py
```diff
@@ -80,6 +80,7 @@ def run_impl(
         commitid,
         commit_yaml,
         arguments_list,
+        arguments=None,
```
giovanni-guidini, line 83: do you know the type of `arguments` or `arguments_list`?

giovanni-guidini (approved):
LGTM 🎉

tasks/upload_processor.py
Swatinem: yes, I have plans to properly type those as `TypedDict` in the future.

PR #818, opened 2024-10-24
delete MetricContext, populate sentry tags from LogContext

matt-codecov:
- deletes `MetricContext`, cleans up usage of SQL metrics
  - this should also eliminate some `metric_context.populate()` db queries although there were not many
- adds `LogContext` fields to sentry tags
- makes `LogContext` the decider of how its fields are added to log records and sentry tags
- cleans up the redundant `task_id` and `task_name` log fields


helpers/log_context.py
```diff
@@ -124,15 +125,25 @@ def populate_from_sqlalchemy(self, dbsession):
 
         self._populated_from_db = True
 
+    def add_to_log_record(self, log_record: dict):
+        log_record["context"] = self.as_dict()
+
+    def add_to_sentry(self):
+        d = self.as_dict()
+        d.pop("sentry_trace_id", None)
+        sentry_sdk.set_tags(d)
```
Swatinem, line 134: Interesting, I see [`set_tags`](https://getsentry.github.io/sentry-python/apidocs.html#sentry_sdk.Scope.set_tags) as a `Scope` API, but not as part of the [top level APIs](https://getsentry.github.io/sentry-python/api.html#sentry_sdk.api.set_tag).

Maybe this is just an oversight in the docs? CC @antonpirker

helpers/telemetry.py
```diff
@@ -32,114 +30,56 @@ def wrapper(*args, **kwargs):
...
+        log.exception("Failed to create telemetry_simple record")
 
+
+@fire_and_forget
+async def attempt_log_simple_metric(name: str, value: float):
+    """
     `attempt_log_simple_metric()` is the @fire_and_forget version of
-    `log_simple_metric()`, meaning it will throw populating/logging on the
+    `log_simple_metric()`, meaning it will throw logging on the
     asyncio queue and then immediately return without needing to be awaited. Can
```
Swatinem, line 68: as this is now using a `ContextVar`, is that guaranteed to be more consistent than a global which might change underneath?

Swatinem (approved):

helpers/telemetry.py
matt-codecov: y'know, good question. i think so; as i understand it the context will inherit a clone of the parent context so the parent context is free to change

PR #820, opened 2024-10-25
chore: improve (very little) some types

giovanni-guidini:
The important change here is making the `FullCommit` attributes nullable.
We have seen evidence in Sentry issues (although you will just have to take my word for it)
that these things can be `None`.
Actually even in the code when creating `FilteredComparison` you can see the `Report` might be
`None`.

In any case, after making this change I adjusted some types here and there,
and make explicit "return None" when the thing we depend on is not available.

### Legal Boilerplate


services/comparison/overlays/critical_path.py
```diff
@@ -1,20 +1,26 @@
...
 from services.repository import get_repo_provider_service
 from services.yaml import get_current_yaml
 
 sentinel = object()
 
 
-def _get_latest_profiling_commit(comparison):
+def _get_latest_profiling_commit(comparison) -> ProfilingCommit | None:
+    """
+    @param comparison: ComparisonProxy (not imported due to circular imports)
```
Swatinem, line 20: You can move that import behind a  `if typing.TYPE_CHECKING:` guard.
See https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING and https://stackoverflow.com/questions/61545580/how-does-mypy-use-typing-type-checking-to-resolve-the-circular-import-annotation

Same below as well.

PR #826, opened 2024-10-28
Fix notifier without any layout

Swatinem:
This accounts for the case when `settings` does not have any `"layout"`.

Also cleans up the code a little bit, using newer type annotations, and removing statsd timers.

---

Should fix [ECDN-WORKER-D7G](https://codecov.sentry.io/issues/5899928678/)
fixes https://github.com/codecov/internal-issues/issues/974

### DIFF
--- a/services/notification/notifiers/mixins/message/__init__.py
+++ b/services/notification/notifiers/mixins/message/__init__.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Callable
+from collections.abc import Callable
 
 from shared.django_apps.core.models import Repository
 from shared.reports.resources import ReportTotals
--- a/services/notification/notifiers/mixins/message/__init__.py
+++ b/services/notification/notifiers/mixins/message/__init__.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Callable, List
+from typing import Callable
 
 from shared.django_apps.core.models import Repository
 from shared.reports.resources import ReportTotals
@@ -7,7 +7,6 @@ from shared.validation.helpers import LayoutStructure
 
 from database.models.core import Owner
 from helpers.environment import is_enterprise
-from helpers.metrics import metrics
 from services.billing import BillingPlan
 from services.comparison import ComparisonProxy, FilteredComparison
 from services.notification.notifiers.mixins.message.helpers import (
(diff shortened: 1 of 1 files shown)


services/notification/notifiers/mixins/message/__init__.py
```diff
@@ -1,13 +1,12 @@
 import logging
-from typing import Callable, List
+from typing import Callable
```
giovanni-guidini, line 2: [nit] `typing.Callable` is deprecated in favor of `collections.abc.Callable` (see [here](https://docs.python.org/3/library/typing.html#annotating-callable-objects))

services/notification/notifiers/mixins/message/__init__.py
```diff
@@ -202,98 +201,92 @@ def _possibly_write_install_app(
...
+
+        if "newfooter" in sections or "condensed_footer" in sections:
             return "newfooter"
+        return None
+
+
+def get_sections(settings: dict) -> list[str]:
+    layout = (settings.get("layout") or "").strip()
+    if not layout:
+        return []
```
giovanni-guidini, line 290: I think if you return nothing the PR comment would be empty.
It's better to return the default value in case there is none.

The default value is [here](https://github.com/codecov/shared/blob/fccd4796bfb4df48eb7c6f17191c145c3c587109/shared/config/__init__.py#L36) (and it should not be lost when the YAML is processed, but merged with the user's option - I think that's the real bug)

services/notification/notifiers/mixins/message/__init__.py
```diff
@@ -202,98 +201,92 @@ def _possibly_write_install_app(
...
-        with metrics.timer(
-            f"worker.services.notifications.notifiers.comment.section.{writer_class.name}"
-        ):
-            # Settings here enable failed tests results for now as a new product
-            message.extend(
-                line
-                for line in writer_class.header_lines(
-                    comparison=comparison, diff=diff, settings=settings
-                )
+        # Settings here enable failed tests results for now as a new product
```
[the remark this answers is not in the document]
Swatinem: No idea TBH :-D

Swatinem:
Closing this, as this was only happening in QA for our own repos, and
https://github.com/codecov/worker/pull/890 and https://github.com/codecov/codecov-api/pull/980 would fix the cause of the error.

The error is caused by `comment: false` as configured in the QA org is not being properly overridden when a repo yaml is only setting/overriding very specific fields of `comment`.

PR #830, opened 2024-10-29
Rewrite the core `UploadProcessor` logic

Swatinem:
The code has been moved out of the `UploadProcessor` task into its own file, and adapted in such a way that it will only deal with a single upload per task invocation.

### DIFF
--- a/services/processing/processing.py
+++ b/services/processing/processing.py
@@ -2,6 +2,7 @@ import logging
 from typing import Callable, NotRequired, TypedDict
 
 import sentry_sdk
+from celery.exceptions import CeleryError
 from shared.yaml import UserYaml
 from sqlalchemy.orm import Session as DbSession
 
@@ -78,7 +79,15 @@ def process_upload(
 
         rewrite_or_delete_upload(archive_service, commit_yaml, report_info)
 
+    except CeleryError:
+        raise
     except Exception:
+        commit.state = "error"
+        db_session.commit()
+        log.exception("Could not properly process commit")
+        raise
+
+    finally:
         # this is a noop in the success case, but makes sure unrecoverable errors
         # or retries are not blocking later merge/notify stages
         state.clear_in_progress_uploads([upload_id])
--- a/services/processing/processing.py
+++ b/services/processing/processing.py
@@ -1,5 +1,6 @@
 import logging
-from typing import Callable, NotRequired, TypedDict
+from collections.abc import Callable
+from typing import NotRequired, TypedDict
 
 import sentry_sdk
 from celery.exceptions import CeleryError
--- a/services/processing/processing.py
+++ b/services/processing/processing.py
@@ -1,0 +1,108 @@
+import logging
+from typing import Callable, NotRequired, TypedDict
+
+import sentry_sdk
+from shared.yaml import UserYaml
+from sqlalchemy.orm import Session as DbSession
+
+from database.models.core import Commit
+from database.models.reports import Upload
+from helpers.reports import delete_archive_setting
+from services.archive import ArchiveService
+from services.processing.intermediate import save_intermediate_report
+from services.processing.state import ProcessingState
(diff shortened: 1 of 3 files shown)


services/processing/processing.py
```diff
@@ -0,0 +1,117 @@
+import logging
+from typing import Callable, NotRequired, TypedDict
```
giovanni-guidini, line 2: [nit] `typing.Callable` is deprecated in favor of `collections.abc.Callable`

services/processing/processing.py
```diff
@@ -0,0 +1,117 @@
...
+        log.info("Finished processing upload", extra={"result": result})
+
+        report_service.update_upload_with_processing_result(upload, processing_result)
+        save_intermediate_report(archive_service, commit_sha, upload_id, report)
+        state.mark_upload_as_processed(upload_id)
+
+        rewrite_or_delete_upload(archive_service, commit_yaml, report_info)
+
+    except CeleryError:
+        raise
```
giovanni-guidini, line 83: [nit, feel free to ignore] only because I've spent quite some time looking at `ruff` output, one of the rules say you are supposed to re-raise errors like so:

```
except CeleryError as err:
    raise Exception("panic") from err
```
    
The only real difference is that it sets the 1st exception as the direct cause of the re-raise (which is true). Idk if it makes difference in Sentry issues page or logs... Read more: https://stackoverflow.com/questions/24752395/python-raise-from-usage

tasks/upload_processor.py
```diff
@@ -66,225 +61,36 @@ class UploadProcessorTask(BaseCodecovTask, name=upload_processor_task_name):
...
-    @sentry_sdk.trace
-    def postprocess_raw_reports(
-        self,
-        report_service: ReportService,
-        commit: Commit,
-        reports: list[RawReportInfo],
-    ):
-        should_delete_archive_setting = delete_archive_setting(
-            report_service.current_yaml
+                self.retry(max_retries=MAX_RETRIES, countdown=FIRST_RETRY_DELAY)
```
giovanni-guidini, line 84: I'm probably missing something, but if it only retries if `self.request.retries` is `0` why do we set a `MAX_RETRIES` of `5`?

tasks/upload_processor.py
```diff
@@ -66,225 +61,36 @@ class UploadProcessorTask(BaseCodecovTask, name=upload_processor_task_name):
...
-
-        raw_reports: list[RawReportInfo] = []
-        try:
-            for arguments in arguments_list:
-                upload_obj = (
-                    db_session.query(Upload)
-                    .filter_by(id_=arguments["upload_pk"])
-                    .first()
-                )
+        def on_processing_error(error: ProcessingError):
```
giovanni-guidini, line 77: I like this 👍 looks clean and elegant

giovanni-guidini (approved):
much code, very LGTM

services/processing/processing.py
Swatinem: This is an interesting concept. Though I want to explicitly pass on the exception un-chained, as I’m not sure if celery can deal with chained exceptions. And I’m also not sure if adding any exception context in such a way is really provides any value in the first place.

tasks/upload_processor.py
Swatinem: To be quite honest, I also don’t really know why that is. I just moved this existing code around.

In the long term I would like to get rid of this failure condition / retry completely. This case happens when the `UploadProcessor` runs when the upload itself has not finished, so the file we were looking for does not exist (yet).

Not sure if shelter can help with this, but I will ask around. There is a couple more things I would love to clean up in relation to this.

giovanni-guidini: I don't think shelter can help much with the upload not being complete because we are waiting on the user uploading the raw report to storage in this case... I think.

PR #832, opened 2024-10-29
Fix leaky usage of time machine

michelletran-codecov:
This was leaking into other tests (other tests were using the wrong times).

Pulled this out from https://github.com/codecov/worker/pull/729

### Legal Boilerplate


Swatinem (approved):
I believe sentry was using `freezegun` at some point, not sure if they migrated away from that, or to that, which provides a decorator for a function. Might be another alternative, but either is fine.

PR #834, opened 2024-10-30
Do not clear the `LogContext` on task finish

Swatinem:
Previously, this would unset the `LogContext` back to an empty version at the end of the task `run` hook. However this would also unset all the previously set Sentry tags. As the Sentry Celery integration automatically wraps the `run` function, the Sentry tags have to persist past the end of said function.

This will now only set an initialized `LogContext` when entering `run`, and keep that around until the next task overwrites it. That way, we should also get an initialized context for any async code that runs outside of the `run` scope.

### DIFF
--- a/helpers/log_context.py
+++ b/helpers/log_context.py
@@ -1,6 +1,6 @@
 import contextvars
 import logging
-from dataclasses import asdict, dataclass, replace
+from dataclasses import asdict, dataclass
 
 import sentry_sdk
 from sentry_sdk import get_current_span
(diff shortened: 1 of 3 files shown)


giovanni-guidini (commented):
I think this would require a change in initialization to also always set `_populated_from_db` to `False` in order to prevent [a new attempt at populating values](https://github.com/codecov/worker/blob/abf43da775d4a31914cd3b82db51d7d0891d5c16/helpers/log_context.py#L50) to not do that and the values of the 1st task bleed into the 2nd

Also not sure how contextvars work... is the context passed into threads or does every async thread has it's own context?

Swatinem:
I’m not sure I follow.
The only code that initializes and sets the `LogContext` is here: https://github.com/codecov/worker/blob/edeadb6d768476ef8e7f8472b43ce55806951c42/tasks/base.py#L251-L263

That is also the only code outside of tests that ever populates the context. The constructor will start out with `_populated_from_db: False`

matt-codecov (approved):

PR #836, opened 2024-10-30
Only show VCR logs on errors

michelletran-codecov:
These logs were getting really noisy when testing locally (especially when running with verbose mode `-vv`). Removing unnecessary logs should make it more manageable to debug, especially since the VCR logs aren't super useful. If someone wants the full log, they can remove this or override manually.

### Legal Boilerplate

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,3 +1,4 @@
+import logging
 import os
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


Swatinem (approved):
❤️ these were annoying me quite a bit :-D

PR #843, opened 2024-10-31
fix: deadlock when inserting tests and rollups

joseph-sentry:
we deadlock when inserting (on conflict do update is the part i think
causes this) tests possibly because each worker task is inserting tests
in a different order, this fixes that by making sure if tasks are inserting
the same tests they will insert them in the same order

does the same for the rollups batch insert on conflict do update

### DIFF
--- a/tasks/test_results_processor.py
+++ b/tasks/test_results_processor.py
@@ -338,7 +338,7 @@ class TestResultsProcessorTask(BaseCodecovTask, name=test_results_processor_task
         # Upsert Tests
         if len(test_data) > 0:
             test_insert = insert(Test.__table__).values(
-                sorted(list(test_data.values()), key=lambda x: str(x["flags_hash"]))
+                sorted(test_data.values(), key=lambda x: str(x["flags_hash"]))
             )
             insert_on_conflict_do_update = test_insert.on_conflict_do_update(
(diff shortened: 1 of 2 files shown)


tasks/test_results_processor.py
```diff
@@ -337,7 +337,9 @@ def create_daily_total():
 
         # Upsert Tests
         if len(test_data) > 0:
-            test_insert = insert(Test.__table__).values(list(test_data.values()))
+            test_insert = insert(Test.__table__).values(
+                sorted(list(test_data.values()), key=lambda x: str(x["flags_hash"]))
```
Swatinem, line 341:
```suggestion
                sorted(test_data.values(), key=lambda x: str(x["flags_hash"]))
```

`sorted` takes an iterator.

Swatinem (approved):
Q: I have seen a ton of unique `INSERT` statements with a variable number of items it is inserting. is there a way to batch these, so we can reduce the number of unique queries?

joseph-sentry:
> Q: I have seen a ton of unique INSERT statements with a variable number of items it is inserting. is there a way to batch these, so we can reduce the number of unique queries?

i don't think there's a way to stop that since each test results file may contain any number of new tests or test instances so there's no way to stop that

however we are inserting in each processor task instead of the one finisher, so technically there is a way to reduce the number of insert statements being run that way

PR #844, opened 2024-11-01
fix: rewriting of test result reports

joseph-sentry:
the rewrite_readable and writing of the storage file should've been outside the loop, having them inside the loop was:
1. doing extra useless work by calling rewrite_readable a bunch of times
2. causing rate limit errors in GCS due to trying to write to the same object so often, resulting in the task failing


tasks/test_results_processor.py
```diff
@@ -495,7 +495,7 @@ def process_individual_arg(
 
             readable_report = self.rewrite_readable(network, report_contents)
```
matt-codecov, line 496: should this also be moved out of the loop?

michelletran-codecov (approved):

PR #846, opened 2024-11-02
904 update django dependency

adrian-codecov:
Updating django as there are vuln tickets

https://github.com/codecov/internal-issues/issues/904
https://github.com/codecov/internal-issues/issues/905

### Legal Boilerplate


Swatinem (approved):
Is there a way to just bump all the dependencies (maybe on a regular schedule) to the latest semver compatible version?

PR #851, opened 2024-11-04
Fix some flakey tests

michelletran-codecov:
Extracted this from https://github.com/codecov/worker/pull/729

A few of the tests become flaky when migrating to the Django models because of some unique references that's defined in Django, but not SQLAlchemy. 

In particular, the way that FactoryBoy creates references can be flakey (i.e. Commits referencing Pulls. If the original `pull` is not flushed, there is a chance that the `commit` gets created first, which inadvertently creates a separate pull with the `id` of the one that you specified to create). So, doing an explicit flush will of `pull` before creating the `commit` will fix this. 

Also fixed a test where DB returned results are not necessarily ordered.

### Legal Boilerplate

### DIFF
--- a/tasks/tests/unit/test_test_results_processor_task.py
+++ b/tasks/tests/unit/test_test_results_processor_task.py
@@ -525,12 +525,9 @@ api/temp/calculator/test_calculator.py:30: AssertionError</failure></testcase></
 
         test_flag_bridges = dbsession.query(TestFlagBridge).all()
 
-        assert [bridge.test_id for bridge in test_flag_bridges] == [
-            tests[1].id,
-            tests[2].id,
-            tests[3].id,
-            tests[4].id,
-        ]
+        assert set(bridge.test_id for bridge in test_flag_bridges) == set(
+            instance.test_id for instance in test_instances
+        )
         for bridge in test_flag_bridges:
             assert bridge.flag == repo_flag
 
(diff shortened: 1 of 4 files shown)


tasks/tests/unit/test_test_results_processor_task.py
```diff
@@ -525,12 +525,10 @@ def test_upload_processor_task_call_existing_test_diff_flags_hash(
...
-        assert [bridge.test_id for bridge in test_flag_bridges] == [
-            tests[1].id,
-            tests[2].id,
-            tests[3].id,
-            tests[4].id,
-        ]
+        # assert that test_flag_bridges is a subset of tests
+        assert set(bridge.test_id for bridge in test_flag_bridges).issubset(
+            set(x.id for x in tests)
+        )
```
Swatinem, line 531: previously this was testing for equality, not a subset relation. is this really what we want?

Swatinem (approved):

tasks/tests/unit/test_test_results_processor_task.py
michelletran-codecov: Hmm, good point, the subset test doesn't really make sense.  We might actually want to check that the test_ids in `test_flag_bridge` match the ones in `test_instances` (non-failure, I'm guessing). The flakiness with this test is that it assumes ordering of `tests`, where we should not have that assumption.

Since I'm having to guess the intent of this check, I can also be convince that this test isn't super useful, and just remove it. I'm leaning towards just modifying it to match the ids in `test_instance` out of conservatism.

PR #852, opened 2024-11-04
Remove `labels-index` feature flag

Swatinem:
This feature flag was never fully turned on, and most likely never will.


giovanni-guidini (approved):
Still in draft, but lgtm.
I mean, we are not using this code anyway, feels good to have less unused code

PR #853, opened 2024-11-05
fix: remove get_existing_tests from test process

joseph-sentry:
this query is very slow and we don't necessarily need it, a side effect is that this will mean the insert on conflict do nothing will contain rows that it doesn't need to but that's fine

depends on: https://github.com/codecov/shared/pull/414


adrian-codecov (approved):
Should there be a test file affected by this?

joseph-sentry:
> Should there be a test file affected by this?

no, this change is internal to the test results processor task

PR #855, opened 2024-11-06
Optimize `update_commit_from_provider_info`

Swatinem:
This makes the `possibly_` variant of this function sync, to avoid an unnecessary `async_to_sync` call.

The main function is changed in such a way that:
- It early-returns when no commit was found, de-indenting the rest of the block
- Most importantly, it moves some of the field writes around, and adds an explicit `flush` at the end. That way, it avoids splitting the `UPDATE` into two different queries, and rather does one update at the end.
- A query indirection doing a `COUNT` before the actual query is removed by just doing the filtering on the Python side
- It might avoid a second provider request for a bitbucket edgecase.

---

Part of https://github.com/codecov/engineering-team/issues/2824

### DIFF
--- a/services/repository.py
+++ b/services/repository.py
@@ -90,6 +90,7 @@ def _get_repo_provider_service_instance(service: str, adapter_params: dict):
     )
 
 
+@sentry_sdk.trace
 async def fetch_appropriate_parent_for_commit(
     repository_service, commit: Commit, git_commit=None
 ):
@@ -107,6 +108,7 @@ async def fetch_appropriate_parent_for_commit(
         possible_commit = possibly_filter_out_branch(commit, possible_commit_query)
         if possible_commit:
             return possible_commit.commitid
+
     ancestors_tree = await repository_service.get_ancestors_tree(commitid)
     elements = [ancestors_tree]
     while elements:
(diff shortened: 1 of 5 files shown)


services/repository.py
```diff
@@ -187,82 +191,84 @@ async def update_commit_from_provider_info(repository_service, commit):
...
+
+    if commit.repository.service == "bitbucket":
+        res = merged_pull(git_commit["message"])
+        if res:
+            pullid = res.groups()[0]
+            if int(pullid) != int(commit.pullid):
+                pull_details = await repository_service.get_pull_request(pullid)
+            commit.branch = pull_details["base"]["branch"]
+
+    db_session.flush()
```
michelletran-codecov, line 262: Just for my owner understanding, I'm curious to how this flush helps with reducing the number of updates. I had assumed that we had to either flush (or commit, which does a flush first) for the update to "happen" in the transaction, so I might have the wrong mental model on how it works!

michelletran-codecov (approved):
LGTM

services/repository.py
Swatinem: The number of updates are reduced by moving code around.
Previously, some fields on the current `Commit` were written, and those were apparently implicitly flushed by sqalchemy as another function was querying `Commit`s.
Now the queries happen first, and all the field updates second, whereby the `UPDATE` should update all the fields at once.

The `flush` is just there to force the `UPDATE` to happen inside of this function, as previously, it would be delayed till later on nested in a completely different call chain.

Though to be quite honest, I haven’t verified this behavior locally. We will see this more clearly in tracing.

Swatinem:
Indeed, queries https://codecov.sentry.io/insights/backend/database/spans/span/0ea8c24d682118d8/ and https://codecov.sentry.io/insights/backend/database/spans/span/82eab313cb32ec97/ were merged into https://codecov.sentry.io/insights/backend/database/spans/span/c24b045b17863463/

PR #862, opened 2024-11-07
Further cleanup `UploadArguments` and normalize `flags`

Swatinem:
Normalizes the `flags`, so they are always a `list`. Otherwise, this uses the `UploadArguments` type in more places, and cleans up usage of `upload_pk` in more places.

Another cleanup was removing the backwards compatibility for the old `processing_results` being passed from the `Processor` to the `Finisher`.


nora-codecov (approved):

tasks/bundle_analysis_processor.py
```diff
@@ -68,33 +67,28 @@ def run_impl(
                 retry_num=self.request.retries,
             ):
                 return self.process_impl_within_lock(
-                    db_session=db_session,
-                    repoid=repoid,
-                    commitid=commitid,
-                    commit_yaml=commit_yaml,
-                    params=deepcopy(params),
```
adrian-codecov, line 75: I see we got rid of deepcopy here but I don't see any writing/changes to anything inside of it. I'm curious why we had this in the first place 🤔

PR #864, opened 2024-11-08
flip default value on upload_token_required_for_public_repos

nora-codecov:
when true tokenless goes live we will merge this, which will make true tokenless default to "on" for new orgs.

https://github.com/codecov/engineering-team/issues/2623


database/models/core.py
```diff
@@ -83,7 +83,7 @@ class Owner(CodecovBaseModel):
     stripe_customer_id = Column(types.Text, server_default=FetchedValue())
     stripe_subscription_id = Column(types.Text, server_default=FetchedValue())
     onboarding_completed = Column(types.Boolean, default=False)
-    upload_token_required_for_public_repos = Column(types.Boolean, default=True)
+    upload_token_required_for_public_repos = Column(types.Boolean, default=False)
```
adrian-codecov, line 86: Is this a DB default or an object default? And could we add a test that confirms this works as expected?

nora-codecov: Object default. With django, the `default` is managed in code and the db does not have a `default` for this field. Since it is non-nullable, worker needs to know what default to provide to the db.

In terms of the test - just a test to show that when creating an `Owner` the field is set to the default value?

adrian-codecov: Yup!

adrian-codecov (approved):

PR #867, opened 2024-11-08
Optimize Report merging

Swatinem:
This change tries to optimize the slow `Report.merge` operation.

When merging report `B` into an existing report `A`, we can just use `B` in case `A` is empty. Also, the speed of merging depends on the size of the right hand size Report (`B` in this example). By swapping the reports, we can make sure to always merge the smaller Report into the larger one.

### DIFF
--- a/services/processing/merging.py
+++ b/services/processing/merging.py
@@ -23,7 +23,7 @@ def merge_reports(
     commit_yaml: UserYaml,
     master_report: Report,
     intermediate_reports: list[IntermediateReport],
-) -> MergeResult:
+) -> tuple[Report, MergeResult]:
     session_mapping: dict[int, int] = dict()
     deleted_sessions: set[int] = set()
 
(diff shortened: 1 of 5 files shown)


services/processing/merging.py
```diff
@@ -34,9 +34,14 @@ def merge_reports(
 
         old_sessionid = next(iter(report.sessions))
         new_sessionid = master_report.next_session_number()
-        change_sessionid(report, old_sessionid, new_sessionid)
         session_mapping[intermediate_report.upload_id] = new_sessionid
 
+        if master_report.is_empty() and old_sessionid == new_sessionid:
+            # if the master report is empty, we can avoid a costly merge operation
```
adrian-codecov, line 40: This happens the very first time there's an upload for a commit right?

adrian-codecov (approved):

services/processing/merging.py
Swatinem: yes exactly, except when using carry-forwards

PR #869, opened 2024-11-08
fix: minimize deadlocks

joseph-sentry:
Fixes: https://github.com/codecov/internal-issues/issues/1019

### DIFF
--- a/tasks/test_results_processor.py
+++ b/tasks/test_results_processor.py
@@ -336,7 +336,7 @@ class TestResultsProcessorTask(BaseCodecovTask, name=test_results_processor_task
             test_insert = insert(Test.__table__).values(
                 sorted(
                     test_data.values(),
-                    key=lambda x: str(x["test_id"]),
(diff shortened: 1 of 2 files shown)


tasks/test_results_processor.py
```diff
@@ -334,7 +334,10 @@ def create_daily_total():
         # Upsert Tests
         if len(test_data) > 0:
             test_insert = insert(Test.__table__).values(
-                sorted(test_data.values(), key=lambda x: str(x["flags_hash"]))
+                sorted(
+                    test_data.values(),
+                    key=lambda x: str(x["test_id"]),
```
adrian-codecov, line 339: Getting 3 errors w/ this change, this intentional?

joseph-sentry: my latest change should fix those

adrian-codecov: Sweet!

adrian-codecov (approved):

PR #874, opened 2024-11-11
Fix `joined=False` functionality

Swatinem:
This functionality was temporarily broken due to parallel processing, and this restores that functionality.

Additionally, this also logs any time this flag is triggered to Sentry to assist in figuring out who is actually using this.

### DIFF
--- a/services/processing/merging.py
+++ b/services/processing/merging.py
@@ -1,4 +1,5 @@
 import functools
+import logging
 from decimal import Decimal
 
 import sentry_sdk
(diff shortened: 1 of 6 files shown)


giovanni-guidini (approved):
I can't see anything in suspicious in the code, but no idea about the `joined` logic

Swatinem:
To be fair, I don’t think this is a widely used feature, and I also don’t think its particularly useful.

It pretty much means that followup uploads will fully override existing coverage values (for those files they do cover).

As the order of uploads is pretty much non-deterministic, I don’t think this works out well in practice.

PR #881, opened 2024-11-12
Fix notifiers assuming a `Report` exists

Swatinem:
The various Notifier classes assume that a `Report` always exists, which was not the case previously, and they would fail in various ways due to `report` being `None`.

Instead of actually fixing all of the usages deep within the notifier code, this just makes sure that an (empty) `Report` indeed exists.

This is suboptimal, as we should really fix the wrong assumptions in the code. However doing so would be quite some work, and this seems like a quicker quickfix.

---

fixes [WORKER-5ZB](https://codecov.sentry.io/issues/4370613101/)
fixes [WORKER-PNW](https://codecov.sentry.io/issues/6017331725/)
should help with https://github.com/codecov/engineering-team/issues/2365

### DIFF
--- a/services/comparison/types.py
+++ b/services/comparison/types.py
@@ -1,7 +1,7 @@
 from dataclasses import dataclass
 from typing import Optional, TypedDict
 
-from shared.reports.resources import Report
+from shared.reports.readonly import ReadOnlyReport
 from shared.yaml import UserYaml
 
 from database.models import Commit
@@ -11,7 +11,7 @@ from services.repository import EnrichedPull
 @dataclass
 class FullCommit(object):
     commit: Commit
-    report: Report
+    report: ReadOnlyReport
 
 
 class ReportUploadedCount(TypedDict):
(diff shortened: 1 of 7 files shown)


services/comparison/types.py
```diff
@@ -39,6 +39,11 @@ class Comparison(object):
     enriched_pull: EnrichedPull
     current_yaml: Optional[UserYaml] = None
 
+    # FIXME: The functions down below do not make sense given that we assume
+    # a `FullCommit` and its `report` to always exist.
```
adrian-codecov, line 43: Hmmm, will changing this to a ReadOnlyReport cause issues then? What if we did `Report | ReadOnlyReport` and do checks in the fns below to do `if report and ...`? Although not sure if it goes beyond the scope of this PR

Swatinem: Changing this to `ReadOnlyReport` is fixing wrong type annotations, as both the caller and downstream code already work with `ReadOnlyReport`.

The problem here is actually:
- Every caller always sets the `FullCommit`, so `self.head` won’t ever be `None`.
- On the other hand though, `FullCommit.report` could be `None` contrary to the declared type.
- Probably a bunch of downstream code was not properly checking `has_head_report`, and just assuming that `head.report` is never `None`, which is the part thats blowing up.

adrian-codecov: Ahhhhh I see I see. Hmmmm yeah if there's a way we can spot/document where downstream code assumes head.report to be not None, that would help, but beyond the scope of this ticket. Is there a future issue to address this?

adrian-codecov (approved):

services/comparison/types.py
Swatinem: I created https://github.com/codecov/engineering-team/issues/2917 to revisit this in the future.

PR #885, opened 2024-11-12
Clean up outdated task documentation

michelletran-codecov:
Some of these tasks are outdated. It is probably better for us to just look at the tasks directly to get a sense of what tasks are available. Also removed some documentation that will get outdated quite easily.

### Legal Boilerplate


adrian-codecov (approved):
Optional: leaving a comment that points people to the task folder. "You can see see all the tasks and what they do here"

PR #886, opened 2024-11-12
fix: MIME type for HTML emails

spalmurray-codecov:
Our emails were adding an extra `text/` to the `text/html` MIME type on our emails, causing only the text content to display in gmail web client (though mailhog handled this fine interestingly).

Also adds "Codecov" as the display name for our emails.

### DIFF
--- a/tasks/tests/integration/test_send_email_task.py
+++ b/tasks/tests/integration/test_send_email_task.py
@@ -79,7 +79,7 @@ class TestSendEmailTask:
             "",
         ]
         assert mail_body[7:-1] == [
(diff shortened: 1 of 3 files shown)


tasks/tests/integration/test_send_email_task.py
```diff
@@ -79,7 +79,7 @@ def test_send_email_integration(
             "",
         ]
         assert mail_body[7:-1] == [
-            'Content-Type: text/text/html; charset="utf-8"',
```
ajay-sentry, line 82: buh

ajay-sentry (approved):

PR #896, opened 2024-11-15
Try using `uv` for worker

Swatinem:
I tried taking inspiration from https://hynek.me/articles/docker-uv/, but hit some snags along the way.

I would need to dig a bit deeper learning how the docker setup fully functions.


docker-compose.yml
```diff
@@ -43,7 +42,6 @@ services:
         tmpfs:
           size: 2048M
   redis:
-    image: redis:6-alpine
```
michelletran-codecov, line 46: We use Redis 6.x in production, so let's stick with 6 here too.

PR #899, opened 2024-11-18
Clean up unused `minio.periodic_callback_ms` field

Swatinem:


giovanni-guidini (approved):
dunno what that is, but if it's not used 🤷

PR #900, opened 2024-11-18
Do not trigger `Notify` twice from `TestResultsFinisher`

Swatinem:
This also fixes up the return value in case the test results comment was updates.

Plus, it does a little formatting and typing cleanup.


joseph-sentry (approved):
good catch!

PR #901, opened 2024-11-18
fix: add some sentry emits to debug json error

joseph-sentry:
i want to find out if the reason the json decode errors are occuring is because we are re-processing uploads that have already been processed before. I also don't want to hard crash if we run into a json decode error

### DIFF
--- a/tasks/test_results_processor.py
+++ b/tasks/test_results_processor.py
@@ -371,7 +371,7 @@ class TestResultsProcessorTask(BaseCodecovTask, name=test_results_processor_task
         with metrics.timer("test_results.processor.process_individual_arg"):
             arg_processing_result: TestResultsProcessingResult = (
                 self.process_individual_arg(
-                    upload_obj, upload_obj.report.commit.repository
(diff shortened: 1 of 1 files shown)


tasks/test_results_processor.py
```diff
@@ -429,7 +430,13 @@ def process_individual_arg(
         archive_service = ArchiveService(repository)
 
         payload_bytes = archive_service.read_file(upload.storage_path)
-        data = json.loads(payload_bytes)
+        try:
+            data = json.loads(payload_bytes)
+        except json.JSONDecodeError as e:
```
adrian-codecov, line 435: Could this be the only exception json.loads gets?

Swatinem (approved):
for both these, I would use `capture_exception`, as you do have an exception object present.

tasks/test_results_processor.py
joseph-sentry: maybe not, but it's the only one we care about in this case

PR #911, opened 2024-11-26
fix: various commit update errors

joseph-sentry:
- it's possible for branch.authors to be null, so check for that
- it's possible for commit.timestamp to be a str, so check and convert it if it is
- it's possible for commit.pullid to be None, so handle that
- it's possible for commit.branch to be None, so handle that

### DIFF
--- a/tasks/commit_update.py
+++ b/tasks/commit_update.py
@@ -117,8 +117,6 @@ class CommitUpdateTask(BaseCodecovTask, name=commit_update_task_name):
                     ):
                         branch.head = commit.commitid
 
-            db_session.flush()
-
         except RepositoryWithoutValidBotError:
             log.warning(
                 "Unable to reach git provider because repo doesn't have a valid bot",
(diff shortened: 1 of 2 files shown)


tasks/commit_update.py
```diff
@@ -44,67 +45,79 @@ def run_impl(
...
+
+                    if (
+                        previous_branch_head is None
+                        or previous_branch_head.deleted == True
+                        or previous_branch_head.timestamp < commit.timestamp
+                    ):
+                        branch.head = commit.commitid
 
-            db_session.flush()
+                db_session.flush()
```
tony-codecov, line 120: This indent means only run `db_session.flush()` if `commit.branch is not None`, is it intended?

tony-codecov (approved):

tasks/commit_update.py
joseph-sentry: honestly the db_session.flush() is a bit redundant here, it happens at the end of the task and after `run_impl` runs we commit anyways so it's safe to remove since it achieves nothing

PR #917, opened 2024-11-28
fix(ta): handle random task retry better

joseph-sentry:
the previous fix for this was not enough. what's currently happening is that we're processing the files, marking the upload as processed then we crash with none of the parsed testruns being persisted to the database

we need to make sure that we persist the processed tests to the database before marking an upload as processed and then we are safe to rewrite it as readable.

so the order for a successful parsing is:
process it -> write testruns to database -> upload.state = process -> db commit -> safe to rewrite

a failed parsing would be:
process it -> upload.state = failed -> db commit -> safe to rewrite

we never want the rewrite or the upload state to be updated before we persist whatever processed testruns we need to persist to the database

### DIFF
--- a/tasks/test_results_processor.py
+++ b/tasks/test_results_processor.py
@@ -99,6 +99,7 @@ class PytestName:
 class TestResultsProcessingResult:
     network_files: list[str] | None
     parsing_results: list[ParsingInfo]
+    report_contents: list[ReadableFile]
 
 
 class TestResultsProcessorTask(BaseCodecovTask, name=test_results_processor_task_name):
@@ -128,6 +129,12 @@ class TestResultsProcessorTask(BaseCodecovTask, name=test_results_processor_task
             .all()
         )
         flaky_test_set = {flake.testid for flake in repo_flakes}
+        repository = (
+            db_session.query(Repository)
+            .filter(Repository.repoid == int(repoid))
+            .first()
+        )
+        archive_service = ArchiveService(repository)
 
         # process each report session's test information
         for arguments in arguments_list:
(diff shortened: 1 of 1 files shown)


tasks/test_results_processor.py
```diff
@@ -433,6 +471,15 @@ def process_individual_upload(
...
         )
 
+        upload_obj.state = "processed"
+        db_session.commit()
+
+        readable_report = self.rewrite_readable(
+            arg_processing_result.network_files,
+            arg_processing_result.report_contents,
+        )
+        archive_service.write_file(upload_obj.storage_path, readable_report.getvalue())
```
adrian-codecov, line 481: Didn't realize you couldn't change states before committing 😮.

Since I've been reading the coverage processor code lately, what are your thoughts on saving intermediate states of these in redis and having the finisher save them in bulk to the DB? Chatting w/ Arpad, it seemed to be to help hammering the DB as frequently, idk if this could benefit from that? Food for thought

adrian-codecov (approved):

tasks/test_results_processor.py
joseph-sentry: > saving intermediate states of these in redis and having the finisher save them in bulk to the DB
probably a good idea honestly, writes to redis is probably cheaper than a bunch of transactions with updates 
writing to the same keys

Swatinem (approved):
lgtm.

I believe the root cause of this is that we do not have transactions across different services (DB, Storage in this case, but Redis counts as well), and that these things are not idempotent.

Another solution would be to just be able to parse "readable" reports as well, without rewriting those yet again. Similar to how coverage processing does it.

PR #921, opened 2024-12-02
Handle Test/Flag inserts in the DB

Swatinem:
This eagerly inserts all the test / flags combinations, and uses a `on_conflict_do_nothing` to avoid duplicates.

Previously, this would resolve/fetch *all* the tests that have *any* repo flag, and deduplicate those on the python side.

That code was very slow, as the underlying query was extremely slow for unknown reasons. I also believe the code was potentially buggy, as it queried for *all* the tests with repo flags, not filtering for the specific flags we were interested in. Though it was probably fine, as the flags are part of the deterministic `test_id`.

In addition, this queries for the `flag_id` directly, turning it into a `set`, instead of using a `map` and a lookup for each test.

---

Fun fact, this function accounts for ~40% of all time in profiling:
Bildschirmfoto 2024-12-02 um 13 34 17

### DIFF
--- a/tasks/test_results_processor.py
+++ b/tasks/test_results_processor.py
@@ -40,35 +40,18 @@ class ReadableFile:
     contents: bytes
 
 
-def get_existing_flag_bridges(
-    db_session: Session, flag_ids: list[int]
-) -> dict[str, TestFlagBridge]:
-    existing_flag_bridges = (
-        db_session.query(TestFlagBridge)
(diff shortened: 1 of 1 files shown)


joseph-sentry (approved):
i'm cool with getting this in and measuring how it improves the perf, i do think the actual long term solution is the one that you mentioned previously: instead of writing to the db in each processor task we should be writing some intermediate result (to redis?) and merging them in the finisher and writing to the db there

PR #924, opened 2024-12-02
Cleanup and optimize `TestResultsFinisher`

Swatinem:
This cleans up the code a bit, and removes usage of statsd metrics.

Some performance improvements here include avoiding querying the related `Upload` of failed tests multiple times, as well as fetching redundant pull request information in the `Notifier`.

Another optimization was related to querying per-outcome aggregations first, before querying for failures specifically. This avoids a N+1 query fetching all the related `Test`s.

---

image

### DIFF
--- a/tasks/test_results_finisher.py
+++ b/tasks/test_results_finisher.py
@@ -5,13 +5,11 @@ from typing import Any
 from asgiref.sync import async_to_sync
 from shared.yaml import UserYaml
 from sqlalchemy.orm import Session
-from test_results_parser import Outcome
 
 from app import celery_app
 from database.enums import FlakeSymptomType, ReportType, TestResultsProcessingError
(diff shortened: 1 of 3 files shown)


tasks/test_results_finisher.py
```diff
@@ -165,31 +148,19 @@ def process_impl_within_lock(
...
         if self.check_if_no_success(previous_result):
             # every processor errored, nothing to notify on
-            metrics.incr("test_results.finisher.failure.no_successful_processing")
-
             queue_notify = False
 
             # if error is None this whole process should be a noop
             if totals.error is not None:
                 # make an attempt to make test results comment
                 notifier = TestResultsNotifier(commit, commit_yaml, None)
```
joseph-sentry, line 158: i think you'd need to pass through a repo service here

joseph-sentry (changes requested):
just one fix needed i think

PR #926, opened 2024-12-02
Update bot if we see that the bot field has changed in the YAML

michelletran-codecov:
This will help us get rid of one of the repo triggers to update the bot/branch field ([here](https://github.com/codecov/codecov-api/blob/168d5f2b8f4f3081c9c4dd1e459ff0511ad33ab5/legacy_migrations/migrations/legacy_sql/main/triggers/repos.py#L13)). Branch is already updated, so we only need to update the bot field.

### Legal Boilerplate

### DIFF
--- a/services/yaml/__init__.py
+++ b/services/yaml/__init__.py
@@ -1,5 +1,7 @@
 import logging
+from typing import Any, Mapping
 
+from shared.django_apps.utils.model_utils import get_ownerid_if_member
 from shared.torngit.exceptions import TorngitClientError, TorngitError
 from shared.validation.exceptions import InvalidYamlException
 from shared.yaml import UserYaml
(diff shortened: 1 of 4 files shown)


services/yaml/__init__.py
```diff
@@ -127,6 +131,22 @@ def save_repo_yaml_to_database_if_needed(current_commit, new_yaml):
...
                 repository.branch = yaml_branch
+
+            maybe_update_repo_bot(new_yaml, repository)
             repository.yaml = new_yaml
             return True
+
     return False
+
+
+def maybe_update_repo_bot(
```
adrian-codecov, line 142: Oh thanks for this! I had done it [in API](https://github.com/codecov/codecov-api/blob/fcee05d457747c372dafa8dae3ef886da15d7eb4/codecov_auth/commands/owner/interactors/set_yaml_on_owner.py#L55) but I missed this being a thing in worker, so thanks!

Also, technically the trigger you pointed to no longer exists since we actually use the migrations that come from shared https://github.com/codecov/codecov-api/blob/fcee05d457747c372dafa8dae3ef886da15d7eb4/codecov/settings_base.py#L436 which I had deleted. We can technically get rid of the migrations folders for the apps in that settings file so it's less confusing. We haven't cause it was a "next step" in that whole model migration thing and never got to it :/

adrian-codecov (approved):

Swatinem (approved):

adrian-codecov (approved):

PR #930, opened 2024-12-02
add Account model to worker's sqlalchemy models representations

nora-codecov:
I think this is blocking the shared release, which I need in order to complete the monthly self-hosted release
This is how I would add the Account model - lmk what you think!


database/models/core.py
```diff
@@ -55,14 +66,10 @@ class Account(CodecovBaseModel, MixinBaseClass):
...
-
-class AccountsUsers(CodecovBaseModel):
-    __tablename__ = "accounts_users"
-    account_id = Column(types.BigInteger, ForeignKey("accounts.id"), primary_key=True)
-    user_id = Column(types.BigInteger, ForeignKey("users.id"), primary_key=True)
-    created_at = Column(types.DateTime, nullable=False, default=datetime.utcnow)
+    users = relationship(
+        "User", secondary="codecov_auth_accountsusers", back_populates="accounts"
+    )
+    organizations = relationship("Owner", back_populates="account")
```
RulaKhaled, line 72: We don't have orgs in account do we?

RulaKhaled (approved):
lgtm, left a quick comment

database/models/core.py
nora-codecov: I think this is how you represent a many-to-many in sqlalchemy, from [these docs](https://docs.sqlalchemy.org/en/20/orm/basic_relationships.html#setting-bi-directional-many-to-many).  You have to explicitly set the `related_name` field, so it looks different from django but is the same thing

michelletran-codecov (approved):

PR #932, opened 2024-12-03
Slight `ParsedRawReport` cleanup

Swatinem:
Removes a bunch of has/get methods in favor of accessing fields directly. Removes the `env` field completely.
Avoids a `BytesIO` indirection in the renamed `as_readable` method.

### DIFF
--- a/services/report/parser/legacy.py
+++ b/services/report/parser/legacy.py
@@ -3,6 +3,7 @@ from io import BytesIO
 
 import sentry_sdk
 
+from services.path_fixer.fixpaths import clean_toc
 from services.report.parser.types import LegacyParsedRawReport, ParsedUploadedReportFile
 
 
@@ -67,7 +68,6 @@ class LegacyReportParser(object):
 
         This function takes the proper steps to find all the relevant sections of a report:
             - toc: the 'network', list of files present on this report
-            - env: the envvars the user set on the upload
             - uploaded_files: the actual report files
             - report_fixes: the report fixes some languages need
 
@@ -99,8 +99,8 @@ class LegacyReportParser(object):
                     while i_start < i_end and raw_report[i_start] in whitespaces:
                         i_start += 1
                 yield {
-                    "contents": raw_report[i_start:i_end],
                     "filename": filename,
+                    "contents": raw_report[i_start:i_end],
                     "footer": separator,
                 }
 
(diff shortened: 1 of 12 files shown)


services/report/parser/types.py
```diff
@@ -41,41 +38,31 @@ class ParsedRawReport(object):
...
-                buffer.write(f"{file}\n".encode("utf-8"))
-            buffer.write(b"<<<<<< network\n\n")
-        for file in self.uploaded_files:
-            buffer.write(f"# path={file.filename}\n".encode("utf-8"))
-            buffer.write(file.contents)
-            buffer.write(b"\n<<<<<< EOF\n\n")
-        buffer.seek(0)
-        return buffer
+
+class LegacyParsedRawReport(ParsedRawReport):
```
giovanni-guidini, line 57: This doesn't read a `as_readable` method?

services/report/parser/legacy.py
```diff
@@ -110,32 +111,27 @@ def parse_raw_report_from_bytes(self, raw_report: bytes) -> LegacyParsedRawRepor
...
-                            filename=sect.get("filename"),
-                            file_contents=sect["contents"],
-                        )
+                    file = ParsedUploadedReportFile(
+                        filename=sect["filename"], file_contents=sect["contents"]
                     )
+                    uploaded_files.append(file)
         return LegacyParsedRawReport(
-            toc=toc_section,
-            env=env_section,
```
giovanni-guidini, line 138: Will we have to update documentation by removing this?
(Not against the removal, just trying to be helpful)

giovanni-guidini (approved):
Well LGTM, but truth is I don't have too much context on this part of the code.

I left some questions, but they seem nit-level

services/report/parser/types.py
Swatinem: Indeed, the only usage of this is here:
https://github.com/codecov/worker/blob/6ba6bba83d4d4ea93a491f1c902c741b3124d7ab/services/processing/processing.py#L100-L102

services/report/parser/legacy.py
Swatinem: which documentation do you mean?

giovanni-guidini: haha good point. I looked for docs mentioning this "env" options in the uploads but found none.

The only place you'd find is running `--help` on the `do-upload` command of the CLI honestly

PR #933, opened 2024-12-03
refactor(ta proc): into functions and add tracing

joseph-sentry:
refactors the test results processing task into functions and add the explicit sentry tracing decorator to ensure they're measured in profiles

### DIFF
--- a/tasks/test_results_processor.py
+++ b/tasks/test_results_processor.py
@@ -54,6 +54,66 @@ def get_repo_flag_ids(db_session: Session, repoid: int, flags: list[str]) -> set
     )
 
 
+def create_daily_totals(
+    daily_totals: dict,
+    test_id: str,
+    repoid: int,
+    duration_seconds: float,
+    outcome: Outcome,
+    branch: str,
+    commitid: str,
+    flaky_test_set: set[str],
+):
+    daily_totals[test_id] = {
+        "test_id": test_id,
+        "repoid": repoid,
+        "last_duration_seconds": duration_seconds,
+        "avg_duration_seconds": duration_seconds,
+        "pass_count": 1 if outcome == str(Outcome.Pass) else 0,
+        "fail_count": 1
+        if outcome == str(Outcome.Failure) or outcome == str(Outcome.Error)
+        else 0,
+        "skip_count": 1 if outcome == str(Outcome.Skip) else 0,
+        "flaky_fail_count": 1
+        if test_id in flaky_test_set
+        and (outcome == str(Outcome.Failure) or outcome == str(Outcome.Error))
+        else 0,
+        "branch": branch,
+        "date": date.today(),
+        "latest_run": datetime.now(),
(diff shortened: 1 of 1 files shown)


services/test_results.py
```diff
@@ -105,6 +106,7 @@ def fetch_repo_flags(self, db_session, repoid):
         self.flag_dict = {flag.flag_name: flag for flag in existing_flags_on_repo}
 
 
+@sentry_sdk.trace
```
Swatinem, line 109: I don’t think tracing this makes any sense.

tasks/test_results_processor.py
```diff
@@ -179,156 +180,219 @@ def _bulk_write_tests_to_db(
...
-                    "framework": test_insert.excluded.framework,
-                    "computed_name": test_insert.excluded.computed_name,
-                    "filename": test_insert.excluded.filename,
-                },
-            )
-            db_session.execute(insert_on_conflict_do_update)
-            db_session.commit()
+            self.save_tests(db_session, sorted_tests)
 
         log.info("Upserted tests to database", extra=dict(upload_id=upload_id))
```
Swatinem, line 208: all these logs here and below should probably be nested in the  `if` they belong to.

tasks/test_results_processor.py
```diff
@@ -179,156 +180,219 @@ def _bulk_write_tests_to_db(
...
-            )
-            db_session.execute(insert_test_instances)
-            db_session.commit()
+            self.save_test_instances(db_session, test_instance_data)
 
         log.info("Inserted test instances to database", extra=dict(upload_id=upload_id))
 
+    @sentry_sdk.trace
+    def create_daily_totals(
+        self,
```
Swatinem, line 236: You probably don’t want to trace this either, as its a very simple function, and you call this thousands of times.

Another thing to note here is that you might want to make this a free function, as it does not depend on `self`.

tasks/test_results_processor.py
```diff
@@ -179,156 +180,219 @@ def _bulk_write_tests_to_db(
...
+            "date": date.today(),
+            "latest_run": datetime.now(),
+            "commits_where_fail": [commitid]
+            if (outcome == str(Outcome.Failure) or outcome == str(Outcome.Error))
+            else [],
+        }
+
+    @sentry_sdk.trace
+    def update_daily_totals(
+        self,
```
Swatinem, line 270: same here.

Swatinem (commented):
you should definitely revert the changes to trace those simple operations that are called in a loop and would thus lead to thousands of spans. probably creating the tracing span is more expensive than the underlying operation.

I also don’t think you need to wrap the sql statements in traces either, as you get automatic traces for the sql already.

Otherwise you can try and look at the traces, and if it turns out not to be readable anymore, or if the additional spans don’t add any value, you can revert them.

Swatinem (approved):

PR #942, opened 2024-12-04
fix(flakes): process uploads only once

joseph-sentry:
we want to process test instances related to an upload only once. If an upload has already been taken into account we don't want to take them into account again.

we are still only processing test instances that are related to code that is on the main branch since this task is only being run when:
- we receive a sync pulls task for the PR associated with this commit (merge existing test instances)
- we process test results for this commit that come from the main branch (we receive new relevant test instances)

This PR then refactors the process flakes task to use bulk create and update for general efficiency and also commits the transaction after processing a single upload since that should be smallest unit of work that must get done

### DIFF
--- a/tasks/process_flakes.py
+++ b/tasks/process_flakes.py
@@ -35,23 +35,23 @@ class ProcessFlakesTask(BaseCodecovTask, name=process_flakes_task_name):
         **kwargs: Any,
     ):
         """
-        this task wants to iterate through uploads for a given commit that have yet to be
-        "flake processed"
+        This task wants to iterate through uploads for a given commit that have yet to be
+        "flake processed".
 
-        for each of those uploads it wants to iterate through its test instances and
+        For each of those uploads it wants to iterate through its test instances and
         update existing flakes' count, recent_passes_count, fail_count, and end_date fields
-        depending on whether the test instance passed or failed
+        depending on whether the test instance passed or failed.
 
-        for each upload it wants to keep track of newly created flakes and keep those in a separate
+        For each upload it wants to keep track of newly created flakes and keep those in a separate
         collection than the existing flakes, so at the end it can bulk create the new flakes and
-        bulk update the existing flakes
+        bulk update the existing flakes.
 
-        it also wants to increment the flaky_fail_count of the relevant DailyTestRollup when it creates
-        a new flake so it keeps track of those changes and bulk updates those as well
+        It also wants to increment the flaky_fail_count of the relevant DailyTestRollup when it creates
+        a new flake so it keeps track of those changes and bulk updates those as well.
 
-        when it's done with an upload it merges the new flakes dictionary into the existing flakes dictionary
+        When it's done with an upload it merges the new flakes dictionary into the existing flakes dictionary
         and then clears the new flakes dictionary so the following upload considers the flakes created during the previous
-        iteration as existing
+        iteration as existing.
         """
         log.info(
             "Received process flakes task",
--- a/tasks/process_flakes.py
+++ b/tasks/process_flakes.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Any
 
 from django.db import transaction as django_transaction
 from django.db.models import Q
@@ -12,13 +13,10 @@ from shared.django_apps.reports.models import (
 )
 
 from app import celery_app
-from helpers.metrics import metrics
 from tasks.base import BaseCodecovTask
 
 log = logging.getLogger(__name__)
 
-
-FlakeDict = dict[tuple[str, int], Flake]
 
 FLAKE_EXPIRY_COUNT = 30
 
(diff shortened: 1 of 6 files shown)


tasks/process_flakes.py
```diff
@@ -26,27 +33,36 @@ def run_impl(
...
+            ReportSession.objects.select_related("report", "report__commit")
+            .filter(
+                report__report_type=CommitReport.ReportType.TEST_RESULTS.value,
+                report__commit__commitid=commit_id,
+                state="processed",
+            )
+            .all()
+        )
+
+        print(f"uploads: {uploads}")
```
Swatinem, line 55:
```suggestion
```

tasks/process_flakes.py
```diff
@@ -71,28 +87,27 @@ def run_impl(
                                     upserted_flake.reduced_error_id,
                                 )
                             ] = upserted_flake
+                upload.state = "flake_processed"
```
Swatinem, line 90: being hit by this problem myself, I’m a bit surprised that `upload.state` is not an enum that rejects new variants. unless `flake_processed` already exists as one of those variants?

tasks/process_flakes.py
```diff
@@ -26,27 +33,36 @@ def run_impl(
...
     ):
         repo_id = int(repo_id)
         log.info(
             "Received process flakes task",
-            extra=dict(repoid=repo_id, commit=commit_id_list),
+            extra=dict(repoid=repo_id, commit=commit_id),
         )
 
+        uploads = (
+            ReportSession.objects.select_related("report", "report__commit")
```
Swatinem, line 46: I don’t think you need the `select_related` here.  That is only used speed up accessors like `upload.report` by pre-fetching that data. Since you are only filtering/selecting by those related tables, but not actually accessing that data, you don’t need the `select_related`.

Swatinem (commented):
changing task arguments has to be a 3-step process:
1. teach the task to accept the *new* argument in a forward-compatible way in addition to the *old* ones
2. after that is fully rolled out, switch tasks to be emitted with only the new arguments
3. after that is fully rolled out, you can remove all the backwards compatible code dealing with the *old* arguments

The reason here is that we do not have atomic deploys/rollouts.
Tasks with *old* arguments might have been scheduled being picked up by a deploy expecting *new* arguments.
Also some parts of the pipeline might already be deployed scheduling tasks with *new* arguments, but other workers processing those tasks still expect the *old* arguments.

I also got that wrong a couple of times myself, so its not that critical, but you could potentially lose/error tasks in that overlapping time window.

tasks/process_flakes.py
joseph-sentry: upload.state is a char field so there's no enforcement whatsoever.

i think this field could be arbitrary since we're sharing the concept of "upload" for all three products that use them, perhaps we should reorganize the Upload to only contain information related to the storage of the raw uploaded file and keep track of the processing state of each type of upload in dedicated tables for each type of upload with enums and enforcement? 

for now i'd just like to arbitrarily set the state and encode the logic to handle each state in the app and clean up this state stuff later

tasks/process_flakes.py
```diff
@@ -23,76 +28,120 @@ class ProcessFlakesTask(BaseCodecovTask, name=process_flakes_task_name):
...
+
+        rollups_to_update: list[DailyTestRollup] = []
+
+        flaky_tests = list(curr_flakes.keys())
+
+        for upload in uploads:
+            test_instances = get_test_instances(upload, flaky_tests)
+            for test_instance in test_instances:
+                if test_instance.outcome == TestInstance.Outcome.PASS.value:
+                    flake = {**new_flakes, **curr_flakes}.get(test_instance.test_id)
```
Swatinem, line 78: you are merging these two dicts into an intermediate dict you throw away immediately. in a loop for every test instance, which is a bad idea for performance.

tasks/process_flakes.py
```diff
@@ -23,76 +28,124 @@ class ProcessFlakesTask(BaseCodecovTask, name=process_flakes_task_name):
...
-        commit_id_list,
-        branch,
-        **kwargs,
+        repo_id: int,
+        commit_id: str,
+        **kwargs: Any,
     ):
-        repo_id = int(repo_id)
+        """
+        this task wants to iterate through uploads for a given commit that have yet to be
```
michelletran-codecov, line 38: nit: Capitalization of sentences (this and below)

```suggestion
        This task wants to iterate through uploads for a given commit that have yet to be
```

tasks/process_flakes.py
```diff
@@ -101,78 +154,73 @@ def get_test_instances(
...
+    else:
+        log.warning(
+            "Could not find rollup when trying to update its flaky fail count",
+            extra=dict(
+                repoid=repo_id,
+                testid=test_instance.test_id,
+                branch=test_instance.branch,
+                date=test_instance.created_at.date(),
+            ),
+        )
```
michelletran-codecov, line 193: Do we have to worry about creating a new roll-up if it doesn't exist here?

michelletran-codecov (approved):
Generally LGTM. Just a question and nit about grammar.

tasks/process_flakes.py
joseph-sentry: no, because we assume it exists, i guess this should really be a `sentry_sdk.capture_exception` because i wanted to know if this is even a possibility, i don't think it has come up in the logs before, so it's probably safe to remove soon

PR #943, opened 2024-12-04
fix: don't cache rollups if branch none

joseph-sentry:


Swatinem (approved):

Swatinem:
The change makes sense, I just wonder how this can happen in reality? Under which circumstances will there be no branch at all?

joseph-sentry:
> The change makes sense, I just wonder how this can happen in reality? Under which circumstances will there be no branch at all?

it's possible that it's something about not being able to fetch the branch from the github api

PR #944, opened 2024-12-05
deps: update shared (use new minio storage service)

joseph-sentry:
- handle bundle analysis changes: BundleChange requires bundle_name
  passed as kwarg now
- remove service/storage.py as its no longer needed
- change a bunch of imports from using `from shared.storage import ...`
  to `import ...` so that the mock_storage fixture just works


JerrySentry (approved):
thanks for the bundle changes :P

services/bundle_analysis/report.py
```diff
@@ -234,7 +234,7 @@ def process_upload(
         """
         commit_report: CommitReport = upload.report
         repo_hash = ArchiveService.get_archive_hash(commit_report.commit.repository)
-        storage_service = get_storage_client()
+        storage_service = shared.storage.get_appropriate_storage_service(commit.repoid)
         bundle_loader = BundleAnalysisReportLoader(storage_service, repo_hash)
```
Swatinem, line 238: I’m wondering why this is not using `ArchiveService` directly? As that is pretty much just `storage_service + repo_hash`

tasks/tests/unit/test_cache_test_rollups_redis.py
```diff
@@ -16,7 +16,7 @@ def test_cache_test_rollups(self, mock_storage, transactional_db):
         repo = RepositoryFactory()
 
         redis = get_redis_connection()
-        storage_service = get_storage_client()
+        storage_service = shared.storage.get_appropriate_storage_service()
```
Swatinem, line 19: I wonder whether its even possible to always pass the `repo_id` to all the calls to `get_appropriate_storage_service`?

In this very example, you are passing the `repo_id` in the code, but not in the test, so these are out of sync. For tests, this does not matter at all. But for other parts of the codebase, I believe it might matter quite a bit if the reader and writer side are out of sync and get a different storage service because of it.

Swatinem (approved):
lgtm, but please take another look whether we consistently pass `repo_id` to `get_appropriate_storage_service`.

PR #947, opened 2024-12-06
add FlareCleanupTask

nora-codecov:
New strategy for managing `flare` storage

depends on https://github.com/codecov/shared/pull/450

When a `pull` is no longer `OPEN`, leave the `flare` in order to reduce locks and wait time.
During low traffic hours (4am UTC?), run this task to clear it out of our database and Archive storage.
So we won't pay for maintaining the `flare` field but also won't slow down other `pull` actions by checking or clearing the field 🎉

### DIFF
--- a/celery_config.py
+++ b/celery_config.py
@@ -10,7 +10,7 @@ from celery.schedules import crontab
 from shared.celery_config import (
     BaseCeleryConfig,
     brolly_stats_rollup_task_name,
-    flare_cleanup_task_name,
+    # flare_cleanup_task_name,
     gh_app_webhook_check_task_name,
     health_check_task_name,
     profiling_finding_task_name,
@@ -89,19 +89,18 @@ def _beat_schedule():
         },
         "trial_expiration_cron": {
             "task": trial_expiration_cron_task_name,
-            # 4 UTC is 12am EDT
-            "schedule": crontab(minute="0", hour="4"),
+            "schedule": crontab(minute="0", hour="4"),  # 4 UTC is 12am EDT
             "kwargs": {
                 "cron_task_generation_time_iso": BeatLazyFunc(get_utc_now_as_iso_format)
             },
         },
-        "flare_cleanup": {
-            "task": flare_cleanup_task_name,
-            "schedule": crontab(minute="0", hour="4"),  # every day, 4am UTC (8pm PT)
-            "kwargs": {
-                "cron_task_generation_time_iso": BeatLazyFunc(get_utc_now_as_iso_format)
-            },
-        },
+        # "flare_cleanup": {
+        #     "task": flare_cleanup_task_name,
+        #     "schedule": crontab(minute="0", hour="5"),  # every day, 5am UTC (10pm PDT)
+        #     "kwargs": {
+        #         "cron_task_generation_time_iso": BeatLazyFunc(get_utc_now_as_iso_format)
+        #     },
+        # },
     }
 
     if get_config("setup", "find_uncollected_profilings", "enabled", default=True):
--- a/celery_config.py
+++ b/celery_config.py
@@ -10,6 +10,7 @@ from celery.schedules import crontab
 from shared.celery_config import (
     BaseCeleryConfig,
     brolly_stats_rollup_task_name,
+    flare_cleanup_task_name,
     gh_app_webhook_check_task_name,
     health_check_task_name,
     profiling_finding_task_name,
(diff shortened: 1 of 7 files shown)


tasks/flare_cleanup.py
```diff
@@ -0,0 +1,63 @@
...
+
+        # clear in Archive
+        non_open_pulls_with_flare_in_archive = non_open_pulls.filter(
+            _flare_storage_path__isnull=False
+        ).select_related("repository")
+        log.info(
+            f"FlareCleanupTask will clear {non_open_pulls_with_flare_in_archive.count()} Archive flares"
+        )
+        # single query, loads all pulls and repos in qset into memory, deletes file in Archive 1 by 1
+        for pull in non_open_pulls_with_flare_in_archive:
```
Swatinem, line 50: In particular, the "load all pulls … into memory" part can be potentially dangerous as it might overwhelm the task.

This should use a limit query and work in batches.

tasks/flare_cleanup.py
```diff
@@ -0,0 +1,63 @@
...
+            archive_service = ArchiveService(repository=pull.repository)
+            archive_service.delete_file(pull._flare_storage_path)
```
Swatinem, line 52: I believe it might be worth grouping these by `repository(.id)` on the python side, to deduplicate creation of per-repo `ArchiveService`. Plus, I believe there is a `delete_files` method to do batch deletions as well. Though I believe that depending on the storage backend, that will still delete each file individually.

tasks/flare_cleanup.py
```diff
@@ -0,0 +1,63 @@
...
+        # single query, objs are not loaded into memory, does not call .save(), does not refresh updatestamp
+        n_updated = non_open_pulls_with_flare_in_db.update(_flare=None)
+        log.info(f"FlareCleanupTask cleared {n_updated} _flares")
+
+        # clear in Archive
+        non_open_pulls_with_flare_in_archive = non_open_pulls.filter(
+            _flare_storage_path__isnull=False
+        ).select_related("repository")
+        log.info(
+            f"FlareCleanupTask will clear {non_open_pulls_with_flare_in_archive.count()} Archive flares"
```
Swatinem, line 47: I believe this `.count()` is a query on its own, which seems a bit unreasonable just for a log line?

tasks/tests/unit/test_flare_cleanup.py
```diff
@@ -0,0 +1,127 @@
...
+            FlareCleanupTask.get_min_seconds_interval_between_executions(),
+            int,
+        )
+        assert FlareCleanupTask.get_min_seconds_interval_between_executions() > 17000
+
+    def test_successful_run(self, transactional_db, mocker):
+        mock_logs = mocker.patch("logging.Logger.info")
+        mock_archive_service = mocker.patch(
+            "shared.django_apps.utils.model_utils.ArchiveService"
+        )
```
Swatinem, line 22: Instead of mocking the `ArchiveService`, you could use the `mock_storage` fixture (I believe that should be hooked up to the shared archive service).

That way, you can actually store files (in memory), and assert that those files are truely being deleted from the storage.

celery_config.py
```diff
@@ -88,12 +89,18 @@ def _beat_schedule():
...
                 "cron_task_generation_time_iso": BeatLazyFunc(get_utc_now_as_iso_format)
             },
         },
+        # "flare_cleanup": {
+        #     "task": flare_cleanup_task_name,
+        #     "schedule": crontab(minute="0", hour="5"),  # every day, 5am UTC (10pm PDT)
+        #     "kwargs": {
+        #         "cron_task_generation_time_iso": BeatLazyFunc(get_utc_now_as_iso_format)
+        #     },
+        # },
```
nora-codecov, line 103: Going to test manually before I add this to the nightly cron jobs

tasks/flare_cleanup.py
nora-codecov: added!

tasks/flare_cleanup.py
nora-codecov: i made 2 changes here:
- removed the need for a Repository in order to do the delete in ArchiveService, so I'm no longer using the repo of the pull at all
- the the bulk_delete function hadn't been added to ArchiveService, added it so now I can use it, and you're correct that it does still delete each file individually, so I wait for that operation to complete before updating the field in the db.

michelletran-codecov (approved):
Generally LGTM

PR #949, opened 2024-12-06
build: upload test results first

joseph-sentry:
this is so if there are failures and issues with uploading coverage at least the PR comment for test results will be accurate.

It makes sense for all users to upload their test results first, then their coverage

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -232,8 +232,8 @@ test_env.run_integration:
 	docker-compose exec worker make test.integration
 
 test_env.upload:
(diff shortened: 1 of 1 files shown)


giovanni-guidini (approved):

giovanni-guidini:
> It makes sense for all users to upload their test results first, then their coverage

I totally agree. Also if it's a requirement in our system we should _at least_ make it clear in the docs (in case it's not already there, probably is) to avoid customers with the same issues as we have

PR #954, opened 2024-12-10
fix: test results processor return type

joseph-sentry:
there's a sentry bug currently happening because the finisher is not handling one of the possible return values of the processor

so i'm generally changing this to make it clearer:
- a processor either succeeds or fails, and this is represented by the return value being a boolean
  - a success means that the processor successfully parsed at least one file in any of the raw uploads it was processing
  - else, fail
- since the finisher is called in a chord with the processor it will receive a list of such booleans, if there is any success in this list of results then there is some valid test result data
- otherwise, we can just fail the finisher and make an error comment and try to notify coverage since there may be some valid coverage data

### DIFF
--- a/tasks/test_results_finisher.py
+++ b/tasks/test_results_finisher.py
@@ -53,7 +53,7 @@ class TestResultsFinisherTask(BaseCodecovTask, name=test_results_finisher_task_n
     def run_impl(
         self,
         db_session: Session,
-        chord_result: dict[str, Any],
+        chord_result: list[bool],
         *,
         repoid: int,
         commitid: str,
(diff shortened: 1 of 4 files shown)


tasks/test_results_finisher.py
```diff
@@ -149,7 +149,7 @@ def process_impl_within_lock(
             db_session.add(totals)
             db_session.flush()
 
-        if self.check_if_no_success(previous_result):
+        if not self.check_if_any_success(previous_result):
```
Swatinem, line 152:
```suggestion
        if not any(previous_result):
```

The fn is now so simple you don’t need it anymore.

Swatinem (approved):
lgtm, though this should probably be split into 3 different PRs/deploys:
- make the finisher forward compatible with the new signature of the processor
- change the processor to emit the new result type
- remove all the compatibility code

PR #955, opened 2024-12-11
Upgrade to Python 3.13

tony-codecov:
Upgrade to Python 3.13 image

### Legal Boilerplate


docker/Dockerfile.requirements
```diff
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1.4
-ARG PYTHON_IMAGE=python:3.12-slim-bookworm
+ARG PYTHON_IMAGE=python:3.13.1-slim-bookworm
```
Swatinem, line 2:
```suggestion
ARG PYTHON_IMAGE=python:3.13-slim-bookworm
```

Not sure if we want to make patch version updates explicit, or rather just always pick the latest?

Swatinem (approved):

docker/Dockerfile.requirements
tony-codecov: Talked with Trent about this and seems like `3.13` to always pick latest is the way to go! Thanks for bringing this up

PR #956, opened 2024-12-12
feat: don't send autoactivate pr comment if owner autoactivate is off

giovanni-guidini:
"I think, for now, we just shouldn't send this PR comment (but still send statuses!) in the event that auto activate is disabled on the org repo settings."

### DIFF
--- a/services/notification/notifiers/comment/conditions.py
+++ b/services/notification/notifiers/comment/conditions.py
@@ -27,12 +27,14 @@ class NotifyCondition(ABC):
 
     failure_explanation: str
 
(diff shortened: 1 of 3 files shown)


services/notification/notifiers/comment/conditions.py
```diff
@@ -150,6 +162,7 @@ def _check_uncovered_patch(comparison: ComparisonProxy) -> bool:
             0.01
         )
 
+    @staticmethod
```
adrian-codecov, line 165: Nice

PR #959, opened 2024-12-16
feat(ta): new prev result format in finisher

joseph-sentry:
this commit modifies the test results finisher to handle the new format
of previous_results that the processor will be outputting in the future

it also adds some typing to identify both the type of the old format and
the type of the new format

we want to go from a list[list[dict[str, str] | list[Any]]] to
list[bool] where the only key in the dict of the first type is:
"successful": bool

because of the rollout strategy we have to start with this change that
will deploy code to the finisher that can handle both formats then
we will push the change to the processor to start producing the new
format and finally we will remove the code that handles the old format
from the finisher

### DIFF
--- a/tasks/test_results_finisher.py
+++ b/tasks/test_results_finisher.py
@@ -127,14 +127,14 @@ class TestResultsFinisherTask(BaseCodecovTask, name=test_results_finisher_task_n
         assert commit, "commit not found"
         repo = commit.repository
 
-        if should_do_flaky_detection(repo, commit_yaml) and (
-            commit.merged is True or commit.branch == repo.branch
-        ):
-            self.app.tasks[process_flakes_task_name].apply_async(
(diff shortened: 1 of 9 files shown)


tasks/test_results_finisher.py
```diff
@@ -340,12 +344,17 @@ def get_flaky_tests(
...
-        )
+
+def check_if_any_success(
+    chord_result: list[bool] | list[list[TAProcessorResult | list[Any]]],
+) -> bool:
+    for result in chord_result:
+        if isinstance(result, list):
+            if any(isinstance(r, dict) and r["successful"] for r in result):
+                return True
+        elif isinstance(result, bool) and result:
```
adrian-codecov, line 355: The `and result` seem redundant here 👀

adrian-codecov (approved):

PR #963, opened 2024-12-17
feat: put the process flakes task behind a lock

joseph-sentry:
we want to reduce database lock contention so we are making it so only one process flakes task will be processing uploads for a specific repo at any time

the idea behind this is similar to what we do with the uploads task where there's a lock around the task and a list that contains the upload ids to process, the difference here is that right now we don't have access to the uploads that we want to process in the tasks that enqueu the process flakes task so instead, we will simply set a key in redis in place of pushing to a list. That key being set means that there is possibly more uploads to pull from the db to process flakes for


tasks/process_flakes.py
```diff
@@ -52,89 +54,113 @@ def run_impl(
...
+            log.warning("Unable to acquire process flakeslock for key %s.", lock_name)
+            return {"successful": False}
+
+        return {"successful": True}
+
+
+def run_impl_within_lock(
+    repo_id: int,
+    commit_id: str,
+):
```
Swatinem, line 86: I love that you moved this into its own function independent of the celery Task abstraction. Can you just give this a meaningful name then? (And possibly as a followup, also move this to a different file from the task)

Swatinem (approved):

PR #973, opened 2024-12-19
Implement storage- and relationship-aware cleanup

Swatinem:
This implements a generic framework for "delete anything referencing this object".

In order to do so, this builds a relationship graph based on Django models/metadata, augmented with some manually defined relationships which are not encoded in django models.

This graph is then processed in topological sort order, meaning that it will delete leaf models first, before advancing to other models, thus avoiding cascading deletes on the database side.

It is also possible to hook up custom code for specific models. This has been used to properly delete any kind of file in storage that is being referenced by these models.
Previous code has existed to delete all the "repo files", which did so by first enumerating *all* the files in storage matching a "per repository" naming pattern. Even though that code might have worked for smaller repositories, I assume that it would not have worked properly for larger repositories with a huge amount of files.

The new storage deletion code is using batches and a thread pool to delete files instead, and it also covers file name patterns (and different storage buckets in the case of BA) that do not match the "repo files" pattern.

---

This new generic deletion framework is being hooked up, replacing the existing `FlushRepo` and `DeleteOwner` tasks.
The tasks are build in such a way that they can time out and be rerun, and they will just pick up where they left off (going over the same topological sort order), assuming that the whole task is not being run in a transaction. (I hope not, but given how many problems we have with DB locking and transactions, who knows 🤷🏻‍♂️ )

---

This is work towards https://github.com/codecov/engineering-team/issues/1127


services/cleanup/models.py
```diff
@@ -0,0 +1,199 @@
...
+MANUAL_QUERY_CHUNKSIZE = 5_000
+DELETE_FILES_BATCHSIZE = 50
+
+
+def cleanup_files_batched(
+    context: CleanupContext, buckets_paths: dict[str, list[str]]
+) -> int:
+    cleaned_files = 0
+
+    # TODO: maybe reuse the executor across calls?
```
giovanni-guidini, line 30: We certainly stand to gainfrom not having to initialize and tear down the pool every time.

But it seems that the calls to `cleanup_files_batched` themselves are not concurrent. Is that on purpose? Like we are now it seems that at any given time there would always be at most 1 pool running anyway

services/cleanup/models.py
```diff
@@ -0,0 +1,199 @@
...
+    cleaned_files = 0
+    repo_hashes: dict[int, str] = {}
+
+    while True:
+        reports = coverage_reports[:MANUAL_QUERY_CHUNKSIZE]
+        if len(reports) == 0:
+            break
+
+        buckets_paths: dict[str, list[str]] = defaultdict(list)
+        for (
```
giovanni-guidini, line 109: [very nit] maybe a `NamedTuple` would make the code more ergonomic here (but less performant for sure)

services/cleanup/models.py
```diff
@@ -0,0 +1,199 @@
...
+from shared.django_apps.core.models import Commit, Pull
+from shared.django_apps.profiling.models import ProfilingUpload
+from shared.django_apps.reports.models import CommitReport, ReportDetails
+from shared.django_apps.reports.models import ReportSession as Upload
+from shared.django_apps.staticanalysis.models import StaticAnalysisSingleFileSnapshot
+
+from services.archive import ArchiveService, MinioEndpoints
+from services.cleanup.utils import CleanupContext, CleanupResult
+
+MANUAL_QUERY_CHUNKSIZE = 5_000
```
giovanni-guidini, line 21: I didn't know you could write numbers like this... TIL, thank you

services/cleanup/owner.py
```diff
@@ -0,0 +1,54 @@
...
+    return summary
+
+
+# TODO: maybe turn this into a `MANUAL_CLEANUP`?
+def clear_owner_references(owner_id: int):
+    """
+    This clears the `ownerid` from various DB arrays where it is being referenced.
+    """
+
+    OwnerProfile.objects.filter(default_org=owner_id).update(default_org=None)
```
giovanni-guidini, line 33: [curious] Do these update queries get commited immediately after running?

services/cleanup/repository.py
```diff
@@ -0,0 +1,89 @@
...
+            return (False, owner_id)
+
+        # We mark the repository as "scheduled for deletion" by setting the `deleted`
+        # flag, moving it to a new shadow owner, and clearing some tokens.
+        shadow_owner = Owner.objects.create(
+            # `Owner` is unique across service/id, and both are non-NULL,
+            # so we cannot duplicate the values just like that, so lets change up the `service_id`
+            # a bit. We need the `Repository.service_id` for further `ArchiveService` deletions.
+            service=owner_service,
+            service_id=f"☠️{owner_service_id}☠️",
```
giovanni-guidini, line 75: haha amazing

services/cleanup/repository.py
```diff
@@ -0,0 +1,89 @@
...
+    cleanup_started, owner_id = start_repo_cleanup(repo_id)
+
+    if cleanup_started:
+        log.info("Started Repository cleanup", extra={"repo_id": repo_id})
+    else:
+        log.info("Continuing Repository cleanup", extra={"repo_id": repo_id})
+
+    repo_query = Repository.objects.filter(repoid=repo_id)
+    summary = run_cleanup(repo_query)
+    Owner.objects.filter(ownerid=owner_id).delete()
```
giovanni-guidini, line 24: just confirming that this is the deletion of the shadow owner we created in `start_repo_cleanup` ?

tasks/delete_owner.py
```diff
@@ -1,118 +1,21 @@
...
 
 
 class DeleteOwnerTask(BaseCodecovTask, name=delete_owner_task_name):
-    """
-    Delete an owner and their data:
-    - Repo archive data for each of their owned repos
-    - Owner entry from db
-    - Cascading deletes of repos, pulls, and branches for the owner
-    """
+    acks_late = True  # retry the task when the worker dies for whatever reason
```
giovanni-guidini, line 14: It won't retry if the task ends with an Exception... just saying, if that is the expectation in "whatever"

giovanni-guidini (commented):
Lot's to unpack here... I like this in general. I think it's pretty cool and the code looks quite sophisticated.

Just so I understand this better, we currently have 2 entry points to use as "root" nodes when traversing the graph of dependencies: the repo or the owner. Is that correct?

services/cleanup/owner.py
Swatinem: Yes, they should.
- Django "runs" queries such as these immediately
- I’m a bit unsure about our usage of transactions though. I would hope that we do not run this task in a transaction, as that would make this non-retryable.

services/cleanup/repository.py
Swatinem: yes it is. That function will detect based on the data whether the "shadow owner" was already created, and will return the ID of that one.

services/cleanup/models.py
Swatinem: I don’t fully understand? Would you prefer to not destruct the tuple into variables right away, but rather access them through a `NamedTuple` object? I’m not sure that would make things more readable 🤔

services/cleanup/models.py
Swatinem: Indeed, it is running only with one pool at a time.

The regular deletions are still running serially in topological sort order. So it does not concurrently delete two unrelated models.

I was thinking about that a bit, but I thought it would be too complex to actually implement. Also reusing the thread pool would mean that I would have to introduce some kind of "waitpool" to make sure all the file deletes are complete before returning, which still keeping the pool alive.
I wanted to avoid that complexity for now.

Swatinem:
> Just so I understand this better, we currently have 2 entry points to use as "root" nodes when traversing the graph of dependencies: the repo or the owner. Is that correct?

Right now yes, those two need a bit of extra treatment, and they already have associated tasks for them.

But you can just put any arbitrary `Queryset` into that function, and it will do the right thing.
In the future, I would maybe move the whole `cleanup` module to `shared`, so its possible to use it within data migrations.
One thing I would like to use it for in the near-ish future is properly removing the `ReportDetails`/`files_array` model.

tasks/delete_owner.py
Swatinem: Do you have more details about the execution model of celery? I feel like after all this time I still don’t understand how it works and behaves under various circumstances.

For these tasks, I would want them to "resume" until they are completed.
This means "resuming" on timeouts (not sure if you can configure them to not have a timeout at all?).
It means resuming when the worker is killed for whatever reason. It also means resuming on any kind of random infrastructure exception.

Even for a fixable programmer error, I would like these tasks to always run to "successful" completion.

I’m not entirely sure how to achieve that with celery 🤷🏻‍♂️

giovanni-guidini (approved):
Thank you for the clarifications

giovanni-guidini (approved):
I particularly like the usage of `cleanup_context` in a contextmanager. Very nice 👍

PR #977, opened 2024-12-20
disable populating log_context based on commit_sha/commit_id

matt-codecov:
apparently populating the log context was slamming the commit table. so this PR stops using the commit table when populating log context

previously:
- if `commit_id` or `commit_sha`+`repo_id` were available, query the commits table to populate every field
- everything is populated this way, including `commit_id` (the DB PK for the commit)

now:
- `commit_id` and `commit_sha` will be set if available but not used to query anything else
- if `repo_id` is available, that will take care of populating the repo + owner

it's possible this will just move the problem to the repos table. if that is the case, it's solvable but will take a little elbow grease:
- transform `populate_from_sqlalchemy()` in `helpers/log_context.py` into `populate_log_context(kwargs) -> (Commit, Repository, Owner)` in `tasks/base.py`
- add a `run_impl()` stub to `BaseCodecovTask` which takes `owner`, `commit`, and `repo` kwargs, make all task subclasses accept the same (with default `None`) values
- in several followup PRs, update task subclasses to use the passed-in models instead of re-fetching them

follow-up issue for model reuse: https://github.com/codecov/engineering-team/issues/3140

### DIFF
--- a/helpers/log_context.py
+++ b/helpers/log_context.py
@@ -6,7 +6,7 @@ import sentry_sdk
 from sentry_sdk import get_current_span
 from shared.config import get_config
 
-from database.models.core import Commit, Owner, Repository
+from database.models.core import Owner, Repository
 
 log = logging.getLogger("log_context")
 
@@ -28,6 +28,11 @@ class LogContext:
     repo_name: str | None = None
     repo_id: int | None = None
     commit_sha: str | None = None
+
+    # TODO: begin populating this again or remove it
+    # we can populate this again if we reduce query load by passing the Commit,
+    # Owner, and Repository models we use to populate the log context into the
+    # various task implementations so they don't fetch the same models again
     commit_id: int | None = None
 
     checkpoints_data: dict = field(default_factory=lambda: {})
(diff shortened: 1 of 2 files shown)


helpers/log_context.py
```diff
@@ -58,48 +58,13 @@ def populate_from_sqlalchemy(self, dbsession):
...
-                    self.repo_name,
-                    self.owner_id,
-                    self.owner_username,
-                    self.owner_service,
-                    self.owner_plan,
-                ) = query.first()
-
-            # repo_id is enough to get repo and owner
-            elif self.repo_id:
+            if self.repo_id:
```
adrian-codecov, line 67: So the intent here is to reenables the log context without the commit?

matt-codecov: yes

the stated problem is that populating the log context slammed the commit table. turns out if we don't query the commit table the only thing we are likely to lose is the DB PK of the commit which we don't tend to use anyway. so this PR stops querying the commit table

this may just move the problem, and we should maybe do the followup described in the PR description anyway

adrian-codecov (approved):

PR #981, opened 2024-12-26
chore: Update Stripe and pin API to latest version

ajay-sentry:
This PR bumps the stripe version from 9.6 -> 11.4.1 and changes the API version from 4/24 -> 12/24.

Just wanted to get ahead of API changes since the last time we did this it was a huge pain... (6 years of updates all at once lol): https://github.com/codecov/codecov-api/pull/569

[Changelog](https://github.com/stripe/stripe-python/releases?page=1)

[Stripe docs](https://docs.stripe.com/upgrades#2024-12-18.acacia)

Only breaking changes were from v11 where we now have a default 2 retries and timeout changed from 2 -> 5 seconds for the stripe client. I figure that's probably fine and we can keep the new default. V10 didn't have any breaking changes we would have been affected by.

Relates to https://github.com/codecov/codecov-api/pull/1071

### Legal Boilerplate


nora-codecov (changes requested):
We are using `uv` instead of `pip-tools` for handling requirements in worker - that's why theres multiple unrelated packages being changed in the `.txt`

Instead of `pip-compile`, please run `uv pip compile requirements.in -o requirements.txt`, that should sort it out :)

nora-codecov (approved):

PR #989, opened 2025-01-03
3169 move notifications presteps to one task

adrian-codecov:
So, this PR stems from us having many notification conditions spread out in different tasks (finisher, notify) and this is my take for it. I decided to create a task, notification_orchestrator, that's meant to be a centralized place for doing checks that determine whether we'd notify or not, rather than doing it when we're already in the notify task (and expecting to notify). The idea is for any codecov product to tap into this orchestrator if they desire to do so, although this first pass is targeted mostly for the coverage case.

I would recommend solely looking at `notification_orchestrator.py` and giving feedback there. The changes in `upload_finisher.py` and `notify.py` task are mostly deletions - these are the changes that now belong in the orchestrator task. I have not yet tested this orchestrator task fully, so this is moreso a pass about it capturing the logic well and if it seems a reasonable/readable approach for notifying. 

I'm happy for this PR to just be adding the task, and in a future PR connecting it to the finisher/notify task (I'd modify this PR to revert the changes in those two files as well). 

I left some inline comments as questions and I'll leave some more as PR comments. Please let me know about this when you get a chance. After the 1st pass, I'll work on the tests as well.

### Legal Boilerplate


tasks/notification_orchestrator.py
```diff
@@ -0,0 +1,631 @@
...
+                    db_session,
+                    repoid=repoid,
+                    commitid=commitid,
+                    commit_yaml=commit_yaml,
+                    processing_results=processing_results,
+                    report_code=report_code,
+                    empty_upload=empty_upload,
+                    **kwargs,
+                )
+        # TODO: What should happen when there's a lock error? I think this should change to some retrying mechanism
```
adrian-codecov, line 103: I'm thinking when 2 finishers are done (presumably with different results), one picks up this lock and the other will wait. I suspect the 2nd one should be able to pick up this task

tasks/notification_orchestrator.py
```diff
@@ -0,0 +1,631 @@
...
+                    extra=dict(
+                        repoid=repoid,
+                        commitid=commitid,
+                        error_type=type(err),
+                        lock_acquired=lock_acquired,
+                    ),
+                ),
+            )
+            UploadFlow.log(UploadFlow.NOTIF_LOCK_ERROR)
+            return {
```
adrian-codecov, line 118: TODO for me: I should change this to `ShouldCallNotifyResponse`

tasks/notification_orchestrator.py
```diff
@@ -0,0 +1,631 @@
...
+        repoid = int(repoid)
+        commit = (
+            db_session.query(Commit)
+            .filter(Commit.repoid == repoid, Commit.commitid == commitid)
+            .first()
+        )
+        assert commit, "Commit not found in database."
+        commit_yaml = UserYaml(commit_yaml)
+
+        log_extra_dict = {
```
adrian-codecov, line 145: I pass around this log bw functions, but there might be a better strategy to do so, perhaps leveraging the context 👀

tasks/notification_orchestrator.py
```diff
@@ -0,0 +1,631 @@
...
+            }:
+                log.info(
+                    "Unable to start notifications. Retrying again later.",
+                    extra={
+                        **log_extra_dict,
+                        "reason": reason,
+                        "countdown": countdown,
+                        "max_retries": max_retries,
+                    },
+                )
```
adrian-codecov, line 204: We should add a UploadFlow log for "reattempting notification" or something

tasks/notification_orchestrator.py
```diff
@@ -0,0 +1,631 @@
...
+        commit: Commit,
+        commit_yaml: UserYaml,
+        processing_results: list[ProcessingResult],
+        log_extra_dict: dict,
+        report_code: str | None = None,
+    ) -> ShouldCallNotifyResponse:
+        # Defines all the logic to consider a notification
+        should_notify_check_functions = [
+            partial(
+                self.upload_processing_checks, processing_results=processing_results
```
adrian-codecov, line 235: These types of notifications could in theory happen before we even grab the lock, but leaving them here for now to be consistent with where the rest of the checks get done.

I also don't think the check order matters here, just as long as we make them, but I could be wrong

tasks/notify.py
```diff
@@ -201,249 +181,142 @@ def run_impl_within_lock(
                 "reason": "test_failures",
             }
 
+        # ASK: this installation/bot related logic impedes notification if unavailable, but it
+        # seems correct for it to be part of the notifier task, thoughts?
```
matt-codecov, line 185: i think the git provider app/token stuff is a tricky gray area

if you imagine a generic notification system, there are roughly three stages:
- deciding whether to post this kind of notification based on business logic
  - i would say app/token stuff doesn't belong in this stage
- computing the underlying data for this notification
  - some data (e.g. patch coverage) may require credentials to compute. so the coverage product for example has to decide whether to totally fail without credentials or just skip patch coverage
- sending the Slack template to Slack, the PR Comment template to the PR Comment, etc
  - the Slack notification doesn't care if we don't have git provider credentials, but the PR comment does

ideally, missing credentials would only skip the things that definitely require credentials and the rest of the system would continue working. and when credentials are missing, that's a misconfiguration on the part of the user and we should alert them to it in the UI but otherwise not treat it as an error

i think your NO task is implementing the first stage of the generic notification system, where app/token stuff shouldn't be a factor. the `NotifyTask` represents the second and third stages, and cleaning that up to work like i described is a separate matter

tasks/notification_orchestrator.py
matt-codecov: i also assume order doesn't matter here but i would not be shocked if it does :P

don't have to do it now, but it might be a good idea to make each of these functions `async` and then `asyncio.gather()` or `asyncio.as_completed()` them to run them concurrently. could speed up notifications a good amount

tasks/notification_orchestrator.py
```diff
@@ -0,0 +1,624 @@
...
+
+        # Creates repository_service and ci_results to use with yaml settings
+        try:
+            installation_name_to_use = get_installation_name_for_owner_for_task(
+                self.name, commit.repository.owner
+            )
+            # This is technically a different repository_service method than the one used in the notify task. I
+            # replaced it because the other method is a) deeply intertwined with the notify file and b) it seems
+            # to be extra specific just for the notification aspect of it, so this should suffice for other checks.
+            # Please look corroborate this isn't a red flag.
```
matt-codecov, line 368: is this a note to yourself or to reviewers?

tasks/notification_orchestrator.py
```diff
@@ -0,0 +1,624 @@
...
+                extra=log_extra_dict,
+            )
+            return ShouldCallNotifyResponse(
+                notification_result=ShouldCallNotifyResult.DO_NOT_NOTIFY,
+                reason="has_require_ci_to_pass_yaml_setting_and_no_ci_results",
+                message="Not scheduling notify because it's a local upload",
+            )
+
+        # Some check on CI skipping from some regex? Idk what this is
+        if regexp_ci_skip.search(commit.message or ""):
```
matt-codecov, line 528: https://github.com/codecov/codecov.io/commit/6d3cb2a89b687533b72bf47adeac85158ffd0f13

matt-codecov (commented):
i didn't 1:1 compare the moved logic to the new logic and there are still TODO comments + open comments form me so i'm not stamping yet, but overall looks great!

PR #990, opened 2025-01-04
add root_organization property to worker, clean up decoration service, change notification urls for GL

nora-codecov:
Decoration service was causing confusion for GitLab users. Cleaned it up and added some comments to clarify the current GL org strategy which is: 
- the root org owns the plan, so when dealing with a GL org's plan, look at the plan from the root org
- membership and activation through `plan_activated_user` also happens on the root org instead of a subgroup org
- if a notification has the `plan` or `member` url, give GL users the correct url (correct url is to root group, not subgroup)

more details on ticket https://github.com/codecov/engineering-team/issues/2710

### DIFF
--- a/database/models/core.py
+++ b/database/models/core.py
@@ -3,6 +3,7 @@ import string
 import uuid
 from datetime import datetime
 from functools import cached_property
+from typing import Optional
 
 from shared.plan.constants import PlanName
 from sqlalchemy import Column, ForeignKey, Index, UniqueConstraint, types
@@ -154,11 +155,11 @@ class Owner(CodecovBaseModel):
     )
 
     @property
-    def slug(self):
+    def slug(self: "Owner") -> str:
         return self.username
 
     @property
-    def root_organization(self):
+    def root_organization(self: "Owner") -> Optional["Owner"]:
         """
         Find the root organization of Gitlab OwnerOrg, by using the root_parent_service_id
         if it exists, otherwise iterating through the parents and cache it in root_parent_service_id
@@ -178,17 +179,19 @@ class Owner(CodecovBaseModel):
             db_session.commit()
         return root
 
-    def _get_owner_by_service_id(self, db_session, service_id):
+    def _get_owner_by_service_id(
+        self: "Owner", db_session: Session, service_id: str
+    ) -> "Owner":
         """
         Helper method to fetch an Owner by service_id.
         """
         return (
             db_session.query(Owner)
             .filter_by(service_id=service_id, service=self.service)
-            .one_or_none()
+            .one()
         )
 
-    def __repr__(self):
+    def __repr__(self: "Owner") -> str:
         return f"Owner<{self.ownerid}@service<{self.service}>>"
 
 
--- a/database/models/core.py
+++ b/database/models/core.py
@@ -105,6 +105,7 @@ class Owner(CodecovBaseModel):
     avatar_url = Column(types.Text, server_default=FetchedValue())
     updatestamp = Column(types.DateTime, server_default=FetchedValue())
     parent_service_id = Column(types.Text, server_default=FetchedValue())
+    root_parent_service_id = Column(types.Text, nullable=True)
     plan_provider = Column(types.Text, server_default=FetchedValue())
     trial_start_date = Column(types.DateTime, server_default=FetchedValue())
     trial_end_date = Column(types.DateTime, server_default=FetchedValue())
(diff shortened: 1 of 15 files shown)


database/models/core.py
```diff
@@ -156,6 +157,37 @@ class Owner(CodecovBaseModel):
...
+    def _get_owner_by_service_id(self, db_session, service_id):
+        """
+        Helper method to fetch an Owner by service_id.
+        """
+        return (
+            db_session.query(Owner)
+            .filter_by(service_id=service_id, service=self.service)
+            .one_or_none()
+        )
+
```
nora-codecov, line 190: this is not new - it's in shared, I'm just replicating it in worker.
I learned that in order to use shared's `PlanService`, which now calls `.root_organization`, sqlalchemy also needs to know about this property, so I had to add it.

database/models/core.py
```diff
@@ -156,6 +157,37 @@ class Owner(CodecovBaseModel):
...
+        db_session = self.get_db_session()
+        if self.root_parent_service_id:
+            return self._get_owner_by_service_id(
+                db_session, self.root_parent_service_id
+            )
+
+        root = None
+        if self.service == "gitlab" and self.parent_service_id:
+            root = self
+            while root.parent_service_id is not None:
```
giovanni-guidini, line 175: I think (but could be wrong) if `self._get_owner_by_service` can return `None` (from the `.one_or_none()` it seems possible, there are no typehints in the functions) then you will have a `NoneType has no property "parent_service_id"` error.

It would be good to add explicit typehints to the functions too

services/decoration.py
```diff
@@ -31,8 +29,8 @@ class DecorationDetails(object):
     decoration_type: Decoration
     reason: str
     should_attempt_author_auto_activation: bool = False
-    activation_org_ownerid: int = None
-    activation_author_ownerid: int = None
+    activation_org_ownerid: int | None = None
```
giovanni-guidini, line 32: 👍 ty

services/decoration.py
```diff
@@ -94,19 +92,12 @@ def determine_decoration_details(
 
         db_session = db_pull.get_db_session()
 
-        if org.service == "gitlab" and org.parent_service_id:
-            # need to get root group so we can check plan info
-            (gl_root_group,) = db_session.query(
-                func.public.get_gitlab_root_group(org.ownerid)
-            ).first()
+        # do not access plan directly - only through PlanService
+        org_plan = PlanService(current_org=org)
```
giovanni-guidini, line 96: The `PlanService` doesn't seem to go looking for the root org (looking at [main](https://github.com/codecov/shared/blob/9a61a94a70722971e56437defe5c31a2399928b0/shared/plan/service.py#L66))

I think I'm missing a code change... does this PR depend on an unmerged change to shared too?

giovanni-guidini: Ah yes, https://github.com/codecov/shared/pull/461 ok ok

database/models/core.py
adrian-codecov: Also, could we ever infinite loop because the parent_service_id is always None? Would we like to cap it after 10 iterations or something?

services/decoration.py
```diff
@@ -50,7 +48,7 @@ def determine_uploads_used(plan_service: PlanService) -> int:
 
 def determine_decoration_details(
     enriched_pull: EnrichedPull, empty_upload=None
-) -> dict:
+) -> DecorationDetails:
```
adrian-codecov, line 51: 🏆

services/urls.py
```diff
@@ -57,6 +58,16 @@ def get_dashboard_base_url() -> str:
         return configured_dashboard_url or "https://app.codecov.io"
 
 
+def _get_username_for_url(repository: Repository) -> str:
+    username = repository.owner.username
+    if repository.owner.service == Service.GITLAB.value:
```
adrian-codecov, line 63: Idk if we save service always as "gitlab" vs "GitLab" or something like that, might be worth lowercasing this just in case

database/models/core.py
nora-codecov: added typehints, and changed that query to `.one()`  @giovanni-guidini 

@adrian-codecov if `parent_service_id` is `None` the `while` breaks - I think you mean the opposite, it will infinite loop if it's never `None`. This condition would only happen if our data was bad and there was a set of `Owners` referencing back and forth at each other. We don't have anything to prevent this, but it really should never happen. We have GL webhooks that should keep this field updated with the state on GL. I think the risk is low :)

adrian-codecov (approved):

PR #998, opened 2025-01-09
Handle TA upload error

joseph-sentry:
- we want to persist upload errors in the ta_processor
- we want to surface those errors when we notify either TA or coverage


michelletran-codecov (approved):
Generally LGTM. Just a few nits.

PR #1004, opened 2025-01-11
clean up status mixins

nora-codecov:
I'm working on [this ticket](https://github.com/codecov/engineering-team/issues/1605) and [this ticket](https://github.com/codecov/engineering-team/issues/1626), which are changing the messaging coming out of this flow around `target` and `threshold`.
`StatusPatchMixin` and `StatusProjectMixin` are similar, but doing things differently, ie
`head_coverage + threshold >= target_coverage` vs `head_coverage >= (target_coverage - threshold)` 
No behavior/logic changes, just cleanup to make them more uniform and easier to understand.

### DIFF
--- a/services/notification/notifiers/mixins/status.py
+++ b/services/notification/notifiers/mixins/status.py
@@ -1,5 +1,6 @@
 import logging
 from decimal import Decimal, InvalidOperation
+from enum import Enum
 
 from services.comparison import ComparisonProxy, FilteredComparison
 from services.yaml.reader import round_number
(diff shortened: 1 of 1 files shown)


services/notification/notifiers/mixins/status.py
```diff
@@ -6,25 +6,31 @@
 
 log = logging.getLogger(__name__)
 
+STATE_SUCCESS = "success"
```
giovanni-guidini, line 9: [nit] I'd use an Enum for this

services/notification/notifiers/mixins/status.py
```diff
@@ -215,6 +233,7 @@ def _apply_adjust_base_behavior(
         quantized_base_adjusted_coverage = base_adjusted_coverage.quantize(
             Decimal("0.00000")
         )
+        ### shouldn't we use threshold here?
```
giovanni-guidini, line 236: This is actually a fair question. I'd ask product if we want to do that.

I don't think we should... the idea of this is that the coverage needs to be better somehow for the check to pass (because it's already failing). The exact "how" is "looking at only adds, are they more covered than before?". Having the threshold can open this up too much.

services/notification/notifiers/mixins/status.py
```diff
@@ -301,8 +321,37 @@ def get_project_status(
...
+        threshold = self.notifier_yaml_settings.get("threshold", "0.0")
+
+        try:
+            # check if user has erroneously added a % to this input and fix
+            threshold = Decimal(str(threshold).replace("%", ""))
+        except (InvalidOperation, TypeError, AttributeError):
+            threshold = Decimal("0.0")
+        return threshold
+
+    def _get_target(self, base_report_totals) -> Decimal:
```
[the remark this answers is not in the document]
giovanni-guidini: and there's no typehint on base_report_totals

giovanni-guidini (approved):
Nice change 👍 
Great initiative making the code easier to understand.

Left some nit comments here and there. The only one I'd say to definitely fix before merging is the incorrect typehint. The rest is up to you.

services/notification/notifiers/mixins/status.py
nora-codecov: ooohhh, i see what you mean - I wasn't remembering that at this point in code they have already failed on the `target` + `threshold` base vs head measure.

I'll reach out to product because my thought here is - if they have a `threshold`, shouldn't we always be doing calculations with the threshold? And when we give them a message back, shouldn't we use/include `threshold` so that they don't sometimes see the number that represents the target, and other times the number that represents adjusted target with threshold?

services/notification/notifiers/mixins/status.py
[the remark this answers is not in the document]
nora-codecov: ty for catching! I changed to tuple at the last minute and forgot to update the typing!

PR #1006, opened 2025-01-13
3210 batch uploads creation

adrian-codecov:
This is the more complete version of uploads, flags + measurements batch insertion. This consolidates all the efforts + changes related to improving N+1 issues. This change adds a fn to add uploads, flags + measurements when this is a v4 upload. All should be batched and improve lock + performance issues 

### Legal Boilerplate


services/tests/test_report.py
```diff
@@ -3727,10 +3727,6 @@ def test_create_report_upload(self, dbsession):
         assert res.totals is None
         assert res.upload_extras == {}
         assert res.upload_type == "uploaded"
-        assert len(res.flags) == 1
```
adrian-codecov, line 3730: This no longer gets evaluated by this test/method since we got rid of the override

tasks/upload.py
```diff
@@ -497,18 +504,116 @@ def run_impl_within_lock(
...
+    def possibly_insert_uploads_and_side_effects(
+        self,
+        db_session: Session,
+        upload_context: UploadContext,
+        commit: Commit,
+        commit_report: CommitReport,
+        report_service: ReportService,
+    ) -> list[UploadArguments]:
+        """
+        This method possibly batch inserts uploads, flags and user measurements.
```
giovanni-guidini, line 556: we know it "possibly" does something from the name. What are the conditions that the upload will or will not be inserted?

tasks/upload.py
```diff
@@ -526,42 +631,67 @@ def run_impl_within_lock(
...
+            measurements_list=measurements_list,
+            uploads_list=uploads_list,
+            upload_flag_map=upload_flag_map,
+        )
+
+    def _bulk_insert_flags(
+        self,
+        db_session: Session,
+        repoid: int,
+        upload_flag_map: dict[Upload, list | str | None] = {},
```
giovanni-guidini, line 645: using a mutable structure as default arg value is not a good idea. The same dict instance will be shared across all function calls.

It's better to make the default `None` and check for that, and then assign the real default of `{}` if needed.

```python
upload_flag_map: dict[Upload, list | str | None] | None = None,
...
if upload_flag_map is None:
    upload_flag_map = {}
```

tasks/upload.py
```diff
@@ -215,6 +215,13 @@ def _should_debounce_processing(upload_context: UploadContext) -> Optional[float
     return None
 
 
+class CreateUploadResponse(TypedDict):
```
giovanni-guidini, line 218: nice nice

tasks/upload.py
adrian-codecov: It's described in the sentence after that, https://github.com/codecov/worker/pull/1006/files#diff-7ea195247ce946074a2441630db2be4bda78feed427fc0a1f1bd170cfc84a7a7R557 😅, only happens for v4 uploads since cli uploads are created in api. Should I rephrase it a bit so it's clearer?

tasks/upload.py
adrian-codecov: How's the naming here? I was not _amazed_ by it 😂

tasks/upload.py
```diff
@@ -497,18 +504,116 @@ def run_impl_within_lock(
...
         created_at = timezone.now()
 
+        # List + helper mapping to track possible upload + flags to insert later
+        uploads_list: list[Upload] = []
+        upload_flag_map: dict[Upload, list | str | None] = {}
+
         for arguments in upload_context.arguments_list():
             arguments = upload_context.normalize_arguments(commit, arguments)
             if "upload_id" not in arguments:
                 upload = report_service.create_report_upload(arguments, commit_report)
```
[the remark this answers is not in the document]
adrian-codecov: Yeah you are right, I was doing that from a previous implementation.

I'd like to use the bulk_save but I need the upload's primary key for the "arguments["upload_id"] = upload.id_" line, so doing that's why we're doing the add/flush (and therefore the bulk_save is kinda redundant). I'll get rid of it for now, but can you think of a way to make it work? 

I've been giving it some thought and the closest thing Ive thought of is to skip the add/flush, do an arguments_upload mapping (without the primary key), do the bulk_save, then requery the db for the uploads we just created and add the upload_id to the arguments based on the mapping, then continue execution. But that requerying seems inefficient and I feel makes this whole thing a bit more complicated. 

I could just get rid of the bulk_save and still benefit from the other changes though

tasks/upload.py
```diff
@@ -215,6 +215,13 @@ def _should_debounce_processing(upload_context: UploadContext) -> Optional[float
     return None
 
 
+class CreateUploadResponse(TypedDict):
+    argument_list: list[UploadArguments] | list
+    measurements_list: list[UserMeasurement] | list
+    uploads_list: list[Upload] | list
+    upload_flag_map: dict[Upload, list | str | None] | dict
```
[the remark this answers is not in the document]
adrian-codecov: Yeah I also prefer dataclasses personally, just following the trend I see throughout the codebase to have typeddicts, but not strongly opposed to any. Will update the typing as well 👍

tasks/upload.py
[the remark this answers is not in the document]
Swatinem: another idea here would be, since getting the `arguments` without an `upload_id` is one of the legacy upload paths:
we could just patch that legacy upload path to create the `Upload` object and trigger a `PreProcess` to do carry-forwarding, then we can delete the code responsible for this here.
its not ideal either :-(

Swatinem:
I believe this could live in `services.processing`, maybe in a new file called `orchestration.py` or some such?

There is a bunch more code cleanup to be done here, related to some unification of how the uploads are being handled in API, like the `normalize_arguments` code can probably be removed completely by now, based on this comment: https://github.com/codecov/worker/blob/5634caa20df86206ba02524373613888f27585b7/services/processing/types.py#L24-L26

tasks/upload.py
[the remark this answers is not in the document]
adrian-codecov: Yeah this is another option. I think the counter of that is that we couldn't batch insert since it would be done at an upload level, that being said, we already do this with the CLI, so arguably another point of performance there.

To me though, it seems the responsibility to make an upload shouldn't be in API (although I get why we currently do it this way), instead done by an inexistent "upload service" that would come w/ the new services. I think worker is the closest thing to that atm, but yeah I preferred to do it here since most of the code was here.

Swatinem (approved):

PR #1007, opened 2025-01-16
Worker: Migrate to Plan / Tier Tables

ajay-sentry:
- [ ] This PR relies on this version of shared: https://github.com/codecov/shared/pull/479.

The main objective of this PR is to migrate all Backend Service logic to use the new plan and tiers tables instead of the Plan constants. You can read more about milestone 3 [here](https://www.notion.so/sentry/Streamlining-Plan-Creation-with-the-Plan-Configuration-Table-1428b10e4b5d808ca5b2d4f5b66493a1).

A core change introduced in this PR is the creation of mocks for plans and tiers, which are used in most test setup functions. For tests requiring usage in all nested functions, we use a fixture with autouse enabled. Additionally, we ensure django_db decorators are included to allow access to the mock database.

The rest of the changes should be fairly straightforward.

Closes https://github.com/codecov/engineering-team/issues/3253

### Legal Boilerplate


database/models/core.py
```diff
@@ -60,7 +60,9 @@ class Account(CodecovBaseModel, MixinBaseClassNoExternalID):
     name = Column(types.String(100), nullable=False, unique=True)
     is_active = Column(types.Boolean, nullable=False, default=True)
     plan = Column(
-        types.String(50), nullable=False, default=PlanName.BASIC_PLAN_NAME.value
+        types.String(50),
+        nullable=False,
+        default=PlanName.BASIC_PLAN_NAME.value,  # TODO: UPDATE WITH NEW FREE PLAN NAME
```
RulaKhaled, line 65: Good catch!

requirements.txt
```diff
@@ -1,5 +1,9 @@
-# This file was autogenerated by uv via the following command:
-#    uv pip compile requirements.in -o requirements.txt
+#
+# This file is autogenerated by pip-compile with Python 3.12
```
RulaKhaled, line 2: Hmm what's the reason behind not using uv here?

services/tests/test_decoration.py
```diff
@@ -491,12 +500,14 @@ def test_get_decoration_type_not_pr_plan(self, dbsession, mocker, enriched_pull)
         assert decoration_details.reason == "Org not on PR plan"
         assert decoration_details.should_attempt_author_auto_activation is False
 
+    @pytest.mark.django_db
+    # what is a users plan?
```
ajay-sentry, line 504: users plan is the one from GHM, which we actually will need to add to the plan table in DB as well now that I think about it

RulaKhaled: oohhh cool cool

tasks/upload.py
```diff
@@ -531,7 +531,6 @@ def run_impl_within_lock(
 
         # Bulk insert coverage measurements
         if measurements:
-            print("test - I am here!", measurements)
```
Swatinem, line 534: lol, I even saw this in production logs these days looking at the repo deletion logs :-D

tests/helpers.py
```diff
@@ -0,0 +1,176 @@
+from shared.django_apps.codecov_auth.models import BillingRate
+from shared.django_apps.codecov_auth.tests.factories import PlanFactory, TierFactory
+from shared.plan.constants import PlanName, PlanPrice, TierName
+
+
+def mock_all_plans_and_tiers():
```
Swatinem, line 6: you are calling this explicitly in a ton of tests.

How is this being handled in production? Were these things created manually, or using a data migration?

Maybe it makes sense to introduce these kinds of "constants" using a data migration in the first place.
Otherwise, is it possible to auto-load these fixtures in all the tests that are using the django DB somehow?
Its a bit annoying that you have to call this manually in all those places. Better to just "have this data around" within your tests.

Swatinem (commented):
Thanks for cleaning up all that print noise in test code (and in production code, lol). If its not too much effort, can you extract those changes as a separate PR?

Otherwise, my comments related to having this mock data around globally instead of pulling it in explicitly. As well as consuming these objects from the DB rather than by mocking function calls would be nice.
Mixing django/sqlalchemy is also something to be cautious about, as I believe it could cause massive problems.

tasks/upload.py
adrian-codecov: yeah this is my bad 😬 I'm getting rid of this in my other big PR

ajay-sentry: Should be fixed now with https://github.com/codecov/engineering-team/issues/3278

PR #1008, opened 2025-01-16
fix new ta tasks bugs

joseph-sentry:
- we were not formatting the name of the tests we were inserting correctly
- we were not getting all the relevant uploads in ta_finisher
- we were always trying to persist data to bigquery even when it was not configured


adrian-codecov (approved):
Is there a way to do local testing w/ bq? Curious to know the manual flow for this if there's anything

PR #1009, opened 2025-01-16
fix: persist has_failed status on upload

joseph-sentry:
we want to commit this change as soon as we know it in order to make sure the upload state is consistent


adrian-codecov (approved):
Are there tests for this?

Swatinem (approved):

PR #1012, opened 2025-01-16
remove testrun pyi file

joseph-sentry:


adrian-codecov (approved):

Swatinem (approved):
the reason being that the actual `.py` file exists, so there is no need to keep the `.pyi` file around?

joseph-sentry:
> the reason being that the actual .py file exists, so there is no need to keep the .pyi file around?

@Swatinem  the reason is that it was breaking the pre-commit linting (not the actual linting). This is meant to be a temporary change until i get this working with the pre-commit linting.

PR #1014, opened 2025-01-16
Failed check and status messaging in comment, part I

nora-codecov:
The goal of this epic is to communicate details about check or status failures in the other notifiers, like pr comment.

This is part 1, which is mostly new data structures and variables, and handling of notifications.

Right now we collect all notifications and execute them in any order, they are all independent and they don't know about each other. In this pr, I split notifications into 2 categories (checks/status notifications, or other notifications). Now the check/status notifications go first, because their outcome will impact the messaging in the other types of notifications.

The specific condition I will tackle next: add custom helper text for users that have specified a `target` in their `yaml`. If their check/status fails, we will add a message to the pr comment telling them that the check/status failed, with advice that mentions `target` and links them to the `target` docs. There are screenshots in the tix linked below.

What are check/status notifications? Check out [the best stack overflow answer I've ever seen](https://stackoverflow.com/questions/67919168/github-checks-api-vs-check-runs-vs-check-suites)

https://github.com/codecov/engineering-team/issues/1605
https://github.com/codecov/engineering-team/issues/1626
Epic here: https://github.com/codecov/engineering-team/issues/1133

### DIFF
--- a/services/notification/notifiers/mixins/status.py
+++ b/services/notification/notifiers/mixins/status.py
@@ -13,6 +13,9 @@ class StatusState(Enum):
     failure = "failure"
 
 
+custom_target_helper_text = "Your project check has failed because the patch coverage is below the target coverage. You can increase the patch coverage or adjust the [target](https://docs.codecov.com/docs/commit-status#target) coverage."
+
+
 class StatusPatchMixin(object):
     def _get_threshold(self) -> Decimal:
         """
@@ -29,7 +32,7 @@ class StatusPatchMixin(object):
 
     def _get_target(
         self, comparison: ComparisonProxy | FilteredComparison
-    ) -> Decimal | None:
+    ) -> tuple[Decimal | None, bool]:
         """
         Target can be configured by user, default is auto, which is the coverage level from the base report.
         Target will be None if no report is found to compare against.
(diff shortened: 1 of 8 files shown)


services/notification/notifiers/mixins/status.py
```diff
@@ -13,7 +14,35 @@ class StatusState(Enum):
...
+
+    state: Literal["success", "failure"]  # StatusState values
+    message: str
+    included_helper_text: dict[str, str]
+
+
+CUSTOM_TARGET_TEXT_PATCH_KEY = "custom_target_helper_text_patch"
+CUSTOM_TARGET_TEXT_PROJECT_KEY = "custom_target_helper_text_project"
+CUSTOM_TARGET_TEXT_VALUE = (
+    "Your {context} {notification_type} has failed because the patch coverage is below the target coverage. "
```
Swatinem, line 31: also looking at the mockup for this, is it reasonable to include the numbers in this message as well, like:
> … has failed because the patch coverage (X%) is below the target coverage (Y%).

Swatinem (approved):
I love the well defined return types now

services/notification/notifiers/mixins/status.py
nora-codecov: That's not in the design, but I think it makes sense - having the %s makes the message better, will add

PR #1023, opened 2025-01-22
Failed check and status messaging in comment, part II

nora-codecov:
when user has:
patch checks notifier OR patch status notifier
AND
comment notifier
AND
has defined a `target` in their codecov yml
AND 
the patch check/status fails
THEN
add some helper text to the comment notifier to point out that the check/status failed and that they can change their `target`.

https://github.com/codecov/engineering-team/issues/1626

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,6 +3,12 @@ codecov:
   notify:
     wait_for_ci: false
 
+coverage:
+  status:
+    patch:
+      default:
+        target: 99
+
 component_management:
   default_rules:
     statuses:
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,17 +3,11 @@ codecov:
   notify:
     wait_for_ci: false
 
-coverage:
-  status:
-    patch:
-      default:
-        target: 99
-
 component_management:
   default_rules:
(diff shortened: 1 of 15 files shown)


codecov.yml
```diff
@@ -7,7 +7,7 @@ component_management:
   default_rules:
     statuses:
       - type: project
-        target: auto
+        target: 99
```
giovanni-guidini, line 10: Was this for testing? Do we want to leave it like so?

services/notification/notifiers/mixins/message/__init__.py
```diff
@@ -85,10 +90,11 @@ def create_message(
         owner: Owner = repo.owner
 
         # Separate PR comment based on plan that can't/won't be tweaked by codecov.yml settings
-        if (
-            owner.plan == BillingPlan.team_monthly.value
-            or owner.plan == BillingPlan.team_yearly.value
-        ):
+        owner_plan = PlanService(owner)
```
giovanni-guidini, line 93: nice 👍

Swatinem (approved):

PR #1025, opened 2025-01-22
[Bundle Analysis] Display asset changes and large file changes in PR comments

JerrySentry:
- Add asset comparisons to bundle analysis PR comments
- Add list of files changed that contributed to asset changes

Looks like this:
Screenshot 2025-01-21 at 3 26 38 PM

### Legal Boilerplate


suejung-sentry (approved):
Looks great! Just a nit that it looks like the sigfigs in the example image sometimes says 100% and sometimes 100.0%. I guess I'd expect the deleted to say 100%

PR #1028, opened 2025-01-23
Simplify dependent subqueries when possible.

Swatinem:
The current implementation would still create subqueries of the form `foreign_pk IN (SELECT pk FROM table WHERE pk=123)`.

We can now detect these trivial lookups, and simplify those to `foreign_pk IN (123)`.

### DIFF
--- a/services/cleanup/relations.py
+++ b/services/cleanup/relations.py
@@ -3,6 +3,8 @@ from collections import defaultdict
 from graphlib import TopologicalSorter
 
 from django.db.models import Model, Q
+from django.db.models.expressions import Col, Expression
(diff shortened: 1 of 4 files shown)


michelletran-codecov (approved):
Generally LGTM. Do you have a feel of how often this optimization case is called and how much performance gains we get?

Swatinem:
It is called for every table depending transitively on `Repository` when doing repo deletion.
How much we gain is hard to judge. It is saving one "index only scan" on the repo table and a join. Not quite sure how much that gains within the DB, probably not a lot for simple queries, maybe a bit more for more complex ones.

PR #1030, opened 2025-01-24
Failed check and status messaging in comment, part III

nora-codecov:
This is the same as the last one (https://github.com/codecov/worker/pull/1023) but for *project* rather than *patch* status and checks.

when user has:
project checks notifier OR project status notifier
AND
comment notifier
AND
has defined a target in their codecov yml
AND
the project check/status fails
THEN
add some helper text to the comment notifier to point out that the check/status failed and that they can change their target.

https://github.com/codecov/engineering-team/issues/1605

### DIFF
--- a/services/notification/notifiers/mixins/status.py
+++ b/services/notification/notifiers/mixins/status.py
@@ -28,8 +28,8 @@ class StatusResult(TypedDict):
 CUSTOM_TARGET_TEXT_PATCH_KEY = "custom_target_helper_text_patch"
 CUSTOM_TARGET_TEXT_PROJECT_KEY = "custom_target_helper_text_project"
 CUSTOM_TARGET_TEXT_VALUE = (
-    "Your {context} {notification_type} has failed because the patch coverage ({coverage}%) is below the target coverage ({target}%). "
-    "You can increase the patch coverage or adjust the "
+    "Your {context} {notification_type} has failed because the {point_of_comparison} coverage ({coverage}%) is below the target coverage ({target}%). "
+    "You can increase the {point_of_comparison} coverage or adjust the "
     "[target](https://docs.codecov.com/docs/commit-status#target) coverage."
 )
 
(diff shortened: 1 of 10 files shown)


services/notification/notifiers/mixins/status.py
```diff
@@ -384,8 +386,14 @@ def get_project_status(
             # Possibly change status
             if removed_code_result:
                 removed_code_state, removed_code_message = removed_code_result
-                return removed_code_state, message + removed_code_message
-        return state, message
+                print(removed_code_state, removed_code_message)
```
giovanni-guidini, line 389: print

services/notification/notifiers/mixins/status.py
```diff
@@ -467,15 +482,23 @@ def _get_project_status(
             # use rounded numbers for messages
             target_rounded = round_number(self.current_yaml, target_coverage)
             message = f"{head_coverage_rounded}% (target {target_rounded}%)"
-            # TODO:
-            # helper_text = HELPER_TEXT_MAP[CUSTOM_TARGET_TEXT].format(
-            # context=self.context, notification_type=notification_type, coverage=head_coverage_rounded, target=target_rounded)
-            # included_helper_text[CUSTOM_TARGET_TEXT] = helper_text
-            return state, message
+            if state == StatusState.failure.value:
+                helper_text = HELPER_TEXT_MAP[CUSTOM_TARGET_TEXT_PROJECT_KEY].format(
```
giovanni-guidini, line 486: It looks like this HELPER_TEXT_MAP (see [here](https://github.com/codecov/worker/blob/e00410f72c1f9f47e5c0f577a25f00e1c530f7a6/services/notification/notifiers/mixins/status.py#L37)) points both the patch and project keys to the same helper text, and "patch" is hard coded in that text

giovanni-guidini (commented):
LGTM, just asking wanted to make sure the helper text map is up to date before approving.

I didn't see it in the diff 😅

services/notification/notifiers/mixins/status.py
nora-codecov: ha! good catch - thank you. fixing.

giovanni-guidini (approved):

PR #1032, opened 2025-01-24
Improve TADriver interface and implementations

joseph-sentry:
depends on: https://github.com/codecov/shared/pull/484

- improve the TADriver interface
- complete the BQDriver and PGDriver implementations
  - the PGDriver changes are mostly moving around existing code
- Introduce ta_utils which is going to replace services.test_results
  -  the reason it's being introduced is because I had to change the interface of TestResultsNotifier but we still want the old one around for the old test results pipeline
- add and use the test_analytics app from shared
- consume the config options in django_scaffold.settings instead of making various get_config calls

### DIFF
--- a/services/ta_utils.py
+++ b/services/ta_utils.py
@@ -1,0 +1,373 @@
+import logging
+from dataclasses import dataclass
+from hashlib import sha256
+
+from shared.plan.constants import FREE_PLAN_REPRESENTATIONS, TEAM_PLAN_REPRESENTATIONS
+from shared.yaml import UserYaml
+from sqlalchemy import desc, distinct, func
+from sqlalchemy.orm import Session, joinedload
+
+from database.models import (
+    Commit,
+    Repository,
+    TestInstance,
+    Upload,
+    UploadError,
+)
+from helpers.notifier import BaseNotifier
+from rollouts import FLAKY_TEST_DETECTION
+from services.license import requires_license
+from services.urls import get_members_url, get_test_analytics_url
+from services.yaml import read_yaml_field
+
+log = logging.getLogger(__name__)
+
+
+def generate_flags_hash(flag_names: list[str]) -> str:
+    return sha256((" ".join(sorted(flag_names))).encode("utf-8")).hexdigest()
+
+
+def generate_test_id(repoid, testsuite, name, flags_hash):
+    return sha256(
+        (" ".join([str(x) for x in [repoid, testsuite, name, flags_hash]])).encode(
+            "utf-8"
+        )
+    ).hexdigest()
+
+
+def latest_failures_for_commit(
+    db_session: Session, repo_id: int, commit_sha: str
+) -> list[TestInstance]:
+    """
+    This will result in a SQL query that looks something like this:
+
+    SELECT DISTINCT ON (rti.test_id) rti.id, ...
+    FROM reports_testinstance rti
+    JOIN reports_upload ru ON ru.id = rti.upload_id
+    LEFT OUTER JOIN reports_test rt ON rt.id = rti.test_id
+    WHERE ...
+    ORDER BY rti.test_id, ru.created_at DESC
+
+    The goal of this query is to return:
+    > the latest test instance failure for each unique test based on upload creation time
+
+    The `DISTINCT ON` test_id with the order by test_id, enforces that we are only fetching one test instance for each test.
+
+    The ordering by `upload.create_at DESC` enforces that we get the latest test instance for that unique test.
+    """
+
+    return (
+        db_session.query(TestInstance)
+        .join(TestInstance.upload)
+        .options(joinedload(TestInstance.test))
+        .filter(TestInstance.repoid == repo_id, TestInstance.commitid == commit_sha)
+        .filter(TestInstance.outcome.in_(["failure", "error"]))
+        .order_by(TestInstance.test_id)
+        .order_by(desc(Upload.created_at))
+        .distinct(TestInstance.test_id)
+        .all()
+    )
+
+
+def get_test_summary_for_commit(
+    db_session: Session, repo_id: int, commit_sha: str
+) -> dict[str, int]:
+    return dict(
+        db_session.query(
+            TestInstance.outcome, func.count(distinct(TestInstance.test_id))
+        )
+        .filter(TestInstance.repoid == repo_id, TestInstance.commitid == commit_sha)
+        .group_by(TestInstance.outcome)
+        .all()
+    )
+
+
+@dataclass
+class FlakeInfo:
+    failed: int
+    count: int
+
+
+@dataclass
+class TestFailure:
+    display_name: str
+    failure_message: str | None
+    duration_seconds: float
+    build_url: str | None
+    flake_info: FlakeInfo | None = None
+
+
+@dataclass
+class TestResultsNotificationPayload:
+    failed: int
+    passed: int
+    skipped: int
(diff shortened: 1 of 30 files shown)


ta_storage/tests/test_bq.py
```diff
@@ -1,485 +1,381 @@
...
-def populate_testruns_for_upload_testruns(dbsession, bq: BQDriver):
-    testruns = []
+    testruns_written = [
+        MessageToDict(
+            ta_testrun_pb2.TestRun.FromString(testrun_bytes),
+            preserving_proto_field_name=True,
+        )
+        for testrun_bytes in mock_bigquery_service.mock_calls[0][1][3]
+    ]
+    assert snapshot("json") == sorted(testruns_written, key=lambda x: x["name"])
```
Swatinem, line 146: you are asserting that the mocked function is being called with the arguments you put in above.
IMO the value of such tests is very low. All you are asserting here is that the serialization to protobuf is working correctly?

ta_storage/tests/test_bq.py
```diff
@@ -1,485 +1,381 @@
...
+    result = driver.pr_comment_agg("abc123")
+
+    mock_bigquery_service.query.assert_called_once()
+    query, params = mock_bigquery_service.query.call_args[0]
+    assert snapshot("txt") == query
+    assert params == [
+        ScalarQueryParameter("repoid", "INT64", 1),
+        ScalarQueryParameter("commit_sha", "STRING", "abc123"),
+    ]
+    assert snapshot("json") == result
```
Swatinem, line 170: here also, the snapshot just contains the `mock_bigquery_service.query.return_value`

ta_storage/bq.py
```diff
@@ -170,47 +226,58 @@ def __init__(self, flaky_test_set: set[bytes] | None = None):
             ),
         )
 
+    def get_live_flakes(self, repo_id: int) -> list[Flake]:
+        return [
+            flake
+            for flake in Flake.objects.filter(
```
Swatinem, line 232:
```suggestion
        return list(Flake.objects.filter(
```

ta_storage/tests/test_bq.py
```diff
@@ -1,485 +1,381 @@
...
-    bq.write_testruns(None, 3, "abcde", "feature_branch", upload, "pytest", testruns)
+    result = driver.pr_comment_agg("abc123")
+
+    mock_bigquery_service.query.assert_called_once()
+    query, params = mock_bigquery_service.query.call_args[0]
+    assert snapshot("txt") == query
+    assert params == [
+        ScalarQueryParameter("repoid", "INT64", 1),
+        ScalarQueryParameter("commit_sha", "STRING", "abc123"),
+    ]
```
Swatinem, line 169: In general, I doubt asserting the generated SQL here is providing any value.

You are building these queries using trivial string concatenation.

This is very different from the way I assert the generated queries in the deletion code, which are fully dynamically created by the ORM, where I have very little insight into how it works under the hood.

tasks/ta_processor.py
```diff
@@ -128,32 +129,43 @@ def process_individual_upload(
...
-                    upload,
-                    framework,
-                    testruns,
-                )
-
-                if bq_enabled:
+
+            pg = PGDriver(repoid, db_session, flaky_test_set)
+            if settings.BIGQUERY_WRITE_ENABLED:
+                bq = BQDriver(repoid)
```
Swatinem, line 135: instead of duplicating this code, how about taking advantage of the new driver interface you have introduced:

```
drivers = [pg, bq] if BQ_ENABLED else [pg] # or any kind of combination based on feature flags
for driver in drivers:
  driver.bulk_write_testruns(parsing_info)
```

Swatinem (commented):
there are a few simplifications possible here, otherwise this looks good.

I am quite unhappy with the tests though. We are once again relying on overly mocked "whitebox" tests. the only thing they are asserting is that you call the mocks with certain parameters, and that the function returns the mocked data unmodified. I don’t think these tests provide any kind of value :-(

ta_storage/bq.py
```diff
@@ -276,14 +364,119 @@ def testruns_for_upload(self, upload_id: int, test_ids: list[bytes]):
...
+        return [
+            {
+                "branch_name": result["branch_name"],
+                "timestamp": result["timestamp"],
+                "outcome": result["outcome"],
+                "test_id": result["test_id"],
+                "flags_hash": result["flags_hash"],
+            }
+            for result in query_result
+        ]
```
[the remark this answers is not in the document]
joseph-sentry: the reason for this was typing, since i wanted the return type to be a typed dict, but there was no guarantee on the result of query having those keys, i ended up doing this, which is no better than doing `cast(query_result, list[TypedDict])` tbh, but i do want it to be a TypedDict

services/ta_utils.py
```diff
@@ -0,0 +1,373 @@
...
+        if char == "`":
+            curr_backtick_count += 1
+        else:
+            curr_backtick_count = 0
+
+        if curr_backtick_count > max_backtick_count:
+            max_backtick_count = curr_backtick_count
+
+    backticks = "`" * (max_backtick_count + 1)
+    return f"{backticks}python\n{content}\n{backticks}"
```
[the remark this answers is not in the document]
joseph-sentry: this was a product / design decision, we just wanted to have _some_ highlighting in the failure message displayed in the comment, i can't remember if this decision was made before or after we had framework detection but either way i can bring this up to them

ta_storage/bq.py
```diff
@@ -326,3 +519,90 @@ def analytics(
...
+                (
+                    bytes(flake.test_id),
+                    bytes(flake.flags_id) if flake.flags_id else None,
+                ): FlakeInfo(
+                    failed=flake.fail_count,
+                    count=flake.count,
+                )
+                for flake in Flake.objects.raw(
+                    "SELECT * FROM flake WHERE repoid = %s AND (test_id, flags_id) IN %s AND end_date IS NULL AND count != recent_passes_count + fail_count",
+                    [self.repo_id, test_ids],
```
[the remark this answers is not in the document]
Swatinem: ohhh, so the problem here is that a `IN` query with a tuple is not expressible with the django query builder?

joseph-sentry:
> I am quite unhappy with the tests though. We are once again relying on overly mocked "whitebox" tests. the only thing they are asserting is that you call the mocks with certain parameters, and that the function returns the mocked data unmodified. I don’t think these tests provide any kind of value :-(

i think you're right. my idea was to ship this quickly and start validating in prod on our own repos and have minimal automated testing because the option of "find a way to connect to BQ in CI" sounded like it would take too long and the option of "use a third party emulator" sounded like a bad idea. the goal of the tests right now is not to validate that anything works but that changes that devs make in the future don't break the existing behaviour. I thought this might have some value but i think you're right that we want tests that validate that things work and that these tests are basically just here to make sure that devs review their changes to this code through the snapshots.

what i can do right now is use https://github.com/goccy/bigquery-emulator but i still don't think it's a good idea to use a third party emulator in the long term and that it we would be best if we find a way to connect to some dev deployment of BQ in the CI at some point in the future

Swatinem:
similar to what I proposed a while back to test *real* GCS access, we can use secrets in CI to run this with a proper test project in GCP, and just skip these tests when the credentials are not available either locally or for forks in CI (though maybe still fail in CI on missing credentials anyway)

I think that would be a very reasonable thing to do

PR #1033, opened 2025-01-24
Complete the new TA pipeline implementations

joseph-sentry:
- creates the new cache analytics task
- creates the new ta process flakes task
- TA finisher uses new TADriver interface and ta_utils
- TA finisher queues up new tasks

### DIFF
--- a/tasks/ta_finisher.py
+++ b/tasks/ta_finisher.py
@@ -13,12 +13,12 @@ from database.enums import FlakeSymptomType, ReportType
 from database.models import (
     Commit,
     CommitReport,
-    Flake,
     Repository,
     TestResultReportTotals,
     Upload,
     UploadError,
 )
+from django_scaffold.settings import BIGQUERY_READ_ENABLED
 from helpers.checkpoint_logger.flows import TestResultsFlow
 from helpers.notifier import NotifierResult
 from helpers.string import EscapeEnum, Replacement, StringEscaper, shorten_file_paths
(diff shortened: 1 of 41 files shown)


tasks/ta_finisher.py
```diff
@@ -121,64 +120,42 @@ def get_totals(
...
-    matching_flakes = list(
-        db_session.query(Flake)
-        .filter(
-            Flake.repoid == repoid,
-            Flake.testid.in_(failure_test_ids),
-            Flake.end_date.is_(None),
-            Flake.count != (Flake.recent_passes_count + Flake.fail_count),
+
+def get_postgres_test_data(
+    db_session: Session, repo: Repository, commit_sha: str, commit_yaml: UserYaml
```
Swatinem, line 144: the `repo` and `commit_yaml` are only used to initialize the driver, and for the `should_do_flaky_detection` check.
instead, you could pass the driver and that flag from the outside. That way you only need a single function instead of two.

tasks/ta_process_flakes.py
```diff
@@ -0,0 +1,81 @@
...
+            return {"successful": False}
+
+        return {"successful": True}
+
+
+def process_flakes_for_repo(repo_id: int, commit_id):
+    # get all uploads pending process flakes in the entire repo? why stop at a given commit :D
+    uploads_to_process = ReportSession.objects.filter(
+        report__report_type=CommitReport.ReportType.TEST_RESULTS.value,
+        report__commit__repository__repoid=repo_id,
```
Swatinem, line 65:
```suggestion
        report__commit__repository=repo_id,
```
Not sure if the ORM is smart enough to do this automatically, but this could potentially avoid a join on `repository`.

tasks/ta_process_flakes.py
```diff
@@ -0,0 +1,81 @@
...
+        report__commit__commitid=commit_id,
+        state__in=["v2_finished"],
+    ).all()
+    if not uploads_to_process:
+        return
+
+    if settings.BIGQUERY_WRITE_ENABLED:
+        bq = BQDriver(repo_id)
+        bq.write_flakes([upload for upload in uploads_to_process])
```
Swatinem, line 74:
```suggestion
        bq.write_flakes(list(uploads_to_process))
```

PR #1038, opened 2025-01-28
Add a cron task running regular cleanups

Swatinem:
So far this is only hooked up to delete *all* th `ReportDetails`, which are unused.

### DIFF
--- a/celery_config.py
+++ b/celery_config.py
@@ -70,6 +70,8 @@ update_branches_task_name = "app.cron.test_instances.BackfillTestInstancesTask"
 
 cache_rollup_cron_task_name = "app.cron.cache_rollup.CacheRollupTask"
 
+regular_cleanup_cron_task_name = "app.cron.cleanup.RegularCleanup"
+
 
 def _beat_schedule():
     beat_schedule = {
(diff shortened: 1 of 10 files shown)


services/cleanup/regular.py
```diff
@@ -0,0 +1,28 @@
...
+
+from shared.django_apps.reports.models import ReportDetails
+
+from services.cleanup.cleanup import run_cleanup
+from services.cleanup.utils import CleanupResult, CleanupSummary
+
+log = logging.getLogger(__name__)
+
+
+def run_regular_cleanup() -> CleanupSummary:
```
giovanni-guidini, line 11: Shouldn't we have some sort of locking or deduplication here to prevent multiple processes trying to delete the same things?

I'm not sure if `run_cleanup` already does this, or if we are delegating to the DB to return different results when we query...
Maybe it's fine to have multiple processes trying to delete the same thing, just confirming.

giovanni-guidini (commented):
just 1 question 🙏

services/cleanup/regular.py
Swatinem: I can think of one case in which it would be bad to have two concurrent processes running the same query, which I can trivially fix. Otherwise this should be fine to run concurrently.

giovanni-guidini (approved):

services/cleanup/regular.py
```diff
@@ -0,0 +1,28 @@
...
+    log.info("Starting regular cleanup job")
```
adrian-codecov, line 12: I'm assuming these logs are intentional

celery_config.py
```diff
@@ -94,6 +96,13 @@ def _beat_schedule():
                 "cron_task_generation_time_iso": BeatLazyFunc(get_utc_now_as_iso_format)
             },
         },
+        "regular_cleanup": {
+            "task": regular_cleanup_cron_task_name,
+            "schedule": crontab(minute="0", hour="2"),
```
adrian-codecov, line 101: I'm curious how the 2 hour cron will fare. Don't anticipate it collapsing codecov, but I think this would be our most recurrent cron 👀

adrian-codecov (approved):

services/cleanup/regular.py
Swatinem: since this is running infrequently, I would like to get a bit more verbose output. Not too much though :-)

celery_config.py
adrian-codecov: oooooh that makes more sense 😂

PR #1040, opened 2025-01-28
fix: Bundle Analysis - display negative number when size change goes down

JerrySentry:
Screenshot 2025-01-28 at 3 51 34 PM

should show negative numbers here as the size went down

### Legal Boilerplate

### DIFF
--- a/services/bundle_analysis/notify/helpers.py
+++ b/services/bundle_analysis/notify/helpers.py
@@ -1,5 +1,5 @@
 import numbers
-from typing import Literal
+from typing import Literal, Optional
 
 from shared.bundle_analysis import BundleAnalysisComparison, BundleChange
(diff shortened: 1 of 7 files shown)


services/bundle_analysis/notify/helpers.py
```diff
@@ -63,8 +63,9 @@ def get_github_app_used(torngit: TorngitBaseAdapter | None) -> int | None:
     return selected_installation_id
 
 
-def bytes_readable(bytes: int) -> str:
+def bytes_readable(bytes: int, show_negative: Optional[bool] = True) -> str:
```
suejung-sentry, line 66: this api is a little confusing to me but I see what you're going for and don't have any suggestions, so sounds good to go with

suejung-sentry (approved):

PR #1042, opened 2025-01-28
Failed check and status messaging in comment, part IV

nora-codecov:
when user has:
project checks notifier OR project status notifier
AND
comment notifier
AND
has defined `removed_code_behavior` as `adjust_base` in their codecov.yml
AND
the project check/status fails
AND
the `adjusted_base` coverage does NOT allow the check/status to pass
THEN
add some helper text to the comment notifier to point out that the check/status failed, with links to `rcb` and `adjust_base` docs, and tell them the coverage on the head vs the adjusted base.

https://github.com/codecov/engineering-team/issues/1607

### DIFF
--- a/services/notification/notifiers/mixins/status.py
+++ b/services/notification/notifiers/mixins/status.py
@@ -28,6 +28,7 @@ class StatusResult(TypedDict):
 CUSTOM_TARGET_TEXT_PATCH_KEY = "custom_target_helper_text_patch"
 CUSTOM_TARGET_TEXT_PROJECT_KEY = "custom_target_helper_text_project"
 CUSTOM_RCB_INDIRECT_CHANGES_KEY = "custom_rcb_indirect_changes_helper_text"
+CUSTOM_RCB_ADJUST_BASE_KEY = "custom_rcb_adjust_base_helper_text"
 CUSTOM_TARGET_TEXT_VALUE = (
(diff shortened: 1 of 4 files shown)


services/notification/notifiers/mixins/status.py
```diff
@@ -27,16 +27,24 @@ class StatusResult(TypedDict):
 
 CUSTOM_TARGET_TEXT_PATCH_KEY = "custom_target_helper_text_patch"
```
giovanni-guidini, line 28: With so many keys, maybe it's more organized to create an Enum for them?

giovanni-guidini (commented):
The code LGTM. Left a nit comment about enum.

I do have 1 question about this:
the helper text is a dictionary, and the helper-text-section will write *all* helper texts in this dictionary.

If `adjust_base` failed then the regular project check also failed - by definition. Wouldn't that also create a helper text?

My question is: if there are 2 helper texts in the help text dict, will the comment have 2 messages about virtually the same thing?
I haven't seen a test for this case.

services/notification/notifiers/mixins/status.py
Swatinem: Or a `TypedDict`, where you have safety and autocompletion when writing keys, but you can iterate over only the values, or the k/v pairs.

nora-codecov: updated!

giovanni-guidini (approved):
Thank you for clarifications

Swatinem (approved):

PR #1046, opened 2025-01-29
Parallelize compute component comparisons

joseph-sentry:
We're doing this for perf reasons. It's put behind a feature flag for now. We're gonna be queuing up one task per component which will "rebuild the world up to that point" by recreating the comparison and ComparisonProxy. There's no risk there because every transaction before that point will have committed. At that point it will find the component that it was tasked with computing the comparison for, and then compute the comparison for it and save that to the db.

### DIFF
--- a/services/comparison_utils.py
+++ b/services/comparison_utils.py
@@ -1,6 +1,7 @@
 import sentry_sdk
 from shared.reports.readonly import ReadOnlyReport
 
+from database.models import CompareCommit
 from services.comparison import ComparisonContext, ComparisonProxy
 from services.comparison.types import Comparison, FullCommit
 from services.report import ReportService
@@ -8,13 +9,12 @@ from services.report import ReportService
 
 @sentry_sdk.trace
 def get_comparison_proxy(
-    comparison, current_yaml, installation_name_to_use: str | None = None
+    comparison: CompareCommit,
+    report_service: ReportService,
 ):
     compare_commit = comparison.compare_commit
     base_commit = comparison.base_commit
-    report_service = ReportService(
-        current_yaml, gh_app_installation_name=installation_name_to_use
-    )
+
     base_report = report_service.get_existing_report_for_commit(
         base_commit, report_class=ReadOnlyReport
     )
@@ -31,5 +31,7 @@ def get_comparison_proxy(
             patch_coverage_base_commitid=patch_coverage_base_commitid,
             enriched_pull=None,
         ),
-        context=ComparisonContext(gh_app_installation_name=installation_name_to_use),
+        context=ComparisonContext(
+            gh_app_installation_name=report_service.gh_app_installation_name
+        ),
     )
--- a/services/comparison_utils.py
+++ b/services/comparison_utils.py
@@ -1,0 +1,35 @@
+import sentry_sdk
+from shared.reports.readonly import ReadOnlyReport
+
+from services.comparison import ComparisonContext, ComparisonProxy
+from services.comparison.types import Comparison, FullCommit
+from services.report import ReportService
+
+
+@sentry_sdk.trace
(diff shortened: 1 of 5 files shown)


adrian-codecov (approved):

services/comparison_utils.py
```diff
@@ -0,0 +1,35 @@
...
+
+from services.comparison import ComparisonContext, ComparisonProxy
+from services.comparison.types import Comparison, FullCommit
+from services.report import ReportService
+
+
+@sentry_sdk.trace
+def get_comparison_proxy(
+    comparison, current_yaml, installation_name_to_use: str | None = None
+):
```
Swatinem, line 12: please fully type this method, it should be straight forward enough :-)

services/comparison_utils.py
```diff
@@ -0,0 +1,35 @@
...
+    compare_commit = comparison.compare_commit
+    base_commit = comparison.base_commit
+    report_service = ReportService(
+        current_yaml, gh_app_installation_name=installation_name_to_use
```
Swatinem, line 16: can you rather pass this as an argument? I see that the `installation_name_to_use` parameter is used down below as well, but I believe you can also access that as part fo the `report_service`.

tasks/compute_component_comparison.py
```diff
@@ -0,0 +1,94 @@
...
+        comparison_proxy = get_comparison_proxy(
+            comparison, current_yaml, installation_name_to_use
+        )
+        head_commit = comparison_proxy.comparison.head.commit
+
+        yaml: UserYaml = async_to_sync(get_current_yaml)(
+            head_commit, comparison_proxy.repository_service
+        )
+
+        components = yaml.get_components()
```
Swatinem, line 81: You use the `current_yaml` (repo_yaml) in one place, and the commit yaml in another place.
Why is there a difference here? And would it make sense to use one of those instead?

tasks/tests/unit/test_compute_comparison.py
```diff
@@ -17,6 +19,9 @@ class TestComputeComparisonTask(object):
     def test_set_state_to_processed(
         self, dbsession, mocker, mock_repo_provider, mock_storage
     ):
+        mocker.patch.object(
+            PARALLEL_COMPONENT_COMPARISON, "check_value", return_value=False
+        )
```
Swatinem, line 24: this should be the default, right? or are you mocking this because the feature flag would access the DB?

Swatinem (approved):

tasks/compute_component_comparison.py
joseph-sentry: this is just the behaviour of the current task, it uses the repo yaml to get the comparison proxy and it uses the commit yaml to get the components, it would probably make sense to only use the commit yaml, but i didn't want to deviate too much from current behaviour

tasks/tests/unit/test_compute_comparison.py
joseph-sentry: > are you mocking this because the feature flag would access the DB
yes

services/comparison_utils.py
joseph-sentry: out of curiousity: is this to reduce the amount of things the function does, or the amount of params it needs? because this makes it the caller's responsibility, even though the caller may not need the report service

Swatinem: My intention was to make the side effects and dependencies of this function clearer. I also think that type makes it a bit clearer whats being used, vs some "random"-seeming arguments.
And yes, it does delegate responsibility to the caller. Which I believe is a good thing, as the caller can reuse the report service across different calls, and can also handle errors on a different level.

It is true that this does duplicate this call, but the duplication is good if it brings more clarity, which I hope it does.

PR #1047, opened 2025-01-30
Do not log the complete upload contents

Swatinem:
Turns out this was logging the complete "readable" contents of the uploaded test results. We shouldn’t do this.


giovanni-guidini (approved):
Why do we use `upload_id` in one log and `upload.external_id` in the other? Shouldn't we use the same thing?

Swatinem:
I have no idea tbh :-D

PR #1051, opened 2025-01-30
test: clean up logs

joseph-sentry:


Swatinem (approved):
have you checked how often this actually happens?

PR #1054, opened 2025-01-30
feat: add parallel component computation to timeseries

joseph-sentry:
similar to what i just did with the compute component task
when it comes time to upserting measurements in the save_commit_measurements
task, i added a feature flag and behind that feature flag i altered the
behaviour so that instead of upserting the measurements for each component
sequentially it will queue up a task for each component so that they can upsert
in parallel

I had to reorganize the code to avoid some circular imports hence save_commit_measurements
the function, being moved to tasks/save_commit_measurements.py

### DIFF
--- a/services/timeseries.py
+++ b/services/timeseries.py
@@ -4,7 +4,6 @@ from typing import Any, Iterable, Mapping, Optional
 
 from shared.components import Component
 from shared.reports.readonly import ReadOnlyReport
-from shared.timeseries.helpers import is_timeseries_enabled
 from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.orm import Session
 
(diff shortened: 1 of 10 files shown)


tasks/save_commit_measurements.py
```diff
@@ -1,17 +1,90 @@
...
+                            commit.commitid,
+                            commit.repoid,
+                            component.component_id,
+                        )
+                        for component in components
+                    ]
+                )
+                g.apply_async()
+            else:
+                upsert_components_measurements(commit, current_yaml, db_session, report)
```
adrian-codecov, line 83: I see, we changed the `maybe_upsert_component_measurements` to this

adrian-codecov (approved):

tasks/upsert_component.py
```diff
@@ -0,0 +1,138 @@
...
+
+        component = component_dict[component_id]
+
+        if component.paths or component.flag_regexes:
+            report_and_component_matching_flags = component.get_matching_flags(
+                list(report.flags.keys())
+            )
+            filtered_report = report.filter(
+                flags=report_and_component_matching_flags, paths=component.paths
+            )
```
Swatinem, line 125: I believe the primary slowness comes from the `report.filter` call, which is the thing we would like to parallelize.

All the rest should be trivial and fast. And I would imagine that the overhead of task spawning, fetching the report, and fetching the yaml is quite a bit of overhead.

How about:
You do the deduplication, resolution and filtering of these components in the main task.
As the main task is fetching the report anyway, you can end up with a list of `component_id, flags, paths` which are already filtered down to only the relevant ones.

Then when you have that filtered and resolved list, you either iterate over that list, or spawn individual tasks.

services/timeseries.py
```diff
@@ -105,67 +72,66 @@ def maybe_upsert_flag_measurements(commit, dataset_names, db_session, report):
             upsert_measurements(db_session, measurements)
 
 
-def maybe_upsert_components_measurements(
-    commit, current_yaml, dataset_names, db_session, report
+def find_duplicate_component_ids(components: list[Component], commit: Commit):
```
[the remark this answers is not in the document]
joseph-sentry: hmm, i just did a bit of exploration and i found that's it's just user (mis) config where they're defining multiple components with the same name, it's possible they might be expecting us to merge those components but alas

michelletran-codecov (approved):

PR #1055, opened 2025-01-31
Failed check and status messaging in comment, part VI

nora-codecov:
when user has:
changes checks notifier OR changes status notifier
AND
comment notifier
AND
the change check/status fails
AND
there are unexpected/indirect changes
THEN
add some helper text to the comment notifier to point out that the check/status failed, with links to docs

https://github.com/codecov/engineering-team/issues/1627

### DIFF
--- a/services/notification/notifiers/tests/unit/test_status.py
+++ b/services/notification/notifiers/tests/unit/test_status.py
@@ -2658,6 +2658,7 @@ class TestChangesStatusNotifier(object):
         expected_result = {
             "message": "No indirect coverage changes found",
             "state": "success",
+            "included_helper_text": {},
         }
         result = notifier.build_payload(sample_comparison)
         assert expected_result == result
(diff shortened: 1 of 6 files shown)


giovanni-guidini (approved):

services/notification/notifiers/tests/unit/test_status.py
```diff
@@ -2584,6 +2587,11 @@ def test_build_payload_with_multiple_changes(
         expected_result = {
             "message": "3 files have indirect coverage changes not visible in diff",
             "state": "failure",
+            "included_helper_text": {
+                INDIRECT_CHANGES_KEY: HELPER_TEXT_MAP[INDIRECT_CHANGES_KEY].format(
+                    context="changes", notification_type="status"
```
Swatinem, line 2592: just like with the other text, I would prefer the actual text to be asserted here.
otherwise the test is just a mirror of the code doing the formatting.

Swatinem (approved):

PR #1058, opened 2025-01-31
fix: change get_or_create_author() to upsert_author(). track username/name/email changes

matt-codecov:
fixes https://github.com/codecov/engineering-team/issues/3289

i rearranged the function a little bit but the only change is that, if the owner already exists, we will make sure `owner.username`, `owner.name`, and `owner.email` are updated and then commit if anything changed.

### Legal Boilerplate

### DIFF
--- a/services/repository.py
+++ b/services/repository.py
@@ -291,16 +291,16 @@ def upsert_author(
 
     if author:
         needs_update = False
-        if author.username != username and username != None:
+        if author.username != username and username is not None:
             author.username = username
             needs_update = True
-        if author.name != name and name != None:
+        if author.name != name and name is not None:
             author.name = name
             needs_update = True
-        if author.email != email and email != None:
+        if author.email != email and email is not None:
             author.email = email
             needs_update = True
-        
+
         if needs_update:
             db_session.commit()
     else:
--- a/services/repository.py
+++ b/services/repository.py
@@ -291,6 +291,7 @@ def upsert_author(
 
     if author:
         needs_update = False
+        db_session.begin(nested=True)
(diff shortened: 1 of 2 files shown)


services/repository.py
```diff
@@ -281,37 +281,53 @@ async def update_commit_from_provider_info(
...
+
     if author:
-        return author
+        needs_update = False
+        db_session.begin(nested=True)
+        if author.username != username and username is not None:
+            author.username = username
+            needs_update = True
+        if author.name != name and name is not None:
+            author.name = name
```
adrian-codecov, line 299: What's the difference bw username and name here?

services/repository.py
```diff
@@ -281,37 +281,53 @@ async def update_commit_from_provider_info(
...
+            needs_update = True
+        if author.email != email and email is not None:
+            author.email = email
+            needs_update = True
+
+        if needs_update:
+            db_session.commit()
+        else:
+            db_session.rollback()
```
adrian-codecov, line 308: Will this rollback to the newly begun session in line 294?

adrian-codecov (approved):

services/repository.py
matt-codecov: username is matt-codecov, name is Matt Hammerly

services/repository.py
matt-codecov: yup https://docs.sqlalchemy.org/en/20/orm/session_transaction.html#using-savepoint

nothing was actually done so we could also just .commit() but figured this was clearer to future readers

Swatinem (approved):

PR #1065, opened 2025-02-05
dev: Add default plan name var

ajay-sentry:
Add default plan name var to worker so it doesn't break when we merge https://github.com/codecov/shared/pull/495

### Legal Boilerplate


adrian-codecov (commented):
Don't think this nor the API PR are needed if you put the `DEFAULT_PLAN_NAME = get_config("setup", "default_plan_name", default="users-developer")` in db_settings.py, see how it's imported here https://github.com/codecov/codecov-api/blob/ca253c138ce000c16126f7b3693a7bbd90ce1086/codecov/settings_base.py#L10-L11

PR #1074, opened 2025-02-07
fix: TA finisher task

joseph-sentry:
i made a refactor to the TestResultsNotifier where it now expects a pull request to be passed to it in the constructor. I made the necessary change to support this in the test_results_finisher task but not in the ta_finisher task

### DIFF
--- a/tasks/ta_finisher.py
+++ b/tasks/ta_finisher.py
@@ -303,13 +303,6 @@ class TAFinisherTask(BaseCodecovTask, name=ta_finisher_task_name):
             repo_service, commit, commit_yaml
         )
 
-        if pull:
-            seat_activation_result = self.seat_activation(
-                db_session, pull, commit, commit_yaml, repo_service
-            )
-            if seat_activation_result:
-                return seat_activation_result
-
         upload_error = (
             db_session.query(UploadError)
             .filter(UploadError.upload_id.in_(uploads.keys()))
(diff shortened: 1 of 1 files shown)


tasks/ta_finisher.py
```diff
@@ -323,33 +316,47 @@ def process_impl_within_lock(
...
-        success = True if notifier_result is NotifierResult.COMMENT_POSTED else False
-        TestResultsFlow.log(TestResultsFlow.TEST_RESULTS_NOTIFY)
-
-        self.extra_dict["success"] = success
-        self.extra_dict["notifier_result"] = notifier_result.value
         log.info("Finished test results notify", extra=self.extra_dict)
 
         return {
-            "notify_attempted": True,
+            "notify_attempted": notifier_result is NotifierResult.COMMENT_POSTED,
```
giovanni-guidini, line 359: Feels like there should be other entries in the enum that indicate the notification was _attempted_ (not necessarily succeeded)

But I didn't check the Enum, so might be wrong. Just pointing it out that right now "attempted" and "succeeded" have the same value

giovanni-guidini (approved):
In the PR description you mentioned changing the interface of the notifier (expecting a pull) but I didn't see any difference in the construction of the notifier service.

I think it might have been a new failure mode? (the NO_PULL thing)
Anyway, code LGTM 🤷‍♂️

tasks/ta_finisher.py
joseph-sentry: there's a `TORNGIT_ERROR` variant of the enum which represents the attempted = True, success = False result.  I am reconsidering having the NotifierResult in general as I think it might be making the code more confusing.

PR #1078, opened 2025-02-10
create utils for accessing testrun timescale models

joseph-sentry:
this PR creates some utility functions for inserting and aggregating testrun data in timescale

`transaction=True` is required for the test_get_testrun_summary for the continuous aggregates to populate the data that is expected to be in the summary for the test to pass

depends on: https://github.com/codecov/shared/pull/508

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,4 +83,4 @@ dev-dependencies = [
 [tool.uv.sources]
 timestring = { git = "https://github.com/codecov/timestring", rev = "d37ceacc5954dff3b5bd2f887936a98a668dda42" }
 test-results-parser = { git = "https://github.com/codecov/test-results-parser", rev = "190bbc8a911099749928e13d5fe57f6027ca1e74" }
-shared = { git = "https://github.com/codecov/shared", rev = "4755e7f77efc711c6ca34c2fb284edce71836402" }
+shared = { git = "https://github.com/codecov/shared", rev = "e9567e60e40893fd873c319b7dcc7ca05b5685e3" }
--- a/services/test_analytics/ta_timeseries.py
+++ b/services/test_analytics/ta_timeseries.py
@@ -1,0 +1,246 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from datetime import datetime, timedelta
+from typing import TypedDict
+
+import test_results_parser
+from django.db import connections
+from django.db.models import Q
+from shared.django_apps.test_analytics.models import Flake
+from shared.django_apps.timeseries.models import (
+    Testrun,
+    TestrunBranchSummary,
+    TestrunSummary,
+)
+
+from services.test_analytics.utils import calc_test_id
+from services.test_results import FlakeInfo
+
+LOWER_BOUND_NUM_DAYS = 60
+
+
+def get_flaky_tests_set(repo_id: int) -> set[bytes]:
+    return set(
+        Flake.objects.filter(repoid=repo_id, end_date__isnull=True)
+        .values_list("test_id", flat=True)
+        .distinct()
+    )
+
+
+def get_flaky_tests_dict(repo_id: int) -> dict[bytes, FlakeInfo]:
+    return {
+        flake.test_id: FlakeInfo(flake.fail_count, flake.count)
+        for flake in Flake.objects.filter(repoid=repo_id, end_date__isnull=True)
+    }
+
+
+def insert_testrun(
+    timestamp: datetime,
+    repo_id: int | None,
+    commit_sha: str | None,
+    branch: str | None,
+    upload_id: int | None,
+    flags: list[str] | None,
+    parsing_info: test_results_parser.ParsingInfo,
+    flaky_test_ids: set[bytes] | None = None,
+):
+    testruns_to_create = []
+    for testrun in parsing_info["testruns"]:
+        test_id = calc_test_id(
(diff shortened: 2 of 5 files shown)


services/ta_timeseries.py
```diff
@@ -0,0 +1,189 @@
...
+            ["flaky_failure", timestamp, test_id, flags_hash],
+        )
+
+
+def get_testrun_summary(
+    repo_id: int, interval: Interval, branch: str | None = None
+) -> list[TestrunSummary]:
+    timestamp_bin = datetime.now() - timedelta(days=interval.value)
+    return list(
+        TestrunSummary.objects.filter(repo_id=repo_id, timestamp_bin__gte=timestamp_bin)
```
Swatinem, line 177: both these functions return the daily buckets to python, instead of aggregating across the interval within the DB, is that really the intention here?

services/tests/test_ta_timeseries/test_ta_timeseries.py
```diff
@@ -0,0 +1,387 @@
...
+        },
+        flaky_test_ids={
+            calc_test_id("flaky_test_name", "test_classname", "test_suite")
+        },
+    )
+
+    testruns = get_testruns_for_flake_detection(
+        1,
+        {
+            calc_test_id("flaky_test_name", "test_classname", "test_suite"),
```
Swatinem, line 198: the weird autoformatting gives us a hint to make this a local variable :-D

services/tests/test_ta_timeseries/test_ta_timeseries.py
```diff
@@ -0,0 +1,387 @@
...
+            """
+        )
+        cursor.execute(
+            """
+            SELECT remove_continuous_aggregate_policy('timeseries_testrun_summary_1day');
+            select add_continuous_aggregate_policy(
+                'timeseries_testrun_summary_1day',
+                start_offset => '7 days',
+                end_offset => NULL,
+                schedule_interval => INTERVAL '1 milliseconds'
```
Swatinem, line 271: this is a bit weird. given that the default is set to one day, does that mean these aggregates are only updated once a day, and if you query these during the day, they would not give you any result for "today"?

I think the "continuous aggregation" should indeed be continuous and live.
meaning that when you `INSERT` into the testruns table, the materialized view with the aggregate should yield aggregates for that **immediately**. without having to mess with some kind of setting, and without having to put a `sleep` into the test.

services/tests/test_ta_timeseries/test_ta_timeseries.py
```diff
@@ -0,0 +1,387 @@
...
+    assert summaries[0].testsuite == "test_suite"
+    assert summaries[0].classname == "test_classname"
+    assert summaries[0].name == "test_name"
+    assert summaries[0].avg_duration_seconds == 2.0
+    assert summaries[0].last_duration_seconds == 3.0
+    assert summaries[0].pass_count == 1
+    assert summaries[0].fail_count == 1
+    assert summaries[0].flaky_fail_count == 0
+    assert summaries[0].skip_count == 0
+    assert summaries[0].flags == [["flag1"], ["flag2"]]
```
Swatinem, line 363: interesting. so the aggregates are aggregating across all the flags? I thought you still wanted to filter these 🤔

Swatinem (commented):
finally some tests that you can actually run locally :-)
though my main concern is whether the aggregates are actually *live*, see the specific comment.

Swatinem (approved):

services/tests/test_ta_timeseries/test_ta_timeseries.py
joseph-sentry: this is a direct port of what we're showing in the UI right now, to change this would require a product-focused discussion

services/ta_timeseries.py
```diff
@@ -0,0 +1,189 @@
...
+            FROM timeseries_testrun
+            WHERE repo_id = %s AND commit_sha = %s AND outcome IN ('failure', 'flaky_failure')
+            GROUP BY test_id, flags_hash
+            """,
+            [repo_id, commit_sha],
+        )
+        return [
+            {
+                "test_id": bytes(test_id),
+                "flags_hash": bytes(flags_hash),
```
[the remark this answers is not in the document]
joseph-sentry: it's a memoryview, which is basically just a pointer, and to dereference we must wrap it in `bytes`

services/ta_timeseries.py
```diff
@@ -0,0 +1,189 @@
...
+
+def insert_testrun(
+    timestamp: datetime,
+    repo_id: int | None,
+    commit_sha: str | None,
+    branch: str | None,
+    upload_id: int | None,
+    flags: list[str] | None,
+    parsing_info: test_results_parser.ParsingInfo,
+    flaky_test_ids: set[bytes] | None = None,
```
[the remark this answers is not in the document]
joseph-sentry: they shouldn't be, except for maybe `flags` and `flaky_test_ids`, i'll fix

services/ta_timeseries.py
joseph-sentry: we can discuss this more, but I thought we wanted to use the same idea from your binary format where the cached file has individual rows for each day so that we can prune out of date rows when reading from the cache file

if we're skipping the cached file and doing this query on the fly each time, then we should aggregate within the DB

but i think we should compare those 2 alternatives in prod with real data to see which one is faster

PR #1079, opened 2025-02-10
fix: References to old free plan and Deprecated BillingPlan enum

ajay-sentry:
This PR updates a couple references of "users-basic" to use the new "DEFAULT_FREE_PLAN" variable, which will also help us if we update that free plan later on so we don't miss these references in the future.

Additionally, I removed all instances of the "billingPlan" enum in favor of the PlanName enum since that one's deprecated

### Legal Boilerplate

### DIFF
--- a/tasks/github_marketplace.py
+++ b/tasks/github_marketplace.py
@@ -2,8 +2,8 @@ import logging
 from datetime import datetime
 
 import requests
-from shared.billing import BillingPlan
 from shared.celery_config import ghm_sync_plans_task_name
+from shared.plan.constants import DEFAULT_FREE_PLAN
 
 from app import celery_app
 from database.models import Owner, Repository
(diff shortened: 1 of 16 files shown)


tasks/github_marketplace.py
```diff
@@ -243,7 +243,7 @@ def create_or_update_to_free_plan(self, db_session, ghm_service, service_id):
 
         if owner:
             # deactivate repos and remove all activated users for this owner
-            owner.plan = BillingPlan.users_basic.value
+            owner.plan = DEFAULT_FREE_PLAN
```
ajay-sentry, line 246: BUG 1

tasks/trial_expiration.py
```diff
@@ -21,7 +21,7 @@ def run_impl(self, db_session, ownerid, *args, **kwargs):
         log.info(
             "Expiring owner's trial and setting back to basic plan", extra=log_extra
         )
-        owner.plan = BillingPlan.users_basic.value
+        owner.plan = DEFAULT_FREE_PLAN
```
ajay-sentry, line 24: BUG 2

nora-codecov (approved):

PR #1081, opened 2025-02-11
Bundle Analysis: fix processor retry

JerrySentry:
Fix a retry issue where it doesn't trigger more than once and change the cooldown to be exponentially increasing

### Legal Boilerplate

### DIFF
--- a/tasks/bundle_analysis_processor.py
+++ b/tasks/bundle_analysis_processor.py
@@ -30,6 +30,8 @@ bundle_analysis_processor_task_name = (
 class BundleAnalysisProcessorTask(
     BaseCodecovTask, name=bundle_analysis_processor_task_name
 ):
+    max_retries = 5
+
     def run_impl(
         self,
         db_session,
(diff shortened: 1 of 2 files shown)


tasks/bundle_analysis_processor.py
```diff
@@ -139,9 +141,22 @@ def process_impl_within_lock(
...
+                    "Attempting to retry bundle analysis upload",
+                    extra=dict(
+                        repoid=repoid,
+                        commit=commitid,
+                        commit_yaml=commit_yaml,
+                        params=params,
+                        result=result.as_dict(),
+                    ),
+                )
+                self.retry(countdown=30 * (2**self.request.retries))
```
suejung-sentry, line 159: Do we need any MAX_BACKOFF or add any jitter to the retries? Also this job will be retrying for ~10 minutes in worst case, is that OK? (honestly not sure what are the best practices in worker for that)

suejung-sentry (approved):
approved this bugfix + a question if we wanted to enhance this

tasks/bundle_analysis_processor.py
JerrySentry: Hmm probably ok to start with a straightforward retry, the issue is that some times it takes longer for the file to get uploaded to GCS and we start the processor before that happens. Once uploaded the file stays there for a month I think. I doubt it would get to the fifth retry but if that happens then users would wait longer for the PR comment to update.

PR #1082, opened 2025-02-11
Remove Impact Analysis

michelletran-codecov:
This includes removing:
* profiling tasks
* profiling models
* impact analysis comments

### Legal Boilerplate


django_scaffold/settings.py
```diff
@@ -37,7 +37,6 @@
     "shared.django_apps.compare",
     "shared.django_apps.core",
     "shared.django_apps.labelanalysis",
-    "shared.django_apps.profiling",
```
Swatinem, line 40: I’m a bit surprised that this does not have any consequences on my recently added cleanup tasks.
Ideally, we would keep the DB models around for a while, until that cleanup task actually cleans up all the associated entities and files from storage.

Swatinem (approved):
🎉

django_scaffold/settings.py
michelletran-codecov: Ah, ya, I think it's using [the Django models,](https://github.com/codecov/worker/blob/f9ecaf5df3487f15c0d77d3531fe73a8103c96cb/services/cleanup/models.py#L11) so I'll need to clean that up as well. I want to do another pass on worker to get rid of the deprecation notice feature flag, so I can clean up the Django model usage in a separate PR.

PR #1083, opened 2025-02-11
chore: remove TA BQ code

joseph-sentry:
we won't be needing this in the future anymore


nora-codecov (approved):
big cleanup 💪

Swatinem (approved):
TBH, I feel a bit out of the loop here. we discussed (not) using bigquery for a while, but I somehow missed the decision to go ahead and remove this.

joseph-sentry:
>TBH, I feel a bit out of the loop here. we discussed (not) using bigquery for a while, but I somehow missed the decision to go ahead and remove this. 

oh sorry about that, there was no formal decision, i just thought since i was adding the Timescale code it would be cleaner to remove the BQ code for now. (It would also make my work with adding the timescale experimentation stuff easier since there's less changes to keep track of).

PR #1090, opened 2025-02-14
Remove Impact Analysis Deprecation Notice

michelletran-codecov:
This mainly removes the Impact Analysis Deprecation Notice from the code.

Also includes a change to remove a table reference from some clean up code.

### Legal Boilerplate

### DIFF
--- a/rollouts/__init__.py
+++ b/rollouts/__init__.py
@@ -9,10 +9,6 @@ SYNC_PULL_USE_MERGE_COMMIT_SHA = Feature("sync_pull_use_merge_commit_sha")
 
 CHECKPOINT_ENABLED_REPOSITORIES = Feature("checkpoint_enabled_repositories")
 
-SHOW_IMPACT_ANALYSIS_DEPRECATION_MSG = Feature(
-    "show_impact_analysis_deprecation_message"
-)
-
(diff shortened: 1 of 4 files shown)


adrian-codecov (approved):

services/cleanup/models.py
```diff
@@ -199,7 +198,6 @@ def cleanup_upload(context: CleanupContext, query: QuerySet) -> CleanupResult:
     CommitReport: cleanup_commitreport,
     Upload: cleanup_upload,
     CommitComparison: partial(cleanup_with_storage_field, "report_storage_path"),
-    ProfilingUpload: partial(cleanup_with_storage_field, "raw_upload_location"),
```
Swatinem, line 202: please revert this. this is needed to actually delete the stored files

Swatinem (changes requested):

services/cleanup/models.py
Swatinem: https://github.com/codecov/worker/pull/1092 I think this should take care to actually clear out the data before dropping the tables for good.

michelletran-codecov: Removed  in favour of the data deletion script.

Swatinem (approved):

PR #1091, opened 2025-02-17
Enable autocommit for cleanup tasks, and other improvements

Swatinem:
This has various improvements to the cleanup tasks:

- First, it adds a `with_autocommit` contextmanager, which enables django autocommit for the various cleanup tasks. As the tasks are designed to be retried on timeouts, we still want to commit deletion progress as we go.
- It adds the `timeseries` cleanup to the `Repository` as well. Previously, when deleting all `Repository`s related to a parent `Owner`, this would have been skipped.
- Captures and ignores all the storage-related errors, and reduces the chunksize a bit. In particular the `regular_cleanup` is currently failing consistently because of some random storage-related errors. This shoul not stop progress on the deletion job.

### DIFF
--- a/services/cleanup/models.py
+++ b/services/cleanup/models.py
@@ -7,18 +7,20 @@ import sentry_sdk
 from django.db.models import Model, Q, QuerySet
 from shared.bundle_analysis import StoragePaths
 from shared.django_apps.compare.models import CommitComparison
-from shared.django_apps.core.models import Commit, Pull
+from shared.django_apps.core.models import Commit, Pull, Repository
 from shared.django_apps.profiling.models import ProfilingUpload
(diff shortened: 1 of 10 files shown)


services/cleanup/models.py
```diff
@@ -31,6 +33,9 @@ def delete_file(bucket_path: tuple[str, str]) -> bool:
             return context.storage.delete_file(bucket_path[0], bucket_path[1])
         except FileNotInStorageError:
             return False
+        except Exception as e:
+            sentry_sdk.capture_exception(e)
```
giovanni-guidini, line 37: this is smart. Today I learned haha

giovanni-guidini (approved):

PR #1095, opened 2025-02-19
update trial expiration task

nora-codecov:
`TrialExpirationTask` was doing outdated behavior, updating it to use PlanService's helper function instead

### DIFF
--- a/tasks/tests/unit/test_trial_expiration.py
+++ b/tasks/tests/unit/test_trial_expiration.py
@@ -1,50 +1,97 @@
-from shared.plan.constants import DEFAULT_FREE_PLAN
+import pytest
+from shared.django_apps.codecov_auth.tests.factories import (
+    OwnerFactory,
+    PlanFactory,
+    TierFactory,
+)
+from shared.plan.constants import DEFAULT_FREE_PLAN, PlanName, TierName
 
 from database.enums import TrialStatus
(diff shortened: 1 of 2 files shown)


adrian-codecov (approved):
Lgtm, @ajay-sentry to double check

tasks/tests/unit/test_trial_expiration.py
```diff
@@ -1,50 +1,100 @@
...
 from tasks.trial_expiration import TrialExpirationTask
 
 
 class TestTrialExpiration(object):
-    def test_trial_expiration_task_with_pretrial_users_count(self, dbsession, mocker):
-        owner = OwnerFactory.create(pretrial_users_count=5)
-        dbsession.add(owner)
-        dbsession.flush()
+    @pytest.fixture(autouse=True)
+    @pytest.mark.django_db
```
michelletran-codecov, line 15: You can also annotate the class instead of the functions to reduce the number of times you have to add this line.

michelletran-codecov (approved):
Just a small suggestion. Thanks for migrating this task to using Django models ❤️

tasks/tests/unit/test_trial_expiration.py
nora-codecov: omg 🤦 obviously 😂 ty for suggestion, fixing

PR #1097, opened 2025-02-19
makefile changes for monorepo prep

matt-codecov:
needed for https://github.com/codecov/umbrella/pull/4

we bindmount our local `worker` into the docker container. currently our local `worker` is the actual git repo, so the `git` invocation works. in umbrella, it's not the actual git repo, so the command fails

apparently this is all i need to change...?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -218,7 +218,7 @@ test_env.install_cli:
 test_env.container_prepare:
 	apt-get update
 	apt-get install -y git build-essential netcat-traditional
-	git config --global --add safe.directory /worker
+	git config --global --add safe.directory /worker || true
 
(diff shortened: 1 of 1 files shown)


adrian-codecov (approved):

Swatinem (commented):
First time I hear about this setting, so I looked it up:

> These config entries specify Git-tracked directories that are considered safe even if they are owned by someone other than the current user. By default, Git will refuse to even parse a Git config of a repository owned by someone else, let alone run its hooks […]

This makes total sense. But I also wonder why we need `git` access at all within the container?

Do you know what would happen if we just remove this, along with the `apt-get install git` above?

PR #1101, opened 2025-02-20
fix: process flakes task redis lock and key

joseph-sentry:
Previously, I made the decision to use a redis lock to exclude multiple process_flakes tasks from running for the same repo at the same time, so that we would minimize database concurrency issues. 

The idea was that I would set this redis key that represents "more flakes need to be processed for this repo" and that tasks that failed to acquire the lock could rely on the currently running task to handle their work so they wouldn't have to block. I never actually implemented the part where the currently running task takes the work of tasks that ceded their work to it.

This commit implements that by replacing the value of the current redis key with a list of commits that need to be processed, this way the currently running task actually has access to the commits that need to be processed.

### DIFF
--- a/tasks/process_flakes.py
+++ b/tasks/process_flakes.py
@@ -1,8 +1,10 @@
 import logging
 from typing import Any
 
+from redis import Redis
 from redis.exceptions import LockError
 from shared.celery_config import process_flakes_task_name
+from sqlalchemy.orm import Session
 
 from app import celery_app
 from services.processing.flake_processing import process_flake_for_repo_commit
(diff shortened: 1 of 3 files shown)


tasks/process_flakes.py
```diff
@@ -53,18 +54,30 @@ def run_impl(
...
-                    redis_client.delete(f"flake_uploads:{repo_id}")
-                    process_flake_for_repo_commit(repo_id, commit_id)
+                try:
+                    while (commit_ids := redis_client.lpop(list_key, 10)) is not None:
+                        for commitid in commit_ids:
+                            process_func(repo_id, commitid.decode())
+
+                except ResponseError as e:
+                    if "WRONGTYPE" in str(e) and commit_id:
+                        while redis_client.getdel(list_key) is not None:
```
Swatinem, line 76: as the corresponding `rpush` has a similar compatibility code, this might run into a race condition:

- this task tries to `lpop`, runs into an exception
- the finisher tries to `lpush`, runs into an exception
- the finisher completes its "delete + rpush" first
- we then run the `getdel` here

I believe you can avoid this backwards compatibility code by just using a different redis key (though still make sure that the `flake_uploads` one is being deleted, as I believe it is not defined with a TTL so might hang around for eternity otherwise.

Swatinem (approved):
approving to unblock you here, but it would be nice to still rework this a bit to be more resilient towards races.

Swatinem (approved):

PR #1103, opened 2025-02-24
fix: Update default plan value on owner model to DEFAULT_PLAN_VAL

ajay-sentry:
@calvin-codecov found that owners were still being created with users-basic, which is a value being derived from the DB

### Legal Boilerplate


nora-codecov (approved):

adrian-codecov (commented):
Did some research myself and confirmed `default` takes precedence over a sql defined default

PR #1107, opened 2025-02-25
add get_repo_provider_service fn to base task

matt-codecov:
`get_repo_provider_service()` can raise two exceptions which we do not consistently catch. from a quick scan of usages, it appears to make sense to handle them the same way across tasks, so this PR adds a wrapper method to `BaseCodecovTask` that behaves as follows:
- calls `get_repo_provider_service()`
- if `RepositoryWithoutValidBotError`, call `save_commit_error()` and return `None`. each task can decide whether it can proceed without a `repo_provider` or not
- if `NoConfiguredAppsAvailable`, there are two ways to handle it:
  - if `exc.rate_limited_count > 0`, retry the task at the start of the next hour
  - otherwise, return `None` and each task can decide whether it can proceed without a `repo_provider` or not

i created https://github.com/codecov/engineering-team/issues/3389 to keep track of moving usages of `get_repo_provider_service()` to this new method since there are many usages to move and some of them may prove to be non-trivial

### DIFF
--- a/tasks/base.py
+++ b/tasks/base.py
@@ -9,6 +9,8 @@ from django.db import transaction as django_transaction
 from shared.celery_router import route_tasks_based_on_user_plan
 from shared.metrics import Counter, Histogram
 from shared.timeseries.helpers import is_timeseries_enabled
+from shared.torngit.base import TorngitBaseAdapter
+from shared.typings.torngit import AdditionalData
 from sqlalchemy.exc import (
     DataError,
     IntegrityError,
@@ -19,10 +21,20 @@ from sqlalchemy.exc import (
 from app import celery_app
 from celery_task_router import _get_user_plan_from_task
 from database.engine import get_db_session
+from database.enums import CommitErrorTypes
+from database.models.core import (
+    GITHUB_APP_INSTALLATION_DEFAULT_NAME,
+    Commit,
+    Repository,
+)
 from helpers.checkpoint_logger import from_kwargs as load_checkpoints_from_kwargs
 from helpers.checkpoint_logger.flows import TestResultsFlow, UploadFlow
+from helpers.clock import get_seconds_to_next_hour
+from helpers.exceptions import NoConfiguredAppsAvailable, RepositoryWithoutValidBotError
 from helpers.log_context import LogContext, set_log_context
+from helpers.save_commit_error import save_commit_error
 from helpers.telemetry import TimeseriesTimer, log_simple_metric
+from services.repository import get_repo_provider_service
 
 log = logging.getLogger("worker")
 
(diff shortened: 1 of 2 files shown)


adrian-codecov (approved):
In principle this looks good to me. I also think this should apply to `get_owner_provider_service` as it also calls the same fn. I believe Gio worked on this fn a while back so might be worth getting his opinion too

tasks/base.py
```diff
@@ -375,3 +390,52 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
...
+            )
+        except RepositoryWithoutValidBotError:
+            save_commit_error(
+                commit,
+                error_code=CommitErrorTypes.REPO_BOT_INVALID.value,
+                error_params=dict(repoid=repository.repoid),
+            )
+            log.warning(
+                "Unable to reach git provider because repo doesn't have a valid bot"
+            )
```
nora-codecov, line 413: why not add additional identifiers to this log like the `extra` of the ones below?

tasks/base.py
```diff
@@ -375,3 +390,52 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
...
 
         return res
+
+    def get_repo_provider_service(
+        self,
+        repository: Repository,
+        installation_name_to_use: str = GITHUB_APP_INSTALLATION_DEFAULT_NAME,
+        additional_data: AdditionalData = None,
+        commit: Commit = None,
+    ) -> TorngitBaseAdapter:
```
nora-codecov, line 400: isn't it `TorngitBaseAdapter` or `None` now?

tasks/base.py
matt-codecov: `NoConfiguredAppsAvailable` has data members, `RepositoryWithoutValidBotError` has nothing to add. `LogContext` is currently disabled but i think that should provide most of the context one might need when we re-enable it

tasks/base.py
```diff
@@ -375,3 +390,51 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
...
+                log.warning(
+                    "Unable to get repo provider service due to rate limits. Retrying again later.",
+                    extra=dict(
+                        apps_available=exp.apps_count,
+                        apps_rate_limited=exp.rate_limited_count,
+                        apps_suspended=exp.suspended_count,
+                        countdown_seconds=retry_delay_seconds,
+                    ),
+                )
+                self._retry(countdown=retry_delay_seconds)
```
Swatinem, line 429: I’m not a big fan of putting more methods onto the "celery task", but instead move more code out of the "celery as task scheduling framework".

I think the only reason for this to be tied to "celery as a framework" is converting the exception to a retry.

PR #1109, opened 2025-02-26
move sync_repos_languages task to get_repo_provider_service wrapper

matt-codecov:
depends on: https://github.com/codecov/worker/pull/1107
part of: https://github.com/codecov/engineering-team/issues/3389

### Legal Boilerplate

### DIFF
--- a/tasks/sync_repo_languages.py
+++ b/tasks/sync_repo_languages.py
@@ -8,7 +8,6 @@ from sqlalchemy.orm.session import Session
 from app import celery_app
 from database.models.core import Repository
 from helpers.clock import get_utc_now
-from helpers.exceptions import RepositoryWithoutValidBotError
 from helpers.github_installation import get_installation_name_for_owner_for_task
 from services.repository import get_repo_provider_service
 from tasks.base import BaseCodecovTask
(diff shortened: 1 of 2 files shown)


tasks/sync_repo_languages.py
```diff
@@ -46,45 +45,50 @@ def run_impl(
             and len(desired_languages_intersection) == 0
         ) or manual_trigger
 
+        if not should_sync_languages:
+            return {"successful": True, "synced": False}
+
+        log_extra = dict(
+            owner_id=repository.ownerid or "",
+            repository_id=repository.repoid,
```
Swatinem, line 53: both of these are now covered by the `LogContext`, so I would not duplicate them here.

tasks/sync_repo_languages.py
```diff
@@ -46,45 +45,50 @@ def run_impl(
...
+        )
+        log.info("Syncing repository languages", extra=log_extra)
+
+        installation_name_to_use = get_installation_name_for_owner_for_task(
+            self.name, repository.owner
+        )
+        repository_service = get_repo_provider_service(
+            repository, installation_name_to_use=installation_name_to_use
+        )
```
Swatinem, line 62: I can remember that I may have created "yet another" wrapper that takes care of this `installation_name_to_use`, and introduces a cache for these things.
Or I tried and failed because these things are not cache-able.

Swatinem (approved):
lgtm, though at least cleaning up the duplicated logs would be nice

PR #1111, opened 2025-02-26
fix: TA should read comment config value

joseph-sentry:
prevously it wasn't checking these settings, but it lead to users getting the TA / upgrade comment only, so we should check it


Swatinem (approved):

Swatinem:
This reminds me that we should clean up the distinction between `ta_finisher` and `test_results_finisher`.

Lets have another chat about what your plans for this were, and lets work towards removing the duplication.

PR #1118, opened 2025-03-01
feat: TA/BA Comment Messaging

ajay-sentry:
This PR aims to create a new set of messaging for users who are not currently using BA or TA directly on the PR comment. It's much less intrusive than what we were doing previously because it's only a single line blurb that's closed by default. But we're still going to be building an opt out in the near future as well.

Closes https://github.com/codecov/internal-issues/issues/826

How it looks on markdown:

Screenshot 2025-03-04 at 3 40 19 PM

Messed up the BA condition statement (so its not showing) when I pushed to stage but it'll basically look like this:

Screenshot 2025-03-04 at 4 10 55 PM

### Legal Boilerplate


joseph-sentry (approved):
lgtm just a quick comment about the message strings

PR #1119, opened 2025-03-03
Bump `sqlalchemy` version

Swatinem:
Previously, `worker` would pin this to v1.3, but that would then fail tests within `shared` in a common workspace, as some tests in `shared` depend on v1.4 features.


giovanni-guidini (approved):

giovanni-guidini:
I expected the changes required to make this move would be more than the few files that were changed, but now I see tests are failing 😅

Swatinem:
The needed changes were a bit annoying, but not too bad :-)

PR #1127, opened 2025-03-05
fix: TA re-run processing summary

joseph-sentry:
we want to dedup set of test instances we consider by test_id in get_test_summary_for_commit the same way we do in
latest_failures_for_commit

we do this because we only want to consider the latest run for each test


ajay-sentry (approved):
send it

PR #1131, opened 2025-03-09
Bundle analysis: carryforward at most once per commit

JerrySentry:
This ensures we don't do BA processor task a bunch of times for each upload that comes in for a commit regardless of the report type. We only need to do cache the bundles one time, all subsequent calls are redundant.

### Legal Boilerplate


services/bundle_analysis/report.py
```diff
@@ -211,17 +211,11 @@ def _attempt_init_from_previous_report(
                 repo_id=commit.repoid,
             ).values_list("bundle_name", flat=True)
```
JerrySentry, line 213: This update logic should be the same as before, basically copy the entire parent commit's ba report over then remove what shouldn't be cached. The old version doesn't do anything different just more confusing.

services/bundle_analysis/report.py
```diff
@@ -211,17 +211,11 @@ def _attempt_init_from_previous_report(
...
 
-            # For each bundle:
-            # if caching is on then update bundle.is_cached property to true
-            # if caching is off then delete that bundle from the report
-            update_fields = {}
```
ajay-sentry, line 217: love this a lot

tasks/upload.py
```diff
@@ -815,6 +818,30 @@ def possibly_carryforward_bundle_report(
...
+        # If there exists a BA report for the commit, we don't need to create a new one
+        # we validate this by checking if there is an report_upload object for the commit
+        # that has a report type of "bundle_analysis"
+        report_upload = (
+            db_session.query(Upload)
+            .join(CommitReport)
+            .filter(
+                CommitReport.commit_id == commit.id,
+                CommitReport.report_type == ReportType.BUNDLE_ANALYSIS.value,
+                Upload.state != "error",
```
ajay-sentry, line 831: Can probably use the enum here --> https://github.com/codecov/codecov-api/blob/ec564808ba8e812c386023c4671d0e4a2526d217/graphql_api/types/enums/enums.py#L76-L81

assuming these are the same types

ajay-sentry: Also, should we check for processed rather than ! error, in the event a report gets stuck for whatever reason or something

ajay-sentry (approved):
quick comment otherwise lgtm

PR #1135, opened 2025-03-12
feat: implement cache rollup task using TA timeseries models

joseph-sentry:
we want to make it so the cache rollup task is capable of reading test analytics information from the timeseries db

this also changes the format of the dataframe being cached, so we'll also change the format of the path at which we will store the cached dataframe

the logic for reading from the timeseries db is:
- if no branch is specified -> read from the repo wide continuous aggs
- if a branch is specified
    - if it's one of the more popular main branch names -> read from the branch scoped continuous aggregates
    - else, directly aggregate from the individual testruns

### DIFF
--- a/services/test_analytics/ta_cache_rollups.py
+++ b/services/test_analytics/ta_cache_rollups.py
@@ -1,0 +1,77 @@
+from datetime import UTC
+from io import BytesIO
+
+import polars as pl
+import shared.storage
+
+from django_scaffold import settings
+from services.test_analytics.ta_timeseries import (
+    get_branch_summary,
+    get_summary,
+    get_testrun_branch_summary_via_testrun,
+)
+
+
+def rollup_blob_path(repoid: int, branch: str | None = None) -> str:
+    return (
+        f"test_analytics/branch_rollups/{repoid}/{branch}.arrow"
+        if branch
+        else f"test_analytics/repo_rollups/{repoid}.arrow"
+    )
+
+
+POLARS_SCHEMA = [
+    "computed_name",
+    ("flags", pl.List(pl.String)),
+    "failing_commits",
+    "last_duration",
+    "avg_duration",
+    "pass_count",
+    "fail_count",
+    "flaky_fail_count",
+    "skip_count",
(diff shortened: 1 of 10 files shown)


services/test_analytics/tests/test_ta_cache_rollups.py
```diff
@@ -0,0 +1,250 @@
...
+        skip_count=0,
+        flaky_fail_count=0,
+        updated_at=dt.datetime.now(dt.timezone.utc),
+        flags=["test-rollups3"],
+    )
+
+    CacheTestRollupsTask().run_impl(
+        _db_session=None,
+        repoid=1,
+        branch="main",
```
Swatinem, line 174: you forgot the test assertions for this case :-)

services/test_analytics/ta_cache_rollups.py
```diff
@@ -0,0 +1,77 @@
...
+    "timestamp_bin",
+]
+
+
+def cache_rollups(repoid: int, branch: str | None = None):
+    storage_service = shared.storage.get_appropriate_storage_service(repoid)
+    serialized_table: BytesIO
+
+    if branch:
+        if branch in {"main", "master", "develop"}:
```
Swatinem, line 43: The `Repository` has a dedicated field for the main branch, we should probably use that instead of hardcoding a set here.
If you still want to hardcode the list, better to define it as a top level const so its more discoverable.

Swatinem (approved):

services/test_analytics/ta_cache_rollups.py
joseph-sentry: >The Repository has a dedicated field for the main branch, we should probably use that instead of hardcoding a set here.
it would make sense to do this, except we want to use the continuous aggregates, and to do this we would need to access the `repo.branch` from timescale, which isn't possible right now. I have ideas on how to do this in the future, but for now, I think this will be good enough for most of our users.

Swatinem: that does make sense, yes.
maybe a boolean `is_main_branch` or something that you feed in when processing, at which time you have access to the repo metadata.

PR #1136, opened 2025-03-12
remove SQL metrics from worker tasks

matt-codecov:
i don't think anyone but me ever actually used these metrics? i am going to delete them, and delete the exposure logging in the `Feature` system in shared, to cut down on writes and logs and whatall

### Legal Boilerplate


adrian-codecov (approved):
Approved pending tests fixed, and confirmation from the team nobody needs/wants this feature

PR #1140, opened 2025-03-13
feat: implement flake processing using timeseries models

joseph-sentry:
- remove code related to old upload states
- create implementation of flake processing that consumes data from timescale
- process flakes uses impl_type

idea is that we'll start with impl_type both for a while which will persist flakes to the current db table and the new one, then at some point we will detect flaky tests in the finisher by checking the new flakes table

### DIFF
--- a/services/test_analytics/ta_process_flakes.py
+++ b/services/test_analytics/ta_process_flakes.py
@@ -1,0 +1,121 @@
+import logging
+from datetime import datetime
+
+from django.db import transaction
+from django.db.models import Q, QuerySet
+from redis.exceptions import LockError
+from shared.django_apps.reports.models import CommitReport, ReportSession
+from shared.django_apps.ta_timeseries.models import Testrun
+from shared.django_apps.test_analytics.models import Flake
+
+from services.redis import get_redis_connection
+
+log = logging.getLogger(__name__)
+
+FAIL_FILTER = Q(outcome="failure") | Q(outcome="flaky_failure") | Q(outcome="error")
+
+LOCK_NAME = "ta_flake_lock:{}"
+KEY_NAME = "ta_flake_key:{}"
+
+
+def get_relevant_uploads(repo_id: int, commit_id: str) -> QuerySet[ReportSession]:
+    return ReportSession.objects.filter(
+        report__report_type=CommitReport.ReportType.TEST_RESULTS.value,
+        report__commit__repository__repoid=repo_id,
+        report__commit__commitid=commit_id,
+        state__in=["processed"],
+    )
+
+
+def fetch_current_flakes(repo_id: int) -> dict[bytes, Flake]:
+    return {
+        bytes(flake.test_id): flake for flake in Flake.objects.filter(repoid=repo_id)
+    }
+
+
+def get_testruns(
+    upload: ReportSession, curr_flakes: dict[bytes, Flake]
+) -> QuerySet[Testrun]:
+    upload_filter = Q(upload_id=upload.id)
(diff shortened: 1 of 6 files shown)


services/test_analytics/ta_process_flakes.py
```diff
@@ -0,0 +1,117 @@
...
+) -> QuerySet[Testrun]:
+    upload_filter = Q(upload_id=upload.id)
+    flaky_pass_filter = Q(outcome="pass") & Q(test_id__in=curr_flakes.keys())
+    return Testrun.objects.filter(upload_filter & (FAIL_FILTER | flaky_pass_filter))
+
+
+def handle_pass(curr_flakes: dict[bytes, Flake], test_id: bytes):
+    curr_flakes[test_id].recent_passes_count += 1
+    curr_flakes[test_id].count += 1
+    if curr_flakes[test_id].recent_passes_count == 30:
```
Swatinem, line 47: moving this magic constant to the top level would make sense. and an explanation of "after X passes in a row, the test is not marked as flaky anymore"

services/test_analytics/ta_process_flakes.py
```diff
@@ -0,0 +1,117 @@
...
+def get_testruns(
+    upload: ReportSession, curr_flakes: dict[bytes, Flake]
+) -> QuerySet[Testrun]:
+    upload_filter = Q(upload_id=upload.id)
+    flaky_pass_filter = Q(outcome="pass") & Q(test_id__in=curr_flakes.keys())
+    return Testrun.objects.filter(upload_filter & (FAIL_FILTER | flaky_pass_filter))
+
+
+def handle_pass(curr_flakes: dict[bytes, Flake], test_id: bytes):
+    curr_flakes[test_id].recent_passes_count += 1
```
Swatinem, line 45: you should check whether `test_id` is even in `curr_flakes`, otherwise this might fail?

services/test_analytics/tests/test_ta_process_flakes.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+def test_flake_expiry_and_recreation(setup_test_data):
+    result = setup_test_data(
+        uploads=[
+            {
+                "state": "processed",
+                "testruns": [
+                    {"test_id": "test1", "outcome": "pass"},
+                    {"test_id": "test1", "outcome": "failure"},
```
Swatinem, line 224: testing with 2x `pass` would trigger the error I mentioned above, as you clear the test from the current flakes, and on the second iteration try to access that yet again.

tasks/process_flakes.py
```diff
@@ -78,8 +80,13 @@ def run_impl(
             extra=dict(repoid=repo_id, commit=commit_id),
         )
 
+        if impl_type == "new" or impl_type == "both":
+            process_flakes_for_repo(repo_id)
```
Swatinem, line 84: in case you use `both`, you are locking twice in a row. maybe if you move the locking logic out of the function, and the invocation into the lock here, that could be avoided.

joseph-sentry: i think this is necessary because they're locking different locks and reading from different keys 

there's an edge case where a commit is leftover if we lock only once:
```
call Task 1 with commit A, old_key = [A], new key = [B]
Task 1: takes repo lock 
Task 1: completes new invocation, old key = [A], new key = []
call Task 2 with commit B, old key = [A, B], new key = [B]
Task 2: fails to take lock and just drops 
Task 1: completes old invocation, old key = [], new key = [B]
```
B is left over and has to wait for another invocation of process flakes to get processed

PR #1142, opened 2025-03-13
types: make TA notification payload generic

joseph-sentry:
making this change because the test_ids used in timescale will be bytes instead of strs


Swatinem (approved):
I wonder if instead it would make sense to just convert this to a string in that case?

Or are we not printing this at all, and all the generics are "internally resolved"?

PR #1148, opened 2025-03-17
Refactor PR message section logic

Swatinem:
This consolidates a bit of code, and makes the logic a bit clearer which configured layout sections end up where.

It also removes a bogus "Additional details" expander in case none of its children are configured.

### DIFF
--- a/services/notification/notifiers/mixins/message/sections.py
+++ b/services/notification/notifiers/mixins/message/sections.py
@@ -1,6 +1,7 @@
 import logging
 import random
 from base64 import b64encode
+from dataclasses import dataclass
 from decimal import Decimal
 from enum import Enum, auto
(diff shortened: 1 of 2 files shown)


services/notification/notifiers/mixins/message/sections.py
```diff
@@ -432,12 +455,11 @@ def tree_cell(typ, path, metrics, _=None):
                     links["pull"], table_header
                 )
             )
-            yield (table_layout)
-            for line in starmap(
+            yield table_layout
+            yield from starmap(
```
giovanni-guidini, line 459: ah nice, TIL

giovanni-guidini (approved):
🥔🥔🥔

PR #1152, opened 2025-03-19
stop disabling django autocommit

matt-codecov:
i noticed the other day that, while prod disables autocommit for django, our CI test runs and most people's development environments don't. that's a problem

turning autocommit on for django is also convenient for three of our longer-term goals:
- moving away from task-wide transactions to minimize sitting on locks
- moving from sqlalchemy to django
- moving responsibilities into more clearly defined/isolated "services"
  - e.g. a GitGateway service would be responsible for keeping git-related data synced and worker tasks can stop tripping over each other trying to update the `repo.updatestamp` model field themselves

with autocommit off, we can move code from sqlalchemy/tasks to django/services, but the service doesn't actually get to call the shots about its own data. if a coverage file fails to process, the worker task can decide to roll back any updates that GitGateway made to PR titles or HEAD SHAs for no reason. autocommit puts the service back in control over its domain

there is very little usage of django inside of worker, and none of it particularly requires a task-wide transaction, so this should be safe to do right now

### Legal Boilerplate


Swatinem (approved):
🎉 lets do this! 💪🏻

PR #1155, opened 2025-03-19
Replace usage of `EditableReport` with `Report`

Swatinem:
I have merged these two classes a while back, so lets replace the usage of the obsolete one.

### DIFF
--- a/services/report/__init__.py
+++ b/services/report/__init__.py
@@ -12,7 +12,6 @@ from asgiref.sync import async_to_sync
 from celery.exceptions import SoftTimeLimitExceeded
 from shared.django_apps.reports.models import ReportType
 from shared.reports.carryforward import generate_carryforward_report
-from shared.reports.editable import EditableReport
 from shared.reports.enums import UploadState, UploadType
 from shared.reports.resources import Report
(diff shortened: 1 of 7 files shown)


services/report/__init__.py
```diff
@@ -303,10 +282,12 @@ def get_existing_report_for_commit(
         files = commit.report_json["files"]
         sessions = commit.report_json["sessions"]
         totals = commit.totals
-        res = self.build_report(
-            chunks, files, sessions, totals, report_class=report_class
+
+        if report_class is None:
+            report_class = Report
```
giovanni-guidini, line 287: can this be anything else now? Other than `Report` I mean.

giovanni-guidini: if it can't be anything else then `report_class` is looking obsolete, but I understand we don't want to mess with the function interface just yet

giovanni-guidini (approved):

services/report/__init__.py
Swatinem: There is still the `ReadOnlyReport` that some callers are passing explicitly. So until I can get rid of that, I will keep this as is.

PR #1156, opened 2025-03-19
Allow unrolling subqueries in cleanup task

Swatinem:
For some reason, our postgres is unable to properly optimize a delete query on certain tables, which are using a subquery in the where clause.

It turns out that unrolling that subquery within Python works just fine however. So lets do just that.

As we are building all of the delete queries dynamically based on django relations (or some manually defined ones as well), we have to use some queryset magic as well to reverse those automatically created subqueries.

So another method was added which can parse the `IN (...)` subquery out of a queryset, in order to execute that separately, and issue N+1 delete queries on the target table.


giovanni-guidini (approved):
That's some very advanced query magic

PR #1165, opened 2025-03-21
Remove ATS-related label/staticanalysis tasks

Swatinem:
This does what the title says, removing the task code related to the obsolete labelanalysis and staticanalysis, which were parts of ATS.


giovanni-guidini (approved):
+0, -4K

What a time to be alive

PR #1167, opened 2025-03-21
fix: update migrate-timeseries.sh

joseph-sentry:


trent-codecov (approved):

migrate_timeseries.py
```diff
@@ -0,0 +1,50 @@
...
+# Setup Django environment
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "django_scaffold.settings")
+django.setup()
+
+from django.conf import settings  # noqa: E402
+
+
+def run_migrate_commands():
+    try:
+        print("Running timeseries:rollouts migrations")
```
matt-codecov, line 15: i don't think rollouts uses timeseries?

migrate_timeseries.py
```diff
@@ -0,0 +1,50 @@
...
+        call_command(
+            "migrate",
+            database="timeseries",
+            app_label="rollouts",
+            settings="django_scaffold.settings",
+            verbosity=1,
+        )
+
+        print("Running timeseries:pg_telemetry migrations")
```
matt-codecov, line 24: the pg stands for "postgres", this doesn't use timeseries. also nothing writes to it anymore, it is all set to be deleted and we will probably do that after i return from PTO

matt-codecov (commented):
seems mostly fine

matt-codecov (approved):

PR #1168, opened 2025-03-21
build: include migrating timeseries

joseph-sentry:
Follow up to: https://github.com/codecov/worker/pull/1167

we previously did not include timeseries in INSTALLED_APPS
and we did not include it in the migrate-timeseries.sh script

but there's no good reason for that, and we should try to migrate
all apps and databases if we can

i think monorepo fixes this in some way but for now we should do this
and when we move over to the monorepo we should take a second look
at all these different scripts and try to consolidate the logic


matt-codecov (approved):
if nothing is broken deploy next week since it's late on friday

PR #1171, opened 2025-03-24
fix: process flakes

joseph-sentry:
we recently turned autocommit back on by default, but in the case of the process flakes task we explicitly want it on.

### DIFF
--- a/services/processing/flake_processing.py
+++ b/services/processing/flake_processing.py
@@ -20,6 +20,21 @@ def process_flake_for_repo_commit(
     repo_id: int,
     commit_id: str,
 ):
+    with django_transaction.atomic():
+        process_flake_in_transaction(repo_id, commit_id)
+
+    log.info(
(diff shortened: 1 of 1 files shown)


services/processing/flake_processing.py
```diff
@@ -20,77 +20,77 @@ def process_flake_for_repo_commit(
...
-        state__in=["processed", "v2_finished"],
-    )
-
-    curr_flakes = fetch_curr_flakes(repo_id)
-    new_flakes: dict[str, Flake] = dict()
-
-    rollups_to_update: list[DailyTestRollup] = []
-
-    flaky_tests = list(curr_flakes.keys())
+    with django_transaction.atomic():
```
Swatinem, line 23: instead of indenting all of this code, how about moving the transaction to the caller (along with the trailing log and return statement)

Swatinem (approved):

PR #1177, opened 2025-03-26
tests: fix sql_alchemy_db fixture for timescale

joseph-sentry:
there's a deadlock error that happens where the timescale background worker is running background jobs and we're altering the schema of the tables on which it's running the background job on, so we stop the background worker momentarily to allow the migrations to run.


joseph-sentry:
> so this is the source of flakyness in CI?

imo this is the source of flakiness. 

i'm not sure why we haven't run into this on prod yet, probably because the background worker on prod doesn't run jobs as often, but i don't see why that same deadlock failure wouldn't be possible to run into on prod

PR #1180, opened 2025-03-26
fix: ta_timeseries flake set and dict

joseph-sentry:
when we get bytes out of the DB using django it returns them as a
memoryview so if you want the actual values in bytes you must call
the bytes constructor on the memoryview


services/test_analytics/ta_timeseries.py
```diff
@@ -22,15 +22,16 @@
 
 def get_flaky_tests_set(repo_id: int) -> set[bytes]:
     return set(
```
Swatinem, line 24: in this case you can also use a set comprehension literal

Swatinem (approved):

PR #1181, opened 2025-03-26
fix: update get_ta_url to handle null branch

joseph-sentry:


services/urls.py
```diff
@@ -199,12 +199,16 @@ def get_plan_url(pull: Pull) -> str:
 
 
 def get_test_analytics_url(repo: Repository, commit: Commit) -> str:
+    if commit.branch is not None:
+        branch_name = quote_plus(commit.branch)
+    else:
+        branch_name = "main"
```
Swatinem, line 205: as you have access to `Repository`, please pick the main branch as it is stored in `repo`. (and also push that through `quote_plus`)

Swatinem (approved):

PR #1184, opened 2025-03-27
Bundle Analysis: dedupe caching processing

JerrySentry:
Another attempt at ensuring we don't do BA processor task a bunch of times for each upload that comes in for a commit regardless of the report type. We only need to do cache the bundles one time, all subsequent calls are redundant. In the previous attempt it was directly querying the Upload model for all upload types which is slow, this rework ensures it doesn't directly query Upload and it only makes query if the repo has BA enabled.

### Legal Boilerplate

### DIFF
--- a/tasks/bundle_analysis_processor.py
+++ b/tasks/bundle_analysis_processor.py
@@ -196,6 +196,17 @@ class BundleAnalysisProcessorTask(
 
             processing_results.append(result.as_dict())
         except (CeleryError, SoftTimeLimitExceeded, SQLAlchemyError):
+            log.exception(
+                "Unable to process bundle analysis upload",
+                extra=dict(
+                    repoid=repoid,
+                    commit=commitid,
+                    commit_yaml=commit_yaml,
+                    params=params,
+                    upload_id=upload.id_,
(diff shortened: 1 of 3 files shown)


tasks/bundle_analysis_processor.py
```diff
@@ -114,7 +114,37 @@ def process_impl_within_lock(
         if upload_id is not None:
             upload = db_session.query(Upload).filter_by(id_=upload_id).first()
         else:
-            commit_report = report_service.initialize_and_save_report(commit)
+            # If the upload is not provided that means this is a processor task for caching
```
ajay-sentry, line 117: # This processor task handles caching for reports. When the 'upload' parameter is missing,
# it indicates this task was triggered by a non-BA upload.
#
# To prevent redundant caching of the same parent report:
# 1. We first check if a BA report already exists for this commit
# 2. We then verify there are uploads associated with it that aren't in an error state
#
# If both conditions are met, we can exit the task early since the caching was likely
# already handled. Otherwise, we need to:
# 1. Create a new BA report and upload
# 2. Proceed with caching data from the parent report


Put this comment into cursor just to clean it up a bit, does it still ring true?

PR #1185, opened 2025-03-28
Refactor and simplify notification logic a bit

Swatinem:
Also fixes a few type errors while I was at it.


giovanni-guidini (approved):
🥔🥔

PR #1186, opened 2025-03-28
Bundle Analysis: Remove logging noise from retries

JerrySentry:
### Legal Boilerplate


tasks/bundle_analysis_processor.py
```diff
@@ -195,18 +195,10 @@ def process_impl_within_lock(
             result.update_upload(carriedforward=carriedforward)
 
             processing_results.append(result.as_dict())
```
JerrySentry, line 197: Putting SQLAlchemyError in the catch all.

ajay-sentry (approved):
lgtm!

PR #1192, opened 2025-03-31
feat: add version metadata in ta_cache_rollups

joseph-sentry:
we want to be able to evolve the schema of the rollups and we can do that by including a version tag in the GCS object metadata.

However, even though in this case we're making the change in worker first, in the future, we should modify the reading code to handle the new schema before modifying the write code, since if we deploy both reader and writer at the same time, its possible a rollup written by a new version of the writer is read by an old version of the reader which doesn't understand the new format

deps on: https://github.com/codecov/shared/pull/590

### DIFF
--- a/services/test_analytics/ta_cache_rollups.py
+++ b/services/test_analytics/ta_cache_rollups.py
@@ -1,5 +1,6 @@
 from datetime import UTC
 from io import BytesIO
+from typing import cast
 
 import polars as pl
 import shared.storage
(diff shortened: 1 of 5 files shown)


services/test_analytics/ta_cache_rollups.py
```diff
@@ -69,15 +95,20 @@ def cache_rollups(repoid: int, branch: str | None = None):
         for summary in summaries
     ]
 
-    serialized_table = pl.DataFrame(
+    df = pl.DataFrame(
         data,
-        POLARS_SCHEMA,
+        V1_POLARS_SCHEMA,
```
adrian-codecov, line 100: How are we going to serialize data for schemas before this V1 in the next 60 days? They won't have the testsuite right, so wouldn't this lead to it line 83 failing?

Swatinem (approved):
lgtm. the good thing is that these are "cache" files, which by definition can be regenerated from the source data at any time.

services/test_analytics/ta_cache_rollups.py
joseph-sentry: the schema is describing the format of what we're writing (the keys in data) so as long as we include a value for each field in the schema in data the schema will be respected:
``` python
schema = ["field1", "field2"]
data = {"field1": value1, "field2": value2} # this is good
data = {"field3": value1} # bad
```

PR #1193, opened 2025-03-31
feat: create ta_finish_upload

joseph-sentry:
this module is responsible for the new implementation of the TA finisher and it gets its test run and flake information from timescale and the new postgres flake table

### DIFF
--- a/services/test_analytics/ta_finish_upload.py
+++ b/services/test_analytics/ta_finish_upload.py
@@ -1,0 +1,275 @@
+import logging
+from typing import Literal
+
+import sentry_sdk
+from asgiref.sync import async_to_sync
+from shared.celery_config import cache_test_rollups_task_name, process_flakes_task_name
+from shared.django_apps.reports.models import ReportSession, UploadError
+from shared.reports.types import UploadType
+from shared.typings.torngit import AdditionalData
+from shared.yaml import UserYaml
+from sqlalchemy.orm import Session
+
+from app import celery_app
+from database.models import Commit, Repository
+from helpers.notifier import NotifierResult
+from helpers.string import shorten_file_paths
+from services.activation import activate_user, schedule_new_user_activated_task
+from services.redis import get_redis_connection
+from services.repository import (
+    EnrichedPull,
+    fetch_and_update_pull_request_information_from_commit,
+    get_repo_provider_service,
+)
+from services.seats import ShouldActivateSeat, determine_seat_activation
+from services.test_analytics.ta_metrics import (
+    read_failures_summary,
+    read_tests_totals_summary,
+)
+from services.test_analytics.ta_process_flakes import KEY_NAME
+from services.test_analytics.ta_timeseries import (
+    TestInstance,
+    get_flaky_tests_dict,
+    get_pr_comment_agg,
+    get_pr_comment_failures,
+)
+from services.test_results import (
+    ErrorPayload,
+    FinisherResult,
+    TACommentInDepthInfo,
+    TestResultsNotificationFailure,
+    TestResultsNotificationPayload,
+    TestResultsNotifier,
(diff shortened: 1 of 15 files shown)


services/test_analytics/ta_finish_upload.py
```diff
@@ -0,0 +1,275 @@
...
+    TestResultsNotificationFailure,
+    TestResultsNotificationPayload,
+    TestResultsNotifier,
+    should_do_flaky_detection,
+)
+
+log = logging.getLogger(__name__)
+
+
+def get_upload_ids(commitid: str) -> dict[int, ReportSession]:
```
ajay-sentry, line 49: maybe get "successful" upload ids? Or something to differentiate this from a generic get by id

services/test_analytics/ta_finish_upload.py
```diff
@@ -0,0 +1,275 @@
...
+        upload.id: upload
+        for upload in ReportSession.objects.filter(
+            report__commit__commitid=commitid,
+            state__in=["processed", "finished", "flake_processed"],
+        )
+    }
+
+
+def get_upload_error(upload_ids: list[int]) -> ErrorPayload | None:
+    error = UploadError.objects.filter(report_session_id__in=upload_ids).first()
```
ajay-sentry, line 60: so if there's multiple errors, we'd just return the first one? And is this deterministic?

services/test_analytics/ta_finish_upload.py
```diff
@@ -0,0 +1,275 @@
...
+                "commit_id": commit.commitid,
+                "impl_type": impl_type,
+            },
+        )
+
+    if commit.branch is not None:
+        celery_app.send_task(
+            cache_test_rollups_task_name,
+            kwargs={
+                "repoid": repo.repoid,
```
ajay-sentry, line 120: is it too late to change this kwarg param to be repo_id to match the process_flakes_task?

services/test_analytics/ta_finish_upload.py
```diff
@@ -0,0 +1,275 @@
...
+    ):
+        redis_client = get_redis_connection()
+        redis_client.set(f"flake_uploads:{repo.repoid}", 0)
+        redis_client.lpush(KEY_NAME.format(repo.repoid), commit.commitid)
+
+        celery_app.send_task(
+            process_flakes_task_name,
+            kwargs={
+                "repo_id": repo.repoid,
+                "commit_id": commit.commitid,
```
Swatinem, line 111: similar to the comment below, `commitid` and `commit_id` is confusing.
the first one is actually the `commit_sha`, and the `X_id` looks a lot like a primary key id, though it isn’t in this case.

services/test_analytics/ta_finish_upload.py
```diff
@@ -0,0 +1,275 @@
...
+            cache_test_rollups_task_name,
+            kwargs={
+                "repoid": repo.repoid,
+                "branch": commit.branch,
+                "impl_type": impl_type,
+            },
+        )
+
+
+def check_seat_activation(db_session: Session, pull: EnrichedPull) -> bool:
```
Swatinem, line 127: I wonder if such a function already exists someplace else?

Swatinem (approved):
lgtm, but might be worth checking the naming again.

services/test_analytics/ta_finish_upload.py
joseph-sentry: i think i'll do that as part of a separate change

services/test_analytics/ta_finish_upload.py
joseph-sentry: > so if there's multiple errors, we'd just return the first one 
yes
> And is this deterministic
good point

PR #1196, opened 2025-04-01
fix: change cache_test_rollups to use repo_id

joseph-sentry:
we want to change the name of the repoid arg to repo_id so it matches the process flakes task

this will take 3 steps:
1. task to accept both repo_id and repoid
2. task callers to send repo_id
3. task to stop accepting repoid


adrian-codecov (approved):
Lgtm

Swatinem (approved):

PR #1198, opened 2025-04-01
fix: step 3 of changing cache task args

joseph-sentry:
since all callers are using repo_id by this point we no longer have
to accept repoid as an arg


adrian-codecov (approved):

Swatinem (commented):
lol, ignore me, should have reviewed the PRs in chronological order :-D

Swatinem (approved):

PR #1199, opened 2025-04-01
fix: remove process flakes commit_sha arg

joseph-sentry:
this is the final step in moving the process flakes task to using the new method of fetching commit shas from redis. Since all TA finisher tasks are writing to redis using the new method, the process flakes task doesn't need to read the old method anymore, so we can remove that.

### DIFF
--- a/tasks/process_flakes.py
+++ b/tasks/process_flakes.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any, Literal
+from typing import Any, Literal, cast
 
 from redis import Redis
 from redis.exceptions import LockError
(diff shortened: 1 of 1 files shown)


tasks/process_flakes.py
```diff
@@ -77,7 +67,7 @@ def run_impl(
         """
         log.info(
             "Received process flakes task",
-            extra=dict(repoid=repo_id, commit=commit_id),
+            extra=dict(repoid=repo_id),
```
adrian-codecov, line 70: Do you still want to log the possible commit_shas?

adrian-codecov (approved):

tasks/process_flakes.py
Swatinem: as the `repo_id` (along with name, and owner info) is part of the log context now, I don’t think we need to add this context explicitly at all.

Swatinem (approved):

PR #1212, opened 2025-04-15
fix(ta): computed name issue

joseph-sentry:
seeing an error in sentry where polars is unable to infer the schema because the computed_name is null for a bunch of testruns

2 things:
- i don't want the computed name to be null (in the future this logic should live in the test results parser lib)
- the polars schema should've been verbose about types from the start to avoid schema inference issues


services/test_analytics/ta_cache_rollups.py
```diff
@@ -36,32 +36,32 @@ def rollup_blob_path(repoid: int, branch: str | None = None) -> str:
...
-    "failing_commits",
-    "last_duration",
-    "avg_duration",
-    "pass_count",
-    "fail_count",
-    "flaky_fail_count",
-    "skip_count",
+    ("computed_name", pl.Utf8),
+    ("testsuite", pl.Utf8),
+    ("flags", pl.List(pl.Utf8)),
```
Swatinem, line 55: It is still defined as `pl.String` in the above "no version" schema.
What is the difference between these two anyway?

Swatinem (approved):

PR #1219, opened 2025-04-23
Add regular `Upload` cleanup

Swatinem:
This adds logic clean up `Upload`s older than 120 days currently.

The logic to do this is carefully crafted to spread out the deletion into smaller chunks, as the production `Upload` table is one of the largest we have, and bulk-deletion of all uploads will probably cause other DB issues.


Swatinem:
This only covers uploads, so it does not touch any "missing report"-related UI. Though the list of uploads for a report will just be empty then.

PR #1223, opened 2025-04-24
fix(ta): pass right arg to cache rollup task

joseph-sentry:
we made the switch from repoid to repo_id previously but i failed to update the calls in CacheRollupCron and the new TA finisher

also the call to process flakes in the new TA finisher no longer requires the commit id arg


Swatinem (approved):
lgtm, though tests are failing

